PR #26, opened 2023-11-03
Edit README file

jsiegle:
Fixing some typos, links, and formatting in the README file

A few questions:
- The output format is only JSON now, not `.mat` files, correct?
- What is the difference between `File > New` and `File > Clear` menu commands?


XX-Yin:
Thanks! Now it’s more formal and clear! 

Yes, it's JSON by default, and .mat is optional. 
File > New and File > Clear direct to the same function. One of them.

PR #46, opened 2023-11-13
Adding verbose errors 

### LINKED ISSUE #43: Error messages are not descriptive
Error messages printed into the system out are not informative, for example "An error occured: could not convert string to float"

We should at least print the file and line number of the error for debugging

alexpiet:
- [ ] Resolves #43


XX-Yin:
This is great and useful!

PR #47, opened 2023-11-13
Fixed an issue when B_NewscalePositions is empty

XX-Yin:
This issue occurs when B_NewscalePositions is empty. Fixed by using try and except.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1771,7 +1771,10 @@ class Window(QMainWindow, Ui_ForagingGUI):
                 self.ShowBasic.setText(Obj['Other_BasicText'])
             # Set newscale position to last position
             if 'B_NewscalePositions' in Obj:
-                last_positions=Obj['B_NewscalePositions'][-1]
+                try:
+                    last_positions=Obj['B_NewscalePositions'][-1]
+                except:
+                    pass
                 if hasattr(self,'current_stage'):
                     try:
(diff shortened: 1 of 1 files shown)


XX-Yin:
Fixed an issue when B_NewscalePositions is empty.

src/foraging_gui/Foraging.py
```diff
@@ -1771,7 +1771,10 @@ def _Open(self):
                 self.ShowBasic.setText(Obj['Other_BasicText'])
             # Set newscale position to last position
             if 'B_NewscalePositions' in Obj:
-                last_positions=Obj['B_NewscalePositions'][-1]
+                try:
```
alexpiet, line 1774: If the problem is that B_NewscalePositions is empty, it would be better to check for it being empty in a if/else block, rather than using a try/except block.

alexpiet (approved):
I'll approve this PR in the spirit of a quick fix.

However I see two problems with this update.
1. Can you provide some text in the main PR about what the problem was, and how you fixed it?
2. Its better to check for variables being empty in a if/else block instead of using try/except. If you need to use a try/except, I think its best practice to catch the exception and print some useful information

src/foraging_gui/Foraging.py
XX-Yin: I agree that it's better to use if/else. You can get a full understanding about what's happening using if/else instead of  try/except.

PR #52, opened 2023-11-14
Creating a logging system for python foraging gui

### LINKED ISSUE #43: Error messages are not descriptive

alexpiet:
**What has changed**
Creates a logging system for warnings and errors from the foraging gui. 
- [x] One log is created for every time the GUI is started. 
- [x] The log is saved at `C:\Users\<username>\Documents\foraging_gui_logs\tower_<number>_gui_log_<time and date>.txt`
- [x] Each log entry contains the time, module, file, function, line number, and a message
- [x] pyOSC prints information to sys.stdout. I could capture stdout to log it, but that seems hacky and likely to cause other problems so I am letting pyOSC go to stdout. 
- [x] The logger captures warnings and adds to the log
- [x] This replaces my previous PR that added print statements to try/except blocks. Now those print statements are logged as errors. 
   - #46 
- [x] Resolves #43
- [ ] Tested on a rig

**Usage**
- To log an item, use `logging.info()`, or `logging.error()`
- More information: https://docs.python.org/3/library/logging.html#
- The log is stored in plain text 

**Expected change in behavior**
- These changes will not be noticeable to the experimenter.
- Instead of printing statements to the terminal window, errors will be logged

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import subprocess
 from datetime import datetime
+import logging
 
 import numpy as np
 from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar
@@ -157,7 +158,7 @@ class OptogeneticsDialog(QDialog,Ui_Optogenetics):
                 if index != -1:
                     eval('self.LaserPowerRight_'+str(Numb)+'.setCurrentIndex(index)')
         except Exception as e:
-            print('dialogs._Frequency: Error ',str(e))
+            logging.error(str(e))
 
     def _activated(self,Numb):
         '''enable/disable items based on protocols and laser start/end'''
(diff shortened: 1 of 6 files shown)


alexpiet:
@hanhou @XX-Yin Is this an OK place to save the logs? "C:\Users<username>\Documents\foraging_gui_logs\gui_log_.txt"

hanhou:
> @hanhou @XX-Yin Is this an OK place to save the logs? "C:\Users\Documents\foraging_gui_logs\gui_log_.txt"

It may be easier for me to fetch the log file if it is located in the same folder as `BehaviorData`, something like `\\LUKWATA\Users\svc_aind_behavior\Documents\BehaviorData\`. But logically it is not part of BehaviorData... Either way is fine I think.

XX-Yin (approved):

hanhou (approved):

XX-Yin (approved):

Issue #59, opened 2023-11-16
Nested logging issues


XX-Yin:
I found that if line 306 is run, lines 307-309 will not run. Instead, it jumps directly to 311 (shown below). In this case, newscale will not be able to connect. I'm wondering if this has anything to do with multithreading. I temporarily solved it by removing "logging.error(str(e))" in the line 306, which is not a satisfied solution and needs to be figured out. 

Foraging.py
![image](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/assets/109394934/be33a4f0-2183-4f96-8854-576060974754)

alexpiet:
Hmm, ok. Probably can fix with a pass, break, or continue statement.

alexpiet:
Can you clarify what you want this function to do? Why do we have nested try/except blocks to begin with?

XX-Yin:
It's not necessary. And we can reorganize it to avoid nested try/except structure for this specific case. There is this nested structure elsewhere in the code. It's better to find a general solution.

alexpiet:
Closing this issue as its non-specific.

Issue #49, opened 2023-11-14
Subject name mismatch between metadata and file name


hanhou:
Sometimes the metadata does not match the saved json file name. I noticed this bug during a sanity check in my data pipeline:

<img width="1230" alt="image" src="https://github.com/AllenNeuralDynamics/dynamic-foraging-task/assets/24734299/ccb1dd76-120f-46bb-a2db-4813a5797aa9">

Looking at the raw json file, it appears that the `subject_id` field was changed **_within a session_**:

<img width="632" alt="image" src="https://github.com/AllenNeuralDynamics/dynamic-foraging-task/assets/24734299/b3cc1f30-c624-4592-9f87-2c7d12837643">

I guess what happened was that the RA corrected the `subject_id` _**after**_ session start. This will be a huge problem for data integrity. To avoid this:

1. I'll ask RAs, whenever they find a metadata error, to stop the session, clear the data, and start again.
2. @XX-Yin Please disable those metadata fields in the GUI when a session is running. A better way maybe to disallow this kind of trial-by-trial changes in metadata.

XX-Yin:
Thanks Han. Yes, I agree to disable these metadata fields during running.

PR #60, opened 2023-11-16
Add reward delay

XX-Yin:
### Describe changes 
1. Added reward delay to the GUI. (0 by default).
2. Print the OSC messages and trial number to the GUI.
3. Remove nested "logging.error(str(e))" in line 306.  
4. Disable metadata fields when running
5. Use ID instead of animal name to name folder

[#71](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/71)
#59
#49

If you want to add a delay to the reward, you need to set "reward delay=".

Not tested in 447. I tested it on my computer and it works as expected.

### Should the experimenters be on the look out for any new behavior?
No

### Others
Even if "reward delay=" is set to zero, caution should be exercised regarding the potential delay between detection of licks and reward delivery. This is not running in hardware and there will probably be a delay of around a few milliseconds (based on my previous testing it was around 2ms with small jitter).

Please remember to notify all experimenters and include the answers to the questions above


src/foraging_gui/rigcontrol.py
```diff
@@ -14,6 +14,7 @@ def msg_handler(self, address, *args):
         msg = OSCMessage(address, args)
         CurrentMessage=[msg.address,args[1][0],msg.values()[2],msg.values()[3]]
         self.msgs.put([msg,args])
+        print(CurrentMessage)
```
alexpiet, line 17: Why print the OSC message to the terminal?

src/foraging_gui/Foraging.py
```diff
@@ -303,7 +303,7 @@ def _StageSerialNum(self):
                 try:
                     instance.io.close()
                 except Exception as e:
-                    logging.error(str(e))
+                    pass#logging.error(str(e))
```
alexpiet, line 306: What happens if we have 
> logging.error(str(e))
> pass

instead of just
> pass

src/foraging_gui/Foraging.py
```diff
@@ -2205,6 +2205,7 @@ def _StartTrialLoop1(self,GeneratedTrials,worker1,workerPlot,workerGenerateAtria
             if self.ANewTrial==1 and self.ToGenerateATrial==1 and self.Start.isChecked(): 
                 self.ANewTrial=0 # can start a new trial when we receive the trial end signal from Bonsai
                 GeneratedTrials.B_CurrentTrialN+=1
+                print('Current trial: '+str(GeneratedTrials.B_CurrentTrialN+1))
```
alexpiet, line 2208: Why have the message about current trial in the terminal and in the log? I'm open with having it replicated, but I'm wondering what the motivation is

src/foraging_gui/rigcontrol.py
XX-Yin: It's nice to see an immediate response when debugging. For example, debugging a lick detector.

src/foraging_gui/Foraging.py
XX-Yin: Yes! Use for debugging or testing when you want to see an immediate response.

src/foraging_gui/Foraging.py
XX-Yin: I haven't tried this, but my speculation is a similar effect.

XX-Yin: Yes, I tested it. The same issue

alexpiet: Ok, this issue is unrelated to the overall reward delay. So we can deal with it separately.

alexpiet:
Can you test it in 447?

XX-Yin:
> Can you test it in 447?

I tested it in the ephys rig and can find a time to test it in 447 tomorrow.

src/foraging_gui/Foraging.py
```diff
@@ -2046,6 +2046,14 @@ def _thread_complete_timer(self):
     def _Timer(self,Time):
         '''sleep some time'''
         time.sleep(Time)
+    
+    def _EnableDisableMetadata(self,state):
```
hanhou, line 2050: The name `_EnableDisableMetadata(self, state)` sounds a bit confusing. How about `_set_metadata_enabled(self, enable: bool)`, logically similar to the native setEnabled().

XX-Yin: sounds good! I will change it.

alexpiet (approved):

Issue #80, opened 2023-12-06
Dialogs:Dialogs.py:_Laser:line 210:'Window' object has no attribute 'RecentCalibrationDate' 


alexpiet:
Error occurred 620 times in the last week

alexpiet:
#103 resolves this issue. Closing now.

PR #103, opened 2023-12-12
Setting default values of GetLaserCalibration() if no calibration file is found

alexpiet:
  
### Describe changes:
- `Foraging._GetLaserCalibration` tries to load the laser calibration file. If none is found, it failed to set the RecentCalibrationDate. 
- Now I sets a default value for the recent calibration date as "None", this prevents an exception from being generated when that variable is accessed. It is used for display purposes only, so I see no downside to setting a default value
- Further, I set an empty dictionary as the value of LaserCalibrationResult, which is added to the `json` file. 
- I do not set a value of `RecentLaserCalibration` because the presence of this attribute is used as a conditional test in the OptogeneticsDialog. 
- After fixing this issue I removed two unnecessary try/except blocks

- [x] resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/80

- No changes for someone running a mouse
- The saved data will now always have a `LaserCalibrationResults` entry, and a `LatestCalibrationDate` entry

- [x] Tested in 447

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -204,10 +204,7 @@ class OptogeneticsDialog(QDialog,Ui_Optogenetics):
             eval('self.OffsetEnd_'+str(Numb)+'.setEnabled('+str(True)+')')
     def _Laser(self,Numb):
         ''' enable/disable items based on laser (blue/green/orange/red/NA)'''
-        try:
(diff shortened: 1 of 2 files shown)


XX-Yin (approved):

XX-Yin (commented):
Great!

PR #108, opened 2023-12-13
[do not merge] Updated Bonsai/GUI connections

### LINKED ISSUE #107: Bonsai connection issues
- The GUI always starts Bonsai, even if it is already running. Should we check to see if bonsai is already running?
- The GUI waits 3 seconds before trying to connect to Bonsai. If Bonsai hasn't opened yet (which is slow), then it returns a time-out. This time sets a flag for a poor connection and results in downstream effects.
  - We could check of Bonsai has connected, then try waiting?

alexpiet:
  
### Describe changes:
Previously, when the GUI launched it would always start a new Bonsai instance, wait 3 seconds, then attempt to connect once. This had two problems, first if Bonsai was already running it would open Bonsai again. The second Bonsai instance would be blocked from accessing ports. Second, if Bonsai took longer than 3 seconds to open then the GUI would not connect and sets an internal flag `self.InitializeBonsaiSuccessfully=0`

Now, the GUI first attempts to connect to an already running Bonsai instance. If it cannot, then it launches Bonsai and checks the connection every 500ms until it connects. If it has not connected after 10 seconds, then it sets the internal flag `self.InitializeBonsaiSuccessfully=0`

- Resolves #107 

- The GUI should reliably connect to Bonsai every time it opens. 
- It should open slightly faster in general, and much faster if Bonsai is already running
- When an error is encountered, the best course of action is still to close both Bonsai and the GUI. 

- [x] Tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -547,6 +547,7 @@ class Window(QMainWindow, Ui_ForagingGUI):
         wait = 0
         while wait < max_wait:
             time.sleep(1)
+            wait += 1
             try:
                 self._ConnectOSC()
             except Exception as e:
@@ -555,6 +556,7 @@ class Window(QMainWindow, Ui_ForagingGUI):
                 logging.info('Connected to Bonsai')               
                 self.InitializeBonsaiSuccessfully=1
                 logging.info('Bonsai started successfully')
+                return
         
         # Could not connect
         logging.info('Could not connect to bonsai with max wait time {} seconds'.format(max_wait))
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -76,15 +76,16 @@ class Window(QMainWindow, Ui_ForagingGUI):
         self.UpdateParameters=1 # permission to update parameters
         self.Visualization.setTitle(str(date.today()))
         self.loggingstarted=-1
-        try: 
-            self._InitializeBonsai()
-            self.InitializeBonsaiSuccessfully=1
-            logging.info('Bonsai started successfully')
-        except Exception as e:
-            logging.error('Initializing Bonsai: {}'.format(str(e)))
-            self.InitializeBonsaiSuccessfully=0
-            self.WarningLabel_2.setText('Start without bonsai connected!')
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -525,12 +517,58 @@ def _GetSettings(self):
...
+        else:
+            # We could connect, set the indicator flag and return
+            logging.info('Connected to already running Bonsai')
+            self.InitializeBonsaiSuccessfully=1
+            logging.info('Bonsai started successfully')
+            return
+
+        # Start Bonsai
+        logging.info('Starting Bonsai')
         self._OpenBonsaiWorkflow()
```
XX-Yin, line 546: If the connection is successful, we should not open another bonsai workflow.

XX-Yin: That doesn't happen here because there is a "return"

src/foraging_gui/Foraging.py
```diff
@@ -525,12 +517,58 @@ def _GetSettings(self):
...
+                self._ConnectOSC()
+            except Exception as e:
+                # We could not connect
+                logging.info('Could not connect, total waiting time {} seconds: '.format(wait)+str(e))
+            else:
+                # We could connect
+                logging.info('Connected to Bonsai after {} seconds'.format(wait))               
+                self.InitializeBonsaiSuccessfully=1
+                logging.info('Bonsai started successfully')
+                return
```
XX-Yin, line 565: The GUI will freeze during these 30 seconds, which is not what we want. We can set max_wait = 3. Normally, it will be completed within 3s, otherwise we can establish the connection manually.

alexpiet: Yes, the GUI will freeze during those 30 seconds, but only if Bonsai has not opened. In the last two weeks 25% of all GUI sessions did not connect in the 3 seconds. So we could set the max_wait to be smaller than 30, but I think it should be greater than 3s. 

What do you mean by establish the connection manually? Do you mean establish the connection later?

alexpiet:
@XX-Yin I changed max_wait to 6s.

XX-Yin (approved):

PR #109, opened 2023-12-14
Check bonsai connection; stop stage before moving; save camera parameters each trial

XX-Yin:
  
### Describe changes:
1. Check bonsai connections when using the camera and laser calibration. 
2. Stop the stage before moving. 
3. Save camera parameters each trial. 
4. Enable the optogenetics2 node. 
5. Add ReconnectBonsai button. 

1.  The GUI will crash If we run some functions without Bonsai connected to the GUI. It's a good to check the connection before running these functions.
2. If the stage has not stopped, the stage will not move when we click the move button.
3. It's better to save the camera parameters each trial. Previously, it only saved the parameter of the last trial. 

No
I tested them in the ephys rig 3. It works.


alexpiet:
What does the ReconnectBonsai button do?

src/foraging_gui/Foraging.py
```diff
@@ -368,6 +370,16 @@ def _ConnectBonsai(self):
                 self.WarningLabelInitializeBonsai.setStyleSheet("color: red;")
                 self.InitializeBonsaiSuccessfully=0
 
+    def _ReconnectBonsai(self):
+        '''Reconnect bonsai'''
+        if self.InitializeBonsaiSuccessfully==1:
+            self.client.close()
```
alexpiet, line 376: Why would we want to close the clients and reconnect?

XX-Yin:
> What does the ReconnectBonsai button do?

It will disconnect the Bonsai from the GUI and connect them again.

src/foraging_gui/Foraging.py
XX-Yin: In some cases, the connection between GUI and Bonsai is interrupted by mistake and we can reconnect them. I also use it for testing when I want to change the bonsai code while the connection is already established so I don't need to reopen the GUI.

src/foraging_gui/Dialogs.py
```diff
@@ -458,6 +458,9 @@ def _LoadCaliPar(self):
 
     def _StartCalibratingLeft(self):
         '''start the calibration loop of left valve'''
+        self.MainWindow._CheckBonsaiConnection()
+        if self.MainWindow.InitializeBonsaiSuccessfully==0:
+            return
```
alexpiet, line 463: What happens if we return? Does nothing happen? Should the user be alerted in some way?

XX-Yin: A warning "Please open Bonsai!" will appear in the main window

src/foraging_gui/Foraging.py
```diff
@@ -2080,12 +2094,16 @@ def _set_metadata_enabled(self, enable: bool):
         self.Experimenter.setEnabled(enable)
         self.Tower.setEnabled(enable)
 
+    def _CheckBonsaiConnection(self):
```
alexpiet, line 2097: This doesn't check the connection, it only establishes a connection if `InitializeBonsaiSuccessfully` = 0. Shouldn't it always check?

XX-Yin: Yes. How about we merge this first? And I will create a separate pull request to solve this issue. This problem only occurs when the Bonsai and GUI are connected and then disconnected.

alexpiet (approved):

src/foraging_gui/Foraging.py
alexpiet: ok

alexpiet: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/111

PR #114, opened 2023-12-15
Cleaning up how settings are loaded

### LINKED ISSUE #110: bonsai_tag variable gets set several times
- Set it once during initialization
- Then remove default checks for len(sys.argv)

### LINKED ISSUE #113: Foraging._GetSettings has redundant code in try/except block

alexpiet:
  
### Describe changes:
Cleans up and streamlines how settings are loaded, and box-specific settings are established. 
- Previously the input argument sys.argv[1] was used multiple times to set the box number. I now set this once at the start, and then we don't need to check it many times. (Defaults to 1)
- I changed the value `Window.bonsai_tag` to `Window.tower_number` for clarity
- I cleaned up the logic of loading settings, so we only define the default values once
- [x] if `ForagingSettings.json` does not exist, or cannot load, the user is alerted in the dialog box
- [x] If `default_saveFolder` and `current_box` are not in the settings file, the user is alerted in the dialog box
- [x] What is going on with: Foraging._OpenBonsaiWorkflow 
    - Everything works on all boxes with the `--start` flag
- [x] Foraging._GetSettings. Logs as info, not error when Tower.setCurrentIndex() cannot be set. 

- resolves #110 
- resolves #113

No changes visible to the user. 

- [x] tested on 447


XX-Yin:
> ### Pull Request instructions:
> * Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
> * Answer the questions below in detail. Your responses will be emailed to experimenters.
> * If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
> * Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
> * If you use the keyword "skip email" in the title, it will skip the email updates
> * Merges from "production_testing" into "main" should use the keyword "production merge" in the title for reliable indexing of updates
> 
> ### Describe changes:
> Cleans up and streamlines how settings are loaded, and box-specific settings are established.
> 
> * Previously the input argument sys.argv[1] was used multiple times to set the box number. I now set this once at the start, and then we don't need to check it many times. (Defaults to 1)
> * I changed the value `Window.bonsai_tag` to `Window.tower_number` for clarity
> * I cleaned up the logic of loading settings, so we only define the default values once
> * [ ]  @XX-Yin I make a log entry if `ForagingSettings.json` does not exist, or cannot load, should I throw an exception, or alert the user?
>   
>   * I think its reasonable to throw an exception - a settings file should always be in the right spot.
> * [ ]  @XX-Yin I make a log entry if we are missing entries in the settings file. Should I throw an exception, or alert the user?
>   
>   * I'm 50/50 here.
> * [ ]  @XX-Yin What is going on with: Foraging._OpenBonsaiWorkflow opening Bonsai different for different boxes
>   
>   * line 643-646 on branch:fix_110
> * [ ]  @XX-Yin Foraging._GetSettings, what is the result of not setting self.Tower.setCurrentIndex()?
>   
>   * And how does self.Tower get populated?
>   * line 510-515 on branch:fix_110
> 
> ### What issues or discussions does this update address?
> * resolves [bonsai_tag variable gets set several times #110](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/110)
> * resolves [Foraging._GetSettings has redundant code in try/except block #113](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/113)
> 
> ### Describe the expected change in behavior from the perspective of the experimenter
> No changes visible to the user.
> 
> ### Describe the outcome of testing this update on a rig in 447
> * [ ]  tested on 447

1. Yes, it's better to alert the user. The GUI can run without the `ForagingSettings.json`, but the "current_box" and the "default_saveFolder" are important to define for each tower. 
2. I suggest we alert the user about the missing of "current_box" and the "default_saveFolder", and throw an except for other fields. 
3. The communication channels (request_port) are different for different bonsai workflow in different towers. 
4. If the self.Tower.setCurrentIndex() is not set, it will use the default item set in the Qt-designer (it's the first item by default).

XX-Yin (approved):

PR #118, opened 2023-12-15
Removing the black console window

### LINKED ISSUE #112: Close the terminal window without closing Bonsai/Gui
In an effort to reduce visual clutter

alexpiet:
  
### Describe changes:
Stops the black console window from appearing when the GUI is launched. The console is window for debugging, but in operations it adds to visual clutter. 

Technical details: This is done by suppressing console generation at two steps.
- First, in the `.bat` file: 
    - `start python Foraging.py` is replaced with `start "" pythonw Foraging.py`
    - The double quotes uses the existing console for running
    - pythonw runs without generating console output
- Second, when we use `subprocess.popen` to launch bonsai
    -  We add a `shell=True` flag to suppress console generation

If you WANT to run the GUI with the console, (for debugging purposes), then you can open the `.bat` file and swap which line is commented out. 

- resolves #112 

When you launch the GUI, the black window will not appear. 

- [x] Tested in 447
- [x] Tested interactions with multiple box numbers

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -641,7 +641,7 @@ class Window(QMainWindow, Ui_ForagingGUI):
         SettingsBox = 'Settings_box{}.csv'.format(self.tower_number)
         CWD=os.path.join(os.path.dirname(os.getcwd()),'workflows')
         if len(sys.argv) == 1:
-            subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',shell=True)
+            subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',cwd=CWD,shell=True)
         else:
             if self.tower_number==1:
                 subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox,cwd=CWD,shell=True)
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -641,12 +641,12 @@ class Window(QMainWindow, Ui_ForagingGUI):
         SettingsBox = 'Settings_box{}.csv'.format(self.tower_number)
         CWD=os.path.join(os.path.dirname(os.getcwd()),'workflows')
         if len(sys.argv) == 1:
-            subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start')
+            subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',shell=True)
         else:
             if self.tower_number==1:
-                subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox,cwd=CWD)
+                subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox,cwd=CWD,shell=True)
             else:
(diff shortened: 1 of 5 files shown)


XX-Yin:
Generally, I agree to remove the python and bonsai console in the 447, but we should keep these consoles in the Tower_EphysRig3. 

Both python and bonsai console are useful in certain situations. Behavioral events from the python console can be used for debugging. The Bonsai console can output video compression messages that are important for tracking video compression.

XX-Yin (commented):
Great! I need to change the LaserCalibration.json, WaterCalibration.json and WaterCalibrationPar.json to LaserCalibration_1.json, WaterCalibration_1.json and WaterCalibrationPar_1.json in the ephys rig.

alexpiet:
> Generally, I agree to remove the python and bonsai console in the 447, but we should keep these consoles in the Tower_EphysRig3.
> 
> Both python and bonsai console are useful in certain situations. Behavioral events from the python console can be used for debugging. The Bonsai console can output video compression messages that are important for tracking video compression.

I will edit the `.bat` startup files with a commented out version that keeps the console open, that way you can easily configure the ephys rig to have the console.

XX-Yin:
> > Generally, I agree to remove the python and bonsai console in the 447, but we should keep these consoles in the Tower_EphysRig3.
> > Both python and bonsai console are useful in certain situations. Behavioral events from the python console can be used for debugging. The Bonsai console can output video compression messages that are important for tracking video compression.
> 
> @XX-Yin I edited the `.bat` startup files with a commented out version that keeps the console open, that way you can easily configure the ephys rig to have the console.
> 
> ```
> :: This version starts without a console window
> start "" pythonw Foraging.py 1
> :: This version starts with a console window
> :: start python Foraging.py 1
> ```

Thanks!

src/foraging_gui/Foraging.py
```diff
@@ -647,13 +647,8 @@ def _OpenBonsaiWorkflow(self,runworkflow=1):
...
         SettingsBox = 'Settings_box{}.csv'.format(self.tower_number)
         CWD=os.path.join(os.path.dirname(os.getcwd()),'workflows')
-        #if len(sys.argv) == 1:
-        #    subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',cwd=CWD)
-        #else:
-        #    if self.tower_number==1:
-        #        subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox,cwd=CWD)
-        #    else:
-        subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',cwd=CWD)
+        subprocess.Popen(self.bonsai_path+' '+self.bonsaiworkflow_path+' -p '+'SettingsPath='+self.SettingFolder+'\\'+SettingsBox+ ' --start',cwd=CWD,shell=True)
```
XX-Yin, line 650: Can you set the shell==False for Tower_EphysRig1, Tower_EphysRig2 and Tower_EphysRig3? I need to open the bonsai console in the ephys rig.

alexpiet: How do you start the GUI on those rigs? Do you use a `.bat` file? Or do you start it in a console window?

I think that if you start the GUI in a console window or a `.bat` file with `python Foraging.py` then the console will open regardless of whether `shell=True` here. I think `shell=False` only opens a console if one is not already open. I could be wrong though. Can you test it on one of those ephys rigs? I dont know what to look for in the console as Bonsai output

alexpiet: @XX-Yin I just tested this on my laptop by trying to start bonsai with a box number that doesn't exist on my computer, and with `python Foraging.py` the bonsai errors are displayed in the console, like they were previously. I believe this will solve your use case, but I'd like to hear from you about how you start the GUI, and whether you can still see the video output

XX-Yin: @alexpiet In the ephys rig, I also started GUI through a `.bat` file (Previously, I used `start python Foraging.py` instead of `start python Foraging.py 1`).  After your modification, I believe I should change to use `start python Foraging.py 1`

alexpiet: Either `start python Foraging.py` or `start python Foraging.py 1` should work. The box number defaults to 1 if it is not passed in. Did you get a chance to try this out on your ephys rig?

XX-Yin: I tested on the ephys rig. It works.

XX-Yin (approved):

Issue #115, opened 2023-12-15
update .bat start file to add any output to the log file?


alexpiet:

alexpiet:
I tried doing this by putting the call to start the GUI inside a try/except block, but the app.exec() has weird interactions

alexpiet:
Resolved by #126

PR #122, opened 2023-12-18
Fix 115

alexpiet:


alexpiet:
This approach didn't work to resolve #115. Deleting PR and branch

PR #126, opened 2023-12-18
Log all errors and crashes

alexpiet:
  
### Describe changes:
Currently, if the GUI crashes from an uncaught exception it crashes without any information. I added a custom excepthook function that is called by the system when an uncaught exception happens. I now log the information, including the traceback into the logfile.

- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/115
    - it does this through a different solution, but its a better solution.
    
No changes. When the GUI crashes, please continue to make a github issue with the relevant details. Now we will be able to get more information from the appropriate logfile. 

- [ ] Tested in 447. I can't test because 447 is under renovation

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -113,6 +113,7 @@ class Window(QMainWindow, Ui_ForagingGUI):
         self._StageSerialNum()
         self.CreateNewFolder=1 # to create new folder structure (a new session)
         self.ManualWaterVolume=[0,0]
+        assert False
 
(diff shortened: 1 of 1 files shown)


XX-Yin:
Do these capture all output in the console?

alexpiet:
> Do these capture all output in the console?

This capture all python exceptions. It will log them to the log file, then additional send them to the console. 

Bonsai outputs, both normal dialog and errors, will not be affected by this change.

XX-Yin (approved):

Issue #117, opened 2023-12-15
GetWaterCalibration does not set defaults


alexpiet:
- set a default empty dictionary, and null calibration date
- clean up logic of loading, like I did with LaserCalibration

alexpiet:
Resolved by #128

PR #128, opened 2023-12-18
Fix water calibration

alexpiet:
  
### Describe changes:
Previously GetWaterCalibration did not set default values if no recent calibration was found, now it does, using parallel logic to GetLaserCalibration

- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/117

No Changes. 

- [ ] Tested in 447. I have not tested because 447 is under-going renovations

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -59,11 +59,7 @@ class Window(QMainWindow, Ui_ForagingGUI):
 
         # Load Laser and Water Calibration Files
         self._GetLaserCalibration()
-        try:
-            self._GetWaterCalibration()
-            logging.info('Loaded Water Calibration')
-        except Exception as e:
-            logging.error('Could not load water calibration file: {}'.format(str(e)))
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -59,11 +59,7 @@ def __init__(self, parent=None,tower_number=1):
 
         # Load Laser and Water Calibration Files
         self._GetLaserCalibration()
-        try:
-            self._GetWaterCalibration()
-            logging.info('Loaded Water Calibration')
-        except Exception as e:
-            logging.error('Could not load water calibration file: {}'.format(str(e)))
+        self._GetWaterCalibration()
```
XX-Yin, line 62: Why do we need to run self._GetWaterCalibration() two times?

alexpiet: It just runs once. The first call on line 61 is for GetLaserCalibration

XX-Yin (approved):

Issue #99, opened 2023-12-09
Discussion: streamline Qt Designer workflow


hanhou:
Now we need to run `pyuic` to generate `.py` from `.ui` each time we update the UI in Qt designer. Another possibility is to use `uic.loadUi()`, which directly loads the UI from `.ui` file to python code. See [here](https://stackoverflow.com/a/61263195/13429084) for a detailed comparison.

I feel like the second approach may be a better choice. I'll play with it and see how it compares with the `pyuic` approach.

PR #130, opened 2023-12-21
[skip email] Han refactor uic loadui

hanhou:
  
### Describe changes:
Streamlined Qt Designer workflow. Using `uic.loadUi` to dynamically load GUI from `.ui` files. Don't need to run `pyuic5` any more. 

- #99 

None. Only developers will notice.

Tested on my local PC. After any changes in Qt Designer, the GUIs are automatically updated when `Foraging.py` is run again.


alexpiet (changes requested):
Add a `.gitignore` file.

hanhou:
> This is great, it cleans up a lot of redundant files.
> 
> * How long does it take to generate the files from the ui?
> * We should add the generated files to a `.gitignore`, and commit that to the repo. This will keep the local copy clean

I didn't notice any significant slow down during initialization. Also, the `uic.loadUi` approach does not generate any new files -- it reads the .ui file at runtime and creates the corresponding PyQt5 objects in memory. So there is nothing to add in `.gitignore`.

alexpiet (approved):

XX-Yin:
Looks great! I initially transferred the .ui files to .py because .ui files are not human readable. Now, it turns out that we don't need to read the .ui file.

Issue #96, opened 2023-12-08
Interact with curriculum manager from the GUI


hanhou:
To close the loop of automatic training, the GUI should be able to access the `df_manager` table from AWS:

<img width="1115" alt="Screenshot 2023-12-04 at 11 11 32" src="https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training/assets/24734299/ccea0ea1-9eb0-47dd-b974-4137879b721f">

<br>Steps:
- [x] Setup AWS credentials on all training rigs (`%userprofile%/.aws/credentials`). I'll use my own credentials for now. But [AWS Identity and Access Management (IAM)](https://docs.aws.amazon.com/IAM/latest/UserGuide/introduction.html) should be a better long-term solution.
- [x] Make the [automatic training repo](https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training) pip installable. Add this dependency to the GUI.
- [x] In the GUI, create an `automatic training` module, where
  - [x] show all available curriculums.
  - [x] for the first session of a new mouse, the user selects the desired `task` and `curriculum version` from a dropdown menu with all available curriculums fetched from S3. The GUI then automatically starts from STAGE_1 (also take care of the first tens of alternating trials https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/95). 
  - [x] The GUI saves `task` - `curriculum version` info to metadata, which will be sent to S3 database together with other data. (eventually, `curriculum version`, together with the full `curriculum schema`, will be inserted to `session.json` file)
  - [x] for following sessions, the GUI fetches the suggested `STAGE`, and automatically sets the training parameters
  - [x] (optional) allow users to override the `STAGE` or change `curriculum version` during training?

Note: Being able to access S3 database is also helpful to solve [this](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/138) and other issues.

Issue #129, opened 2023-12-20
Adopt AIND-Schemas for standardization of data


alexpiet:
- [x] rig
- [x] subject
- [ ] session

alexpiet:
- rig jsons should be in repo
- SLIMS has database of instruments
-

alexpiet:
#303 partially overlapping

hanhou:
- https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/68

PR #131, opened 2023-12-21
(to han_devlop) Add curriculum manager to the GUI

hanhou:
- https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/96

A new button `Auto Train` is added
image

It pulls out all available curriculums from the S3 database and show the diagrams.
image

Not ready for testing in 447.
- [ ] Add AWS credentials to 447 PCs
- [ ] Test in 447


src/foraging_gui/Dialogs.py
```diff
@@ -9,13 +9,16 @@
...
 from PyQt5 import QtWidgets, uic
-from PyQt5.QtCore import QThreadPool,Qt
+from PyQt5.QtCore import QThreadPool,Qt, QAbstractTableModel
+from PyQt5.QtSvg import QSvgWidget
 
 from MyFunctions import Worker
 from Visualization import PlotWaterCalibration
+from aind_auto_train.curriculum_manager import CurriculumManager
 
+logger = logging.getLogger(__name__)
```
alexpiet, line 21: What is this line doing?

src/foraging_gui/Dialogs.py
```diff
@@ -1706,3 +1709,92 @@ def _KeepOpen(self):
             # change button color
             self.KeepOpen.setStyleSheet("background-color : none")
             self.KeepOpen.setChecked(False)
+
+
+class PandasModel(QAbstractTableModel):
```
alexpiet, line 1714: Can you provide a docstring for this class?

pyproject.toml
```diff
@@ -24,7 +24,8 @@ dependencies = [
     'pandas',
     'pyserial',
     "pyOSC3@git+https://github.com/glopesdev/pyosc3.git@master",
-    "newscale@git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target"
+    "newscale@git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target",
+    "aind-auto-train@git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main"
```
alexpiet, line 28: Can you also add this dependency to the requirements.txt

alexpiet (changes requested):
In addition to the specific comments
- Can you, in general, provide more documentation? Since this is just a partial PR to your branch, maybe this isn't the place for it. 
- I'd like us to move towards using standardized Schemas for different data objects. (I made a stub issue #129). It might be worth thinking now about how a curriculum or training stage might be formalized. Its probably not a blocking issue for this PR, but just raising the issue. Probably something we can review in Jan.

pyproject.toml
hanhou: added.

src/foraging_gui/Dialogs.py
hanhou: done.

src/foraging_gui/Dialogs.py
hanhou: I meant to add a logging info but I forgot. Now I added it at line 1752.

hanhou:
> In addition to the specific comments
> 
> * Can you, in general, provide more documentation? Since this is just a partial PR to your branch, maybe this isn't the place for it.
> * I'd like us to move towards using standardized Schemas for different data objects. (I made a stub issue [Adopt AIND-Schemas for standardization of data #129](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/129)). It might be worth thinking now about how a curriculum or training stage might be formalized. Its probably not a blocking issue for this PR, but just raising the issue. Probably something we can review in Jan.

Sure, once I settle on the functionality and layout of the AutoTrain part of the GUI, I'll provide step-by-step documentation for the RAs. I'll also provide more documentation for developers in my `aind-auto-train' repo on how to add more tasks and curriculums.

As for standardized schemas, I've been developing the `aind-auto-train' repo with the same idea in mind. For example, I have general purpose classes [`TaskParas`](https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training/blob/5bf9cf033a3e8a527ba47d84737afe17512f2c83/code/aind_auto_train/schema/task.py#L60) and [`Curriculum`](https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training/blob/5bf9cf033a3e8a527ba47d84737afe17512f2c83/code/aind_auto_train/schema/curriculum.py#L46) that can be inherited to define new training stages and curriculums with some degree of standardization. But let's discuss this in more detail in Jan.

src/foraging_gui/Dialogs.py
alexpiet: The thing that jumped out at me is that you are calling the logger in a different way from how I am. I just use `logging.info(<string>)`. I dont know what logging.getLogger(__name__) does. 

I'll approve the PR, because this is a minor detail. but we should be consistent

alexpiet (approved):

src/foraging_gui/Dialogs.py
hanhou: oh, I think it's better to define a logger for each module by
```python
logger = logging.getLogger(<module_name>)
logger.info('...')
```
see here https://www.codingexplorations.com/blog/best-practices-for-logging-in-python-a-comprehensive-guide

alexpiet: That site says that making a logger for each module helps customize the format, and track how messages are generated. Do you know if each module's logger inherits the format of the root logger?

I just made the root logger automatically log the module and line number

hanhou: Yes, once `logging.basicConfig` is called, all `logging.getLogger` will return logger object that inherits the root logger, i.e., streaming everything to the log file specified in basicConfig. For example, here is a log file I got:

image

The highlighted logs were from my modules with `logger = logging.getLogger`.

hanhou: btw, we can add `logging.StreamHandler()` to `handlers` in `basicConfig` to also send a copy of log info to the console (instead of `print` @XX-Yin)
```python
logging.basicConfig(
    format=log_format,
    level=logging.INFO,
    datefmt=log_datefmt,
    handlers=[logging.FileHandler(logging_filename),
              logging.StreamHandler()]
    )
```

alexpiet: awesome!

PR #132, opened 2023-12-23
(to_han_develop) Add auto training manager

hanhou:
### Describe changes:

- [x] added a GUI for the auto training table
- [x] fetch the next suggested stage, which the user can also override
- [x] load parameters from curriculum
- [x] pressing the green button automatically sets training parameters in the main window

image


hanhou:
Issue this partial PR before it becomes too big...

XX-Yin (approved):

PR #133, opened 2023-12-23
[skip email] Send a copy of log info to the terminal

hanhou:
  
### Describe changes:
Send a copy of log info to the console, if necessary.

By default, this is turned off:

image

If the field `enable_show_log_info_in_console` exists in `ForagingSettings.json` and is set to `True`, we get verbose log info in the console.

image

image

- https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/131#discussion_r1434625354

None. Only for debugging.

Not yet. It should work.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -481,7 +481,7 @@ class Window(QMainWindow):
             'newscale_port_tower2':'',
             'newscale_port_tower3':'',
             'newscale_port_tower4':'',
-            'enable_show_log_info_in_console':False,
+            'show_log_info_in_console':False,
         }
         
         # Try to load the settings file        
@@ -523,8 +523,8 @@ class Window(QMainWindow):
         self.newscale_port_tower4=Settings['newscale_port_tower4']
         
         # Also stream log info to the console if enabled
-        if ('enable_show_log_info_in_console' in Settings 
-            and Settings['enable_show_log_info_in_console']):
+        if ('show_log_info_in_console' in Settings 
+            and Settings['show_log_info_in_console']):
             logger = logging.getLogger()
             handler = logging.StreamHandler()
             # Using the same format and level as the root logger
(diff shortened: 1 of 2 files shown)


XX-Yin:
Why do we need to send all log information to the console? I only print the behavior timestamps to the console so that I can know the current behavior event. The other information is a bit distracting, and it's better to be stored in the log file.

hanhou:
> Why do we need to send all log information to the console? I only print the behavior timestamps to the console so that I can know the current behavior event. The other information is a bit distracting, and it's better to be stored in the log file.

Fair point. I wanted to add this because it's helpful during developing. I'll add a switch to the setting file then. By default, the GUI will not copy log info to the console, but it can be enabled if needed.

hanhou:
Now the log info is streamed to the console only if `enable_show_log_info_in_console` exists in `ForagingSettings.json` and is set to `True`. @XX-Yin

src/foraging_gui/Foraging.py
```diff
@@ -480,7 +480,8 @@ def _GetSettings(self):
             'newscale_port_tower1':'',
             'newscale_port_tower2':'',
             'newscale_port_tower3':'',
-            'newscale_port_tower4':''
+            'newscale_port_tower4':'',
+            'enable_show_log_info_in_console':False,
```
alexpiet, line 484: Can you make a note in the README about this additional setting?

src/foraging_gui/Foraging.py
```diff
@@ -520,6 +521,17 @@ def _GetSettings(self):
         self.newscale_port_tower2=Settings['newscale_port_tower2']
         self.newscale_port_tower3=Settings['newscale_port_tower3']
         self.newscale_port_tower4=Settings['newscale_port_tower4']
+        
+        # Also stream log info to the console if enabled
```
alexpiet, line 525: Minor - I would call it `show_log_info_in_console`, since the "enable" feels redundant with the setting being True/False

alexpiet (approved):

src/foraging_gui/Foraging.py
hanhou: fixed.

src/foraging_gui/Foraging.py
hanhou: added.

PR #144, opened 2024-01-08
[do not merge] updating gui layout

alexpiet:
  
### Describe changes:
This is a small update to the layout of the GUI. I am only moving boxes around to remove white space and condense the size of the GUI. 

- https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/102

The GUI will have a more compact layout

MicrosoftTeams-image4377fb35a2b7c24bb898d423395abf0c4e52cca4c96952b70bde167944d51163

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/ForagingGUI.ui
+++ b/src/foraging_gui/ForagingGUI.ui
@@ -12,8 +12,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1875</width>
-    <height>1032</height>
+    <width>1382</width>
+    <height>1025</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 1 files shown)


XX-Yin:
It is better to make the layout of the GUI dynamic (changeable). The smaller GUI is only preferable in the training rig where the screen is limited so far. I will prefer to use the current GUI in the ephys rig where we have enough space to show more information.

alexpiet:
@XX-Yin What do you mean by dynamic?

This update does not remove any information from the screen, it just reorganizes it to be more compact.

XX-Yin:
What I have in mind is that we could have two layouts (one for training rig and one for ephys rig). And these two layouts can be changed freely in the GUI. 

Your current layout condensed the visualization and basic information panel.

alexpiet:
@XX-Yin I see, so you are suggesting we add a button to the GUI that toggles between a condensed view (training room), and more elaborate view (ephys rigs). That seems like a good suggestion. Can you implement that? I think a streamlined GUI is a high priority for the RAs, and you are much more familiar with Qt designer than I am

hanhou:
Maybe a responsive UI design can be helpful? Compare the difference sizes of the same AutoTraining dialog:
image

alexpiet:
Yes, tabs would be good.

XX-Yin:
We can improve the GUI layout in two parts.
1) One is to switch between two layouts through buttons.
2) Another is to use tabs like Han suggested to save space.

The two approaches are not exclusive.

Yes, I can try to make some changes.

PR #151, opened 2024-01-09
[to_han_develop] Feat improve viz

hanhou:
- [x] added two buttons in the GUI to show interactive diagrams in the browser.
image
- [x] added a button showing the interactive plotly chart for all training history.
image


hanhou:
although a better way for the second point may be just linking the GUI to a page in the streamlit app.

XX-Yin (approved):

PR #158, opened 2024-01-12
Remove Tower variable from GUI window

### LINKED ISSUE #156: Naming convention changes break "Tower" index logic
Right now, when the settings are loaded, the tower is set by determining the index in the options of the GUI window for current box. The options are outdated now, so the tower index will never be set. 
- One option is to simply add the new boxes to the options in the tower index.
- alternative (my preference), we remove all logic and instances of "tower" since its only used to create folders for data saving.
- When we update the room again to use one box per computer, we can greatly simplify this logic, to just have the current_box setting in the setting files specify the full box "447-1-A"

alexpiet:
  
### Describe changes:
- Previously, the GUI window had a field "Tower", which was auto-populated from the current box to be something like "Red-1". This field was only used to save session data in folders segregated by tower, then animal. 
- The logic for setting the field was broken by recent naming convention changes. 
- I think we should remove the tower field altogether, since the data is already neatly organized by animal, and an animal will never perform a session simultaneously in to two boxes. 
- Remove the Tower field from the GUI.
- Saves the box name into the saved data
- Adds desktop shortcuts for opening the GUI with the console

- resolves #156
- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/163

- The "tower" field will no longer exist in the GUI
- Data will not be saved in folders based on tower. Data will be saved in `C:\behavior_data\<box>\<animal>\<session>\`

- [x] tested in 447


alexpiet:
@XX-Yin This is now ready for review. I have one minor question about the set_metadata_enabled

XX-Yin:
> @XX-Yin This is now ready for review. I have one minor question about the set_metadata_enabled

When we start a session, these metadata fields are not allowed to change during the session.

XX-Yin:
One small problem is that the experimenter is aligned to the right. Should be aligned to the left.
image

src/foraging_gui/Foraging.py
```diff
@@ -815,7 +815,7 @@ def _OpenBehaviorFolder(self):
         except Exception as e:
             logging.error(str(e))
             try:
-                AnimalFolder=os.path.join(self.default_saveFolder, self.Tower.currentText(),self.ID.text())
+                AnimalFolder=os.path.join(self.default_saveFolder, self.ID.text())
```
XX-Yin, line 818: I suggest we keep the original folder structure to contain the box so we can easily identify which box the session came from without opening the json file.

src/foraging_gui/Foraging.py
```diff
@@ -1702,7 +1702,7 @@ def _GetSaveFolder(self,CTrainingFolder=1,CHarpFolder=1,CVideoFolder=1,CPhotomet
         '''The new data storage structure. Each session forms an independent folder. Training data, Harp register events, video data, photometry data and ephys data are in different subfolders'''
         current_time = datetime.now()
         formatted_datetime = current_time.strftime("%Y-%m-%d_%H-%M-%S")
-        self.SessionFolder=os.path.join(self.default_saveFolder, self.Tower.currentText(),self.ID.text(), f'{self.ID.text()}_{formatted_datetime}')
+        self.SessionFolder=os.path.join(self.default_saveFolder, self.ID.text(), f'{self.ID.text()}_{formatted_datetime}')
```
XX-Yin, line 1705: I suggest we keep the original folder structure to contain the box so we can easily identify which box the session came from without opening the json file.

src/foraging_gui/Foraging.py
```diff
@@ -1741,19 +1741,19 @@ def _GetSaveFolder(self,CTrainingFolder=1,CHarpFolder=1,CVideoFolder=1,CPhotomet
 
     def _GetSaveFileName(self):
         '''Get the name of the save file. This is an old data structure and has been deprecated.'''
-        SaveFileMat = os.path.join(self.default_saveFolder, self.Tower.currentText(),self.ID.text(), f'{self.ID.text()}_{date.today()}.mat')
-        SaveFileJson= os.path.join(self.default_saveFolder, self.Tower.currentText(),self.ID.text(), f'{self.ID.text()}_{date.today()}.json')
-        SaveFileParJson= os.path.join(self.default_saveFolder, self.Tower.currentText(),self.ID.text(), f'{self.ID.text()}_{date.today()}_par.json')
+        SaveFileMat = os.path.join(self.default_saveFolder, self.ID.text(), f'{self.ID.text()}_{date.today()}.mat')
```
XX-Yin, line 1744: I suggest we keep the original folder structure to contain the box so we can easily identify which box the session came from without opening the json file.

alexpiet:
> One small problem is that the experimenter is aligned to the right. Should be aligned to the left. image

Good catch. The problem was actually that the window was too small for the default text. I made the window slightly bigger.

XX-Yin (commented):
One issue is that the box is not stored in the json file.

src/foraging_gui/Foraging.py
alexpiet: Yeah ok. I will add the current box as the folder.

alexpiet:
> One issue is that the box is not stored in the json file.

What do you mean? What json file?

XX-Yin:
> > One issue is that the box is not stored in the json file.
> 
> What do you mean? What json file?

What I mean is that when we click the "Save" button in the GUI, the name of the box should also be saved to the json file where we store the data. It is located inside the `TrainingFolder`. We can add a line in the `_Save` function of the `Foraging.py`

alexpiet:
> > > One issue is that the box is not stored in the json file.
> > 
> > 
> > What do you mean? What json file?
> 
> What I mean is that when we click the "Save" button in the GUI, the name of the box should also be saved to the json file where we store the data. It is located inside the `TrainingFolder`. We can add a line in the `_Save` function of the `Foraging.py`

Ok, that seems like something we should fix. Is that something that is particularly a concern because of this update? I guess because previously the data was organized by folders for each tower? Is the Tower saved in the json?

Should I add the folder for each box AND add the box to the json? Seems like both are valuable.

XX-Yin:
> > > > One issue is that the box is not stored in the json file.
> > > 
> > > 
> > > What do you mean? What json file?
> > 
> > 
> > What I mean is that when we click the "Save" button in the GUI, the name of the box should also be saved to the json file where we store the data. It is located inside the `TrainingFolder`. We can add a line in the `_Save` function of the `Foraging.py`
> 
> Ok, that seems like something we should fix. Is that something that is particularly a concern because of this update? I guess because previously the data was organized by folders for each tower? Is the Tower saved in the json?
> 
> Should I add the folder for each box AND add the box to the json? Seems like both are valuable.

I don't see other concerns except box name was not saved to the json file. The Tower was previously saved in the json file as the Tower field was in the GUI. 

Yes, I think both are valuable.

alexpiet:
@XX-Yin I've implemented all your suggestions

XX-Yin (approved):
Looks great!

PR #165, opened 2024-01-16
Updating Reconnect Bonsai

### LINKED ISSUE #111: check bonsai connection only establishes a connection, it does not check a previously established connection
Should have a buttons in the GUI for:
  - Check Bonsai connection. Which checks if there is a connection, and if not, connects
  - Reconnect Bonsai, which attempts to reconnect with existing bonsai. 

- Can we log the bonsai process ID, and force close it?

alexpiet:
  
### Describe changes:
- Previously "reconnect bonsai" closed the OSC connection, but would not start a bonsai instance if one was not open
- Now, "reconnect bonsai" closes the OSC connection, and when it reconnects will open a new bonsai instance if one is not running. 
- Removes the redundant function "_CheckBonsaiConnection". This function called `_ConnectBonsai`, but only if `self.InitializeBonsaiSuccessfully==0`, which is the same condition `_ConnectBonsai` uses before doing anything. I have replaced all calls to `_CheckBonsaiConnection` with `_ConnectBonsai`
- [x] 3x typos in Foraging.py. Should be `_ConnectBonsai`, not `ConnectBonsai`
- [x] Replace calls to `_CheckBonsaiConnection` in Dialogs.py

- resolves #111 

- Now if Bonsai crashes/closes, you can click on "Control", then "Reconnect Bonsai" and Bonsai will reopen
- Note that if bonsai crashes while the task is running the GUI will crash, this only helps when the task is not running. 

- [x] tested

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -485,7 +485,7 @@ class WaterCalibrationDialog(QDialog):
 
     def _StartCalibratingLeft(self):
         '''start the calibration loop of left valve'''
-        self.MainWindow._CheckBonsaiConnection()
+        self.MainWindow._ConnectBonsai()
         if self.MainWindow.InitializeBonsaiSuccessfully==0:
             return
         if self.StartCalibratingLeft.isChecked():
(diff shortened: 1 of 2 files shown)


XX-Yin (commented):
We also need to replace _CheckBonsaiConnection with ConnectBonsai In the Dialogs.py

alexpiet:
> We also need to replace _CheckBonsaiConnection with ConnectBonsai In the Dialogs.py

Fixed!

XX-Yin (approved):

PR #169, opened 2024-01-17
Fixes stability issues with bonsai, session start/stop, and two minor issues

### LINKED ISSUE #45: Relative import statements
- `Foraging.py` and `Dialogs.py` have relative import statements:
    `from Visualizations import PlotV` instead of `from foraging_gui.Visualizations import PlotV`
- relative import statements are not ideal since they can lead behavior that depends on where the function is called from. 
- To fix the code is simple, but the repository needs to be re installed in "editable" mode: `pip install -e .` instead of `pip install .`
   - To be clear, fixing this problem just requires updating the code and re-running `pip install`. However if we run `pip install -e .` then future updates won't require re-running `pip install`

### LINKED ISSUE #166: Bonsai hangs after reconnecting
This happens on `production_testing`. After using the "reconnect bonsai" function, only one trial is generated, and does not effectively start the trial loop. 

To reproduce:
- Open GUI
- hit "start", let trials run, hit "stop" (Not sure if this is necessary or not)
- click "control/reconnect bonsai"
- hit "start", only one trial will be generated

### LINKED ISSUE #167: Stop session does not add any information to the log
- Should log when the user presses the stop button

### LINKED ISSUE #168: Catch exceptions from bad connections to bonsai
I can add logic in the except hook to catch these errors. Set state variables, and alert the user. Even if they have to close the GUI, at least we can save data, and understand what is happening

alexpiet:
  
### Describe changes:
- Changes relative import statements in Foraging.py and Dialogs.py to absolute import statements
- Adds logging information about session start/stop and trial start/stop
- When Bonsai reconnects, and trials have been generated already. The user is warned to start a new session. Otherwise bonsai hangs in a loop. 
- When trials start, either at the start of a new session, or after a pause, we check if the bonsai connection is still active. If not, we stop trials, and warn the user. We specifically check for a ConnectionAbortedError. 
- Catches errors with trial initiation and alerts the user
- Alerts the user if it has been 5 minutes since the last trial started (we can adjust that duration). The use case here is bonsai disconnects, or hangs, or crashes. Or the trial generation loop gets stuck in an infinite loop. The user is prompted to either keep waiting, or stop the trials. 
- Changes the `Box with console.bat` files to stay open after the GUI closes. This is valuable if the GUI crashes and error information is displayed
- Fixes a bug with "Tower" information being requested in the NWB saving step
- Sets the title for the black console window with the current box name

- resolves #45 
- resolves #167
- resolves #166 
- resolves #168

- The GUI should be more stable if Bonsai crashes. 
- If Bonsai crashes when trials are not running, click "Control" then "reconnect Bonsai", then "new session"
- If Bonsai crashes when trials are running, a timeout window will popup and let you force stop the session. Then you can save the session, and then reconnect to bonsai with "Control", then "reconnect Bonsai"

- [x] tested in 447
- [x] Updated the startup .bat files in 447


src/foraging_gui/Foraging.py
```diff
@@ -2307,7 +2333,22 @@ def _StartTrialLoop(self,GeneratedTrials,worker1):
                     # get reward and generate a new trial
                     self.NewTrialRewardOrder=0     
                 #initiate the generated trial
-                GeneratedTrials._InitiateATrial(self.Channel,self.Channel4)
+                try:
+                    GeneratedTrials._InitiateATrial(self.Channel,self.Channel4)
+                except Exception as e:
```
XX-Yin, line 2338: There are other possibilities that you got an error from GeneratedTrials._InitiateATrial(self.Channel,self.Channel4)

src/foraging_gui/Foraging.py
```diff
@@ -392,7 +392,10 @@ def _ReconnectBonsai(self):
         logging.info('attempting to restart bonsai')
         self.InitializeBonsaiSuccessfully=0       
         self._InitializeBonsai()
+        if self.InitializeBonsaiSuccessfully ==1 and hasattr(self, 'GeneratedTrials'):
```
XX-Yin, line 395: What is this doing?

src/foraging_gui/Foraging.py
```diff
@@ -2191,26 +2206,34 @@ def _Start(self):
...
-        if self.StartANewSession==1 and self.ANewTrial==0:
-            self.WarningLabel.setText('Waiting for the finish of the last trial!')
-            self.WarningLabel.setStyleSheet("color: red;")
-            while 1:
-                QApplication.processEvents()
-                if self.ANewTrial==1:
-                    self.WarningLabel.setText('')
-                    self.WarningLabel.setStyleSheet("color: red;")
-                    break
+            self._StopCurrentSession()
```
XX-Yin, line 2213: There is another constraint "self.StartANewSession==1" that is not contained in the self._StopCurrentSession(). We don't need to wait for the finish of the current trial if the Newsession button is not pressed. For example, the Start button is wrongly unpressed and we hope to press it again. 

I suggest we keep the original code.

src/foraging_gui/Foraging.py
alexpiet: You are correct! I've updated this so it now checks for "ConnectionAbortedError".

src/foraging_gui/Foraging.py
alexpiet: I added a comment, but its not showing up in the Files Changed viewer. 

There is a bug #166, where if trials have been generated, and then bonsai reconnects, when new trials are generated the trial loop hangs. I cannot resolve the bug, so I am warning users to start a new session.

src/foraging_gui/Foraging.py
alexpiet: I'm not sure I understand your concern.
All "StopCurrentSession" does is toggle the start button color, and wait for the last trial to finish. This is exactly the logic that was implemented if the Start button was unchecked. 

If the user presses "start" on accident, they can still press "Start" again to resume trials. This code update doesn't impact that.

alexpiet:
@XX-Yin This is ready for you to review again, I've made a lot of changes.

XX-Yin:
> @XX-Yin This is ready for you to review again, I've made a lot of changes.
> 
> One question for you: What should the time-out duration be? Right now I set it to 5 minutes without the trial ending.

5m is good. We will not have trial longer than that

src/foraging_gui/Foraging.py
XX-Yin: If there is no constraint "self.StartANewSession==1", when you click the "Start" button when "self.ANewTrial==1", you will get an error like below:

MicrosoftTeams-image (1)

XX-Yin: We still need to add constraint "self.StartANewSession==1. 

The other modifications look good!

alexpiet:
@XX-Yin identified a bug in my code where if the user hits the start/stop button many times the "StopCurrentSession" process will eat up memory the GUI will crash. To be clear this will only happen if the user is spamming the button 100+ times. We stopped this by only triggering the "StopCurrentSession" process when we want to start a new session. 

@XX-Yin I tested our update, and everything looks good. If you approve I will merge this in after the weekly production merge

PR #172, opened 2024-01-19
Major refactor of the GUI

### LINKED ISSUE #170: Reorganize GUI
- See this discussion https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/102

hanhou:
### Describe changes:
Reorganized the main GUI and the AutoTrain GUI.
- Now it's much easier to control multiple GUIs with one screen
- The GUI is now responsive, i.e., supports dynamic scaling
- Tabs and scrollable areas make button placement and information display much cleaner and more streamlined.

This screenshot shows two GUIs running side-by-side, leaving room for camera views in the bottom. Note they are on different "tabs".

image

And the user can freely resize the GUI without blocking important information and control buttons

image

Together with the updated AutoTrain dialog

image

- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/102
- resolves #170 
- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/128
- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/129 

- Although there is no change in functionality, experimenters may need some time to familiarize themselves with the new configuration
- Bug reports and feedback from the experimenters are important.

- [x] tested on my PC
- [x] test in 447-2-D with Travis


alexpiet:
This looks great! Thanks for doing the detailed work. Lets get @XX-Yin's approval before we merge, since he will also be using the GUI on the ephys computer

ellahiltonvano:
This looks great, Han, thank you so much! We will provide feedback as we begin to use it, but I already think it will be much more functional!

avalon-amaya:
This is amazing thank you!

XX-Yin:
These changes look great for use in the training box. However, it is better to access most of the information (training parameters, stage control, optogenetics, visualization panel...) from the main window, since we need to check them frequently. We also have a screen dedicated to behavior in the ephys room. We should figure out how to reconcile these two configurations (ephys and training). Sorry, I'm unable to approve the pull request at this time even it's great for behavior training as it will take more time to make an ephys version based on it.

hanhou:
Since eventually we don't have to (and should not) change any training parameters or lickport position during ephys recording, I don't see a long term need for an ephys-specific version that has all information in one page like before. Even if you have some ephys-specific "plugins", it would be still better to build upon the new responsive and tab design. So for now, I would suggest we merge this PR to enhance the user experience in the behavior room first. Then you could take your time to adapt it for the ephys rig if necessary (by loading rig-specific ui etc).

alexpiet:
@XX-Yin @hanhou Lets discuss this in person on monday. I think maintaining to versions of the GUI for ephys and behavior is going to be a huge headache. I bet we can find a streamlined version of the GUI that works for both.

XX-Yin:
Yes, let's discuss on Monday. One solution is to maintain two ForagingGUI.ui (one for training rig and one for ephys rig). We can use one of them when we open the GUI.

hanhou:
Tested the new GUI on Rig 2D with Travis.

hanhou:
I have tested this new GUI in 447-2-D. Per discussion #178, I'm merging this to `production_test` now

alexpiet:
Oh, I thought the idea was to merge it tomorrow after the production merge, so it gets tested next monday. I guess having Travis test it today is sufficient.

hanhou:
Oh sorry I messed up with the standard procedure. We should not merge anything to production branch until the last one is merged to main, right? Should I reverse it?

alexpiet:
Yes, the standard procedure would be to not merge things to production_testing between monday morning (when the testing rig updates to production_testing) and tuesday (after the production_merge). 

If you want to be a super rule-follower you can reverse it. But since travis tested it with a live mouse today, and we are not quite back to normal procedure, I think its OK to keep the PR as - is. Your call

hanhou:
Okay, I'll keep the PR as it is this time and will not merge anything between Monday morning and Tuesday in the future.

Issue #146, opened 2024-01-08
Handle the first session in auto training


hanhou:
As we previously discussed, the with-session transition from stage 1.1 (tens of L-R-L-R alternating trials) to stage 1.2 (real foraging) should be handled locally in the GUI. This is one of the only two missing parts in our automated training system (https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/73).

Here I propose a more general way of doing this. Instead of hardcoding the L-R-L-R part and defining a transition rule to 1.2 for **_only the first session_**, we can in fact add a general `warm-up` option into the `task schema`, i.e., add following parameters to the GUI:
- `warmup_enabled: bool = False`: a switch to turn on/off the warm up section
- `warmup_block_length: int = 1`: the length of block during warm-up
- `warmup_reward_probs: list = [0, 1]`: the reward probs during warm-up
- `warmup_min_trials: int = 30`: the minimum length of warm-up
- `warmup_max_choice_ratio_bias: float = 0.1`: the maximum deviation from a `choice ratio` of 0.5 (no bias)
- `warmup_min_finish_ratio: float = 0.7`: the minimum of the `finished ratio` (finished trials / total trials)
- `warmup_trials_to_compute_ratios: int = 20`: the length of the sliding window by which we compute the two ratios

The warmup is always simple left-right alternating blocks with `warmup_block_length` and `warmup_reward_probs`, and we always use this simple criterion to end the warmup:

```
if (
number of warmup trials >= `warmup_min_trials`
and for recent `warmup_trials_to_compute_ratios` trials,
    finished ratio >= `warmup_min_finish_ratio`
    and 
    abs(proportion of choosing Right - 0.5) <= `warmup_max_choice_ratio_bias`
)
```

With this design, we could just turn `warmup_enabled` on for the first session to implement stage 1.1 -> stage 1.2. It also makes it possible if one would like to include a warm-up to not only the first session (presumably with larger warmup_block_length and lower warmup_reward_probs). Most importantly, everything is defined and tracked under the same task schema, and the first session is no longer special.

@XX-Yin @alexpiet what do you think? If this makes sense, Xinxin, could you implement this interface in the GUI?

hanhou:
btw, we could implement the within-session automated lickport movement in the same way, which is the other one of the only two missing pieces.

alexpiet:
The most general solution would allow people to let trial structure change throughout the session. 

[...]

PR #176, opened 2024-01-21
Added the auto train to the Control

XX-Yin:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "production_testing" into "main" should use the keyword "production merge" in the title for reliable indexing of updates
  
### Describe changes:
1. Added the AutoTrain button to the Control so that we can access Auto Train from the Control list in the menu bar.
2. Moved the shortcut flag to the Auto Train in the Control. 

### What issues or discussions does this update address?
Resolves #175 

### Describe the expected change in behavior from the perspective of the experimenter
No

### Describe the outcome of testing this update on a rig in 447
Tested on my own computer.

PR #177, opened 2024-01-21
Handle the first session of automatic training; support behavior and ephys ui

XX-Yin:
  
### Describe changes:
1. Added warm up to handle the first session of auto training. 
2. Make the current GUI compatible with behavior and ephys ui. It's controled by `default_ui` in the `ForagingSettings.json`. It will use the behavior ui by default. Set it to ForagingGUI_Ephys.ui if you want to use the ephys version. 

image
image

Resolves [issue ](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/146#issuecomment-1897200865)
Resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/178#event-11560569637

No

- [x] test in my own computer
- [x] test in 447
- [x] test in ephys rig


XX-Yin:
@hanhou The warmup and related parameters (warm_min_trial;warm_min_finish_ratio;warm_max_choice_ratio_bias;warm_windowsize) should be added to the curriculum. At least , we should add the warmup. The warmup should be set to "on" in the first stage.

src/foraging_gui/Foraging.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, parent=None,box_number=1):
         self._LickSta()
         self._InitializeMotorStage()
         self._StageSerialNum()
+        self._warmup()
```
alexpiet, line 112: The warmup period is separate from the actual session?

src/workflows/foraging.bonsai
```diff
@@ -1324,7 +1324,7 @@
                       <harp:VisualIndicators>On</harp:VisualIndicators>
                       <harp:Heartbeat>Disabled</harp:Heartbeat>
                       <harp:IgnoreErrors>false</harp:IgnoreErrors>
-                      <harp:PortName>COM14</harp:PortName>
+                      <harp:PortName>COM25</harp:PortName>
```
alexpiet, line 1327: why is this changing?

hanhou:
> @hanhou The warmup and related parameters (warm_min_trial;warm_min_finish_ratio;warm_max_choice_ratio_bias;warm_windowsize) should be added to the curriculum. At least , we should add the warmup. The warmup should be set to "on" in the first stage.

@XX-Yin and @hanhou had a brief discussion this morning, and we agreed that:
1. The warmup part is agnostic to how the parameters are set (either by AutoTrain, old stage setting, or manually). Hence, the code for warmup does not need to interact with AutoTrain.
2. The warmup part share all other parameters with the following actual training part except these warmup-specific parameters such as `warmup_block_length` and `warmup_reward_probs`, as described at #146 .
3. As described at #146, warmup could be turned on in any training stage as a general bias-fighting measure.

hanhou:
Similiar to #176, @XX-Yin please also apply the UI change to #172.

src/workflows/foraging.bonsai
XX-Yin: This is for my computer. It's dependent on Settings_box1.csv and will not affect running on the training rig.

XX-Yin:
> Similiar to #176, @XX-Yin please also apply the UI change to #172, once you're done with the warmup code.

Sure!

src/workflows/foraging.bonsai
hanhou: In this case I would recommend not adding bonsai files to git's staged changes when you commit, which is what I typically do.

src/foraging_gui/Foraging.py
XX-Yin: The warmup is dependent on the actual session. The logic is when the warmup is turned on, we will use specific parameters for some fields (parameters in the `Warm up:`, `Advanced block:`, `Block:` , `Base reward probability:` and `Auto water:` ) and leave other parameters related to that session unchanged. When warmup is turned off, the parameters will return to the state before warmup was turned on.

src/workflows/foraging.bonsai
XX-Yin: Sounds good!

src/foraging_gui/Foraging.py
```diff
@@ -204,6 +204,70 @@ def connectSignalsSlots(self):
...
+    def _warmup(self):
+        '''warm up the session before starting.
+            Use warm up with caution. Usually, it is only used for the first time training. 
+            Turn on the warm up only when all parameters are set correctly, otherwise it would revert 
+            to some incorrect parameters when it was turned off.
+        '''
+        # set warm up parameters
+        if self.warmup.currentText()=='on':
+            # get parameters before the warm up is on
+            self._GetTrainingParameters(prefix='WB_')
```
hanhou, line 216: What does `WB_` mean? WarmupBackup? I suggest we use clearer naming conventions to improve readability.

XX-Yin: Yes, it represents WarmupBackup or WarmupBefore. I make a note to explain the abbreviation and changed the prefix to `WarmupBackup_`

src/foraging_gui/Foraging.py
```diff
@@ -204,6 +204,70 @@ def connectSignalsSlots(self):
...
+            self.BlockMin.setText('1')
+            self.BlockMax.setText('1')
+            self.BlockMinReward.setText('1')
+
+            self.AutoReward.setChecked(True)
+            self._AutoReward()
+            self.AutoWaterType.setCurrentIndex(self.AutoWaterType.findText('Natural'))
+            self.Multiplier.setText('0.8')
+            self.Unrewarded.setText('0')
+            self.Ignored.setText('0')
```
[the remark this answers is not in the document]
XX-Yin: When we turn on the warmup, we set some parameters to default. However, I also keep the flexibility to also change other parameters, which means we need to recover other training parameters. 

To ensure the safety, I don't load the `warmup` from the json file.  And the `warmup` should be turned on when all parameters are set from the autotrain.

src/foraging_gui/Foraging.py
```diff
@@ -857,72 +921,81 @@ def _TrainingStage(self):
...
-                            if Tag==0:
-                                widget.setChecked(bool(value[-1]))
-                            elif Tag==1:
-                                widget.setChecked(value)
-                            self._AutoReward()
-                else:
-                    widget = widget_dict[key]
-                    if not (isinstance(widget, QtWidgets.QComboBox) or isinstance(widget, QtWidgets.QPushButton)):
-                        widget.clear()
+                self._set_parameters(key,widget_dict,self.TrainingStagePar[Task][CurrentTrainingStage])
```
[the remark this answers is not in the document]
XX-Yin: I use three parameters `key`, `widget_dict` and `parameters` here, and use a for loop to iterate through all the keys on purpose. In this way, I can exclude some keys from the widget_dict. There are multiple ways to do it. It's more relevant to encapsulation of for loops before or after.

The training stage is still useful even the training parameters are determined by autotrain. In some cases, we still need to explore some parameters in experiments, and we can use the training stage parameters. 

In autotrain, you have frozen the training phase parameters, and it is safe.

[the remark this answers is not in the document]
XX-Yin: Regarding point 4, I agree with you. I probably won't change it here, but will in a future update.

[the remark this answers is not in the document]
hanhou: > I use three parameters `key`, `widget_dict` and `parameters` here, and use a for loop to iterate through all the keys on purpose. In this way, I can exclude some keys from the widget_dict. There are multiple ways to do it. It's more relevant to encapsulation of for loops before or after.

I know your code works. I'm just proposing maybe a more intuitive and clear way of encapsulating the `set_parameter` function. It's not about whether you use for loops before or after, but about which list you apply the for loops on. I suggest you run the for loop on `para_dict`, where you explicitly list what parameters you want to change, instead of looping over all widgets in the `TrainingParameters` Qt Group and exclude those parameters you don't want. I know both ways do the same job, but the former looks more reusable, concise, and robust. For example, as I mentioned, the way you do a for loop over the children of `TrainingParameters` makes your code depend on the GUI configuration, which sounds less robust to me.

> The training stage is still useful even the training parameters are determined by autotrain. In some cases, we still need to explore some parameters in experiments, and we can use the training stage parameters.
> In autotrain, you have frozen the training phase parameters, and it is safe.

In my suggestion, I did not exclude the use of the training stage:
>This dictionary could come from a full task schema (like your _**self.TrainingStagePar[Task][CurrentTrainingStage]**_ and my automatic training stage)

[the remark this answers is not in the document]
XX-Yin: I think that you are not satisfied with the internal organization of `_set_parameters` instead of encapsulation of for loops outside or in the `_set_parameters`. Let's optimize it in a separate PR. There are a lot of stuff going on in this branch.

src/foraging_gui/Foraging.py
```diff
@@ -2699,8 +2699,8 @@ def _UpdateSuggestedWater(self,ManualWater=0):
                 # maximum 3.5ml
                 if suggested_water>3.5:
                     suggested_water=3.5
-                    self.TotalWaterWarning.setText('Supplemental water is >3.5! Health issue and LAS should be alerted!')
-                    self.TotalWaterWarning.setStyleSheet(self.default_color)
+                    self.TotalWaterWarning.setText('Supplemental water is >3.5! Health issue and \n LAS should be alerted!')
```
hanhou, line 2702: Please use an if-else condition here to keep the previous text unchanged for behavior rigs (we don't need the "\n" for the behavior GUI).

hanhou:
It looks like a lot of the recent changes are not about handling the first session, but about making the code compatible with the two versions of GUIs. Also, the PR is getting really big. Is there any chance to split the PR into different parts?

src/foraging_gui/Foraging.py
```diff
@@ -264,7 +264,7 @@ def _revert_to_previous_parameters(self):
         parameters={}
         for attr_name in dir(self):
             if attr_name.startswith('WarmupBackup_') and attr_name!='WarmupBackup_' and attr_name!='WarmupBackup_warmup':
-                parameters[attr_name[3:]]=getattr(self,attr_name)
+                parameters[attr_name[13:]]=getattr(self,attr_name)
```
hanhou, line 283: Useing `attr_name.strip('WarmupBackup_')` is better than hardcoding `13` here. Otherwise the code breaks if one change the prefix.

XX-Yin:
> It looks like a lot of the recent changes are not about handling the first session, but about making the code compatible with the two versions of GUIs. Also, the PR is getting really big. Is there any chance to split the PR into different parts?

Yes, there are two parts of the PR (listed below). It's possible to split them into two PRs, but it requires a lot of additional work. I don't find it's worth doing right now.  

I developed the first part based on the previous ui, and production testing already used the new ui, so I developed the second part. This update handles both ui and it only adds minimal additional work to maintain them. We just need to copy the new widget from one ui to another.

1. Handle the first session. 
2. Support the two ui.

alexpiet:
I'd like to say again, for the record, that I think maintaining two `.ui` files is a bad idea.

src/foraging_gui/Foraging.py
XX-Yin: Done!

src/foraging_gui/Foraging.py
XX-Yin: I have updated it.

XX-Yin:
> I'd like to say again, for the record, that I think maintaining two `.ui` files is a bad idea.

We only need to copy the widget from one ui to the other ui given they share the same name. I think it's worth it given the necessity.

src/foraging_gui/Foraging.py
XX-Yin: We can't use `attr_name.strip('WarmupBackup_')` here. It should be attr_name.replace('WarmupBackup_','') instead. See below:
 
image

hanhou: Oh yes, my bad.

XX-Yin:
I tested this update in the 447 and ephys rig, and it's ready for review.

hanhou (commented):
I did some simulations and seems like I found a bug. 

image

In this example, from 1600 to 3500 sec, the warmup is turned off, but the block is not updated (still [1.0, 0.0]). Could you confirm whether this is a bug?

XX-Yin:
> I did some simulations and seems like I found a bug.
> 
> image
> 
> In this example, from 1600 to 3500 sec, the warmup is turned off, but the block is not updated (still [1.0, 0.0]). Could you confirm whether this is a bug?

Can you send me the parameters you were using? It seems that the Advanced Block is on.

hanhou:
> > I did some simulations and seems like I found a bug.
> > image
> > In this example, from 1600 to 3500 sec, the warmup is turned off, but the block is not updated (still [1.0, 0.0]). Could you confirm whether this is a bug?
> 
> Can you send me the parameters you were using? It seems that the Advanced Block is on.

I first select Stage 1.2, and turn the warm up ON, then start the session

image

I can confirm that when warmup is on, Advanced Block is turned off automatically. But when it ends, Advanced Block is turned on again because it is on in Stage 1.2. That's why you see the block is frozen between 1600 to 2500 sec. However, since warmup is already off during that time, the reward probability should be updated to [0.8, 0.0].

XX-Yin:
> > > I did some simulations and seems like I found a bug.
> > > image
> > > In this example, from 1600 to 3500 sec, the warmup is turned off, but the block is not updated (still [1.0, 0.0]). Could you confirm whether this is a bug?
> > 
> > 
> > Can you send me the parameters you were using? It seems that the Advanced Block is on.
> 
> I first select Stage 1.2, and turn the warm up ON, then start the session
> 
> image
> 
> I can confirm that when warmup is on, Advanced Block is turned off automatically. When it ends, Advanced Block is turned on again because it is on in Stage 1.2. That's why you see the block is frozen between 1600 to 2500 sec. However, since warmup is already off during that time, the reward probability should be updated to [0.8, 0.0], starting from 1600 sec.

The block parameters ([0.8 0.0]) will take effective in the next block. The logic here is 1) We went to a new trial. 2) Turned off the warmup. 3) Parameter changed. 4) The block parameters will not change as it's still in the current block. 

If we would like to go to next block immediately after stopping of warmup, we could trigger the NextBlock.

hanhou:
> > > > I did some simulations and seems like I found a bug.
> > > > image
> > > > In this example, from 1600 to 3500 sec, the warmup is turned off, but the block is not updated (still [1.0, 0.0]). Could you confirm whether this is a bug?
> > > 
> > > 
> > > Can you send me the parameters you were using? It seems that the Advanced Block is on.
> > 
> > 
> > I first select Stage 1.2, and turn the warm up ON, then start the session
> > image
> > I can confirm that when warmup is on, Advanced Block is turned off automatically. When it ends, Advanced Block is turned on again because it is on in Stage 1.2. That's why you see the block is frozen between 1600 to 2500 sec. However, since warmup is already off during that time, the reward probability should be updated to [0.8, 0.0], starting from 1600 sec.
> 
> The block parameters ([0.8 0.0]) will take effective in the next block. The logic here is 1) We went to a new trial. 2) Turned off the warmup. 3) Parameter changed. 4) The block parameters will not change as it's still in the current block.
> 
> If we would like to go to next block immediately after stopping of warmup, we could trigger the NextBlock.

I would presume "warm up finished" always means going back to real training and moving to the next block immediately, right? If you update some of the parameters (like advanced block) but don't update the reward probability, you ends up with a block that does not make any sense.

hanhou (commented):
And any clue why this time the warm up never ends?

image

XX-Yin:
> And any clue why this time the warm up never ends?
> 
> image

What is the simulation time methods here?

hanhou:
> > And any clue why this time the warm up never ends?
> > image
> 
> What is the simulation time methods here?

Win-stay-lose-shift. But it should not depend on the simulation method, right?

XX-Yin:
> I also have two suggestions:
> 
> 1. add some indications to the GUI showing the status of warmup. For example, in the Warning section or the Information section, show messages like this
>    
>    * "Warm up started: trials 15 (x), finished ratio 0.9 (pass), bias 0.8 (x)"
>    * "Warm up started: trials 30 (pass), finished ratio 0.9 (pass), bias 0.7 (x)"
>    * "Warm up finished!"
> 2. Now you change the "warm up" button to "off" when warmup ends. I would suggest we keep that button as "ON" because "warm up = ON" is a part of the training parameter of this session. In other words, finishing a warmup does not mean that we changing the training parameter. Instead, we can use an internal flag like `self.warmup_is_running` to track whether warmup is still ongoing.

They are good points. I will add these indicators. 

After changing the warmup, the training parameters will also change. However, these parameters are interdependent.

hanhou:
- For the record, I'm approving this PR per this dicussion https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/207#issuecomment-1917663223. The issues about warmup will be solved by another PR.

hanhou (approved):

PR #181, opened 2024-01-23
Fixed a crash bug

XX-Yin:
  
### Describe changes:
I commented some lines related to a recent update. Specifically, the error was happening on line 1087 in the foraging.py (`old = getattr(Parameters,'TP_'+child.objectName())`) when we tried to change the baseline weight after stopping the GUI. 

Sue will submit an issue to describe the error information in more detail. 
No
Tested in the ephys rig.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1062,6 +1062,7 @@ class Window(QMainWindow):
                             child.setValue(int(getattr(Parameters, 'TP_'+child.objectName())))
                         else:
                             child.setText(getattr(Parameters, 'TP_'+child.objectName()))
+                    '''
                     else:
                         # If this parameter changed, add the change to the log
                         old = getattr(Parameters,'TP_'+child.objectName())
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -1070,7 +1071,8 @@ def keyPressEvent(self, event=None):
                         new = float(child.text())
                         if new != old:
                             logging.info('Changing parameter: {}, {} -> {}'.format(child.objectName(), old,new))
-
+                    '''
+
```
alexpiet, line 1075: Your update removes the logging of the changed parameters. Instead of commented out the code, can we fix the problem?

XX-Yin: Yes, we need to fix it. I haven't explored which field caused this crash, so I just commented them temporarily for using in the ephys rig.

alexpiet: Can you post the output of the GUI log?

alexpiet (approved):

alexpiet:
@XX-Yin We can merge this in today so the rigs are stable, and then I will fix the underlying bug.

PR #187, opened 2024-01-23
Run GUI without bonsai IDE

alexpiet:
  
### Describe changes:
- adds a keyword argument to run the GUI without opening the bonsai IDE. The bonsai workflow is still running in the background. 
   - The OSC channels are unstable when bonsai starts without the IDE, so I force reconnect at the end of the startup sequence
- Adds desktop shortcuts that use the new keyword argument

- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/85)

- The bonsai window will not appear, but bonsai is running in the background
- If the GUI states "please open bonsai", then click on "reconnect bonsai"

- [x] tested in 447
- [ ] setup desktop shortcuts in 447
- [ ] feedback from RAs


alexpiet:

src/desktop_shortcuts/start_popup.bat
```diff
@@ -0,0 +1,24 @@
...
+echo                      {=  ^^  =}
+echo                       ^>  -  ^<
+echo        ___________.""`-------`"".____________
+echo       /  Hi, the GUI is starting!     O      \
+echo       \                      o               /
+echo       /  .    O       Please wait!       o   \
+echo       \                                      /         __
+echo       /        You can close this window!    \     _.-'  `.
+echo       \______________o__________o____________/ .-~^^        `~--'
+echo                     ___)( )(___        `-.___.'
```
hanhou, line 19: This looks awesome! How did you make it?

hanhou (approved):

src/foraging_gui/Foraging.py
```diff
@@ -40,7 +40,7 @@ def default(self, obj):
         return super(NumpyEncoder, self).default(obj)
     
 class Window(QMainWindow):
-    def __init__(self, parent=None,box_number=1):
+    def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
```
XX-Yin, line 43: Here, "start_bonsai_ide=True" is redundant with "start_bonsai_ide=True" in "if __name__ == "__main__":"?

src/desktop_shortcuts/start_popup.bat
alexpiet: I just googled "mouse ascii art" and modified one of the results

src/foraging_gui/Foraging.py
alexpiet: Yes, I set the default value here, in the class definition, and define the value in the main function. To be clear, we do the same thing with `box_number`. To me, its better to be explicit. But I can remove the values in one place if you like

XX-Yin: I think it's better to delete `start_bonsai_ide=True` from the `__init__(self, parent=None,box_number=1,start_bonsai_ide=True)` as it will be `True` by default even it's set to be `False`.

hanhou: > it will be `True` by default even it's set to be `False`.

It will be `true` _**only if**_ it is not set when calling the function (the meaning of "by default"). I would leave it there for clarity. Having a default value is also a good practice for maintaining backward compatibility when you add an argument to an existing function.

XX-Yin: Yeah, either way is ok. 

There is always a default input start_bonsai_ide=True [here](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/187/files/8ede33d9b96b497ee105def0aa9b95bb958a13ae#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR2763), so I suggest delete the `=True` from the `__init__(self, parent=None,box_number=1,start_bonsai_ide=True)`

PR #214, opened 2024-02-01
Add notification for session stopping due to ignore trials, max trials, or max time

alexpiet:
  
### Describe changes:
- When the session stops generating trials due to any reason other than the user pressing "start", a notification window is generated alerting the user that the session has stopped and the reason. The user presses "OK". 
- The title of the window is "Box <box number>" to clarify which session has stopped.
- Cleans up the logic for checking for session ends
- Identifies and fixes a bug. If the response history is greater than StopIgnore, then the other stopping conditions will never be met. 

- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/237

When the session stops generating trials due to any reason other than the user pressing "start", a notification window is generated alerting the user that the session has stopped and the reason. The user presses "OK". 

- tested in 447

### DIFF
--- a/src/foraging_gui/MyFunctions.py
+++ b/src/foraging_gui/MyFunctions.py
@@ -14,6 +14,7 @@ from PyQt5 import QtWidgets
 from PyQt5 import QtCore
 
 
+
 if PLATFORM == 'win32':
     from newscale.usbxpress import USBXpressLib, USBXpressDevice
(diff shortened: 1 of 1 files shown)


alexpiet:
@hanhou @XX-Yin this is ready for review

XX-Yin (approved):
Looks great!

Issue #213, opened 2024-02-01
Fix warmup


hanhou:
- For the record, I'm approving this PR per this dicussion https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/207#issuecomment-1917663223. The issues about warmup will be solved by another PR.

_Originally posted by @hanhou in https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/177#issuecomment-1917698175_

hanhou:
@XX-Yin let's fix warmup bugs and test it thoroughly so that it can be integrated into autotrain.

XX-Yin:
@hanhou Yes, I will find the time sooner! I've been super busy this week and barely had a long stretch of time to work on code.

hanhou:
No problem!

PR #234, opened 2024-02-06
Finished warm up

XX-Yin:
  
### Describe changes:
1) Added warning to show metrics and state related to warm up. 
2) Fixed a bug related to the finish length. 

Resolves: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/213#issuecomment-1921951325
Resolves: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/177#issuecomment-1917698175

If the warmup is turned on (typically for the first session of a mouse), the session starts with previous Coupled-Baiting Stage 1.1

image

After warmup termination criteria are met, goes back to normal training, i.e., parameters before warmup
image

 

Tested on my own computer.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2407,6 +2407,7 @@ class Window(QMainWindow):
             # generate a new session id
             self.WarningLabel.setText('')
             self.WarningLabel.setStyleSheet("color: gray;")
+            self.win.WarmupWarning.setText('')
             # start a new logging
             try:
                 self.Ot_log_folder=self._restartlogging()
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -256,6 +256,7 @@ class Window(QMainWindow):
             self.label_118.setEnabled(True)
 
             # set warm up default parameters
+            self.Task.setCurrentIndex(self.Task.findText('Coupled Baiting'))
             self.BaseRewardSum.setText('1')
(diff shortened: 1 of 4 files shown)


XX-Yin:
This is the warning.
image

hanhou:
It looks great now!

hanhou (approved):

hanhou:
Just notice another bug: we should cache the `Task` before warmup. During warmup, task should be set to "Coupled Baiting". After warmup, we should then recover `Task` as well.

hanhou (changes requested):
please also cache, set, and restore `Task`

XX-Yin:
> please also cache, set, and restore `Task`

Very good catch! I updated it!

hanhou (approved):
It looks good now. The only remaining things in my [wishlist](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/146#issue-2071349341) is making `warmup_block_length` and `warmup_reward_probs` adjustable so that it could be a general-purpose warmup part, not just for the first session. But it is good enough for now.

XX-Yin:
> It looks good now. The only remaining things in my [wishlist](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/146#issue-2071349341) is making `warmup_block_length` and `warmup_reward_probs` adjustable so that it could be a general-purpose warmup part, not just for the first session. But it is good enough for now.

Yes, it's fixed. We can make changes if we need it in the future.

Issue #199, opened 2024-01-25
do newscale motor updates get logged?


alexpiet:
They should, so we can debug issues

Issue #237, opened 2024-02-07
Multiple Newscale errors


alexpiet:
Making a consolidated issue for uncaught newscale errors:
- [ ] https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/243
- [ ] https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/228
- [ ] https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/199
- [ ] https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/157

alexpiet:
Closing as redundant with #243

PR #240, opened 2024-02-08
Streamlined newscale control logic

alexpiet:
  
### Describe changes:
- I reorganized the code that connects to the newscale stages. I was able to greatly simplify the code, and add error checks for various failure modes. 
- I removed the "Scan stages" function from the drop down menu (in both .ui files)
- I removed the "Stage" option box from the "Motor Stage" panel (in both .ui files). 
- I added "GetPosition()" to the startup sequence, so the newscale position is already loaded without a manual press of the Get positions button. 
- I stopped forcing a disconnect of the newscale instances, because it was alreadys throwing an SI_INVALID_HANDLE error
- I added more logging statements so we can understand when the newscale stage doesn't connect
- I removed the MotorStage and Manipulator dialogs, since all they did were open blank windows. 
- removed the icons for manipulator and motor stages

- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/157
- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/199
- partially addresses https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/237

- You will no longer see a box for "Stage: "
- The stage positions will automatically load at startup
MicrosoftTeams-image5a1c0981fa016a3762a7f06ed8c8e206cdf53407cc7fc015e59f67bcca0679fe

- tested in 447

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -1153,10 +1153,10 @@ def is_file_in_use(file_path):
         except OSError as e:
             return True
 
-class ManipulatorDialog(QDialog):
-    def __init__(self, MainWindow, parent=None):
-        super().__init__(parent)
-        uic.loadUi('Manipulator.ui', self)
+#class ManipulatorDialog(QDialog):
+#    def __init__(self, MainWindow, parent=None):
+#        super().__init__(parent)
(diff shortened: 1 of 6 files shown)


alexpiet:

XX-Yin (approved):
Looks good!

Issue #191, opened 2024-01-24
Cannot convert input to float: , ''


alexpiet:
This appears to be the root cause of #183. #185 solves this by checking for the attribute, but we should look for the empty widget in the UI file and remove it

XX-Yin:
@alexpiet thanks for checking. I am not sure that I understand this issue. I checked the Frequency in the json file, and it was set correctly. 

What do you mean this parameter does not get set of updated correctly?

XX-Yin:
The frequency of each condition is defined as Frequency_1, Frequency_2, Frequency_3 and Frequency_4 respectively.

alexpiet:
@XX-Yin Sorry for my lack of clarity. I was worried that the Frequency parameters weren't getting saved when altered in the GUI, but I checked and they are. 

The issue is that the individual options within the QCombobox are QLineEdit widgets, so the code that checks for parameter changes iterates through them, but they don't have name fields, so we see the empty widgets. I have a PR to remove the error in this case. 

Everything seems to be working fine! Sorry for the false alarm

alexpiet:
Probably also the cause of this:
15 | Cannot convert input to float: , '' | Foraging.py | keyPressEvent | line 1685 | ephys3

alexpiet:
Happening often.
245	Cannot convert input to float: , ''	Foraging.py	keyPressEvent	line 1765	1B, 1C, 1D, 2A, 2B, 3A, 3C, 3D, 7C, ephys3

PR #241, opened 2024-02-09
Cleaned up parameter change logs

alexpiet:
  
### Describe changes:
- The GUI log had multiple errors due to either the parameter name being the empty string, or the text of the parameter being the empty string. I added checks in the code that processes parameter changes to handle these case. 
- I removed the "AnimalName" and "WeightBefore" fields in the parameter processing code since they are outdated and not used. 
- The log was throwing errors when parameter changes were processed when the newscale positions were "" (either before GetPositions() was called, or no newscale present). I handle this case. 

- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/191

- None

- [x] Tested in 447


XX-Yin (approved):
Thanks for careful checking and fixing this!

PR #246, opened 2024-02-12
Added session control for optogenetics

XX-Yin:
  
### Describe changes:
1. To periodically turn on/off optogenetics if `session control` is `on` .  `Fraction of session (trials)` controls the period. `Start with` controls the on or off state of the first period. When `Alternating across sessions` is `on` in the json file, the `Start with` will be changed to the opposite when loading this json file. 
image

2. Periodically  and automatically turn on/off optogenetics in a session. In this way, we can do consecutive perturbation and perform within session control.  

Turn off `Session control` if you don't use this function. It is also turned off by default. 

- [x] Tested on my own computer. 
image
- [x] test in ephys

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2013,7 +2013,13 @@ class Window(QMainWindow):
                             elif Tag==1:
                                 index = widget.findText(value)
                             if index != -1:
-                                widget.setCurrentIndex(index)
+                                # Alternating on/off for SessionStartWith if SessionAlternating is on
+                                if key=='SessionStartWith' and 'Opto_dialog' in Obj:
+                                    if Obj['Opto_dialog']['SessionAlternating']=='on':
+                                        index=1-index
+                                        widget.setCurrentIndex(index)
+                                else:
+                                    widget.setCurrentIndex(index)
                         elif isinstance(widget, QtWidgets.QDoubleSpinBox):
                             if Tag==0:
                                 widget.setValue(float(value[-1]))
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2015,9 +2015,10 @@ class Window(QMainWindow):
                             if index != -1:
                                 # Alternating on/off for SessionStartWith if SessionAlternating is on
                                 if key=='SessionStartWith' and 'Opto_dialog' in Obj:
-                                    if Obj['Opto_dialog']['SessionAlternating']=='on':
-                                        index=1-index
-                                        widget.setCurrentIndex(index)
+                                    if 'SessionAlternating' in Obj['Opto_dialog'] and 'SessionWideControl' in Obj['Opto_dialog']:
+                                        if Obj['Opto_dialog']['SessionAlternating']=='on' and Obj['OptogeneticsB']=='on' and Obj['Opto_dialog']['SessionWideControl']=='on':
+                                            index=1-index
+                                            widget.setCurrentIndex(index)
                                 else:
                                     widget.setCurrentIndex(index)
                         elif isinstance(widget, QtWidgets.QDoubleSpinBox):
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -53,6 +53,7 @@ class OptogeneticsDialog(QDialog):
         self._Laser_3()
         self._Laser_4()
         self._Laser_calibration()
+        self._SessionWideControl()
     def _connectSignalsSlots(self):
         self.Laser_1.currentIndexChanged.connect(self._Laser_1)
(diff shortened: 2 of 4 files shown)


alexpiet:
I'm not sure I understand. If `session control = on`, the session goes through blocks of opto and blocks of no opto?

src/foraging_gui/Foraging.py
```diff
@@ -2013,7 +2013,13 @@ def _Open(self):
                             elif Tag==1:
                                 index = widget.findText(value)
                             if index != -1:
-                                widget.setCurrentIndex(index)
+                                # Alternating on/off for SessionStartWith if SessionAlternating is on
```
alexpiet, line 2016: So this is alternating whether the session starts in an opto block, or not?

XX-Yin:
> I'm not sure I understand. If `session control = on`, the session goes through blocks of opto and blocks of no opto?

Yes, that's right!

hanhou:
I don't understand either. How does it interact with other parameters such as Fraction of session?

Also, can you find a better name? 'session control' is not very self-explanatory. It looks more like a block control? Although from the screenshot it does not align with any blocks...

src/foraging_gui/Foraging.py
XX-Yin: This is a very good point. Usually, we performed the optogenetics in consecutive sessions, so alternating the `SessionStartWith` should be fine. However, it's better to always check if this is an opto session. I have updated the code.

XX-Yin:
> blocks

Fraction of session determines the block size (e.g. the maximum trial is set to 1000, and the fraction of session is 0.1, then the block size is 100; the opto selection will be alternatively turned on/off every 100 trials). These blue dots are in blocks with session control on. 
image

I changed the name (see below). 
image

I also plan to develop a block control function for opto (e.g turning on opto in the first half of the block and turning off the opto in the second half of the block).

hanhou:
So if `session_control = False`, random photostim is given trial-by-trial based on the probability of `Fraction of session`, like we usually did before?

> maximum trial

Is maximum trial the one set in the main parameter window? It can cause some confusion if the photostim control panel does not fully determine the photostim schedule, but depends on another parameter elsewhere.

How about this:
1. Replace the binary `Session control` with a dropdown `Photostim schedule = {"random", "alternating chunks", "first half of each block", "last half of each block",...}`. This will allow you to expand the ways of doing photostim in the future. Also, note that I use "chunks" for photostim to avoid confusion with "block" of reward prob.
2. Only enable `Fraction of session` when `Photostim schedule == "random"`. If `Photostim schedule == "alternating chunks"`, use another variable `chunk size` to explicitly set the chunk size.

hanhou:
>  image

Another question, why the blue dots are not continuous? Were photostim conditioned on other things in this example?

XX-Yin:
> elsewhere

No, sorry for this confusion. If  `session_control = False`, we just select stimulation condition like normal (it's not dependent on `Fraction of session`). The `Fraction of session`  here is used to control the block size when `session_control` is `on`.

I updated the code to disable parameters shown below when  `session_control = False`. This can avoid some confusion.
image

hanhou:
> > elsewhere
> 
> No, sorry for this confusion. If `session_control = False`, we just select stimulation condition like normal (it's not dependent on `Fraction of session`). The `Fraction of session` here is used to control the block size when `session_control` is `on`.
> 
> I updated the code to disable parameters shown below when `session_control = False`. This can avoid some confusion. image

I see. So this part is new, just for "alternating chunks", and you use `probability` in the above panels to control the trial-by-trial probability? Will this `probability` be also effective when "alternating chunks" is ON, which explains why the blue dots are discontinuous?

XX-Yin:
> > image
> 
> Another question, why the blue dots are not continuous? Were photostim conditioned on other things in this example?

The logic is when the session control state is on, we still need to select stimulation condition. The probability is determined by probability for each condition, like shown below. Previously, it was set to 0.25. In this example shown below, it was set to 1 and it was continuous. 
image
image

XX-Yin:
> chunks

Yes, the probability for each condition is still active.

hanhou:
I'm approving it for speeding up the experiments. But I would suggest improving the photostimulation pannel by:
1. using better names 
2. using dropdown to allow future expansion
3. Isolating the parts for "within-trial control" (align to which event, etc.) and "photostimulation schedule" (given on which trials)

alexpiet:
> Btw, I'm trying to pin Sue, but she is not invited to this repo. @alexpiet are you the admin? Could you add her in?

I believe @jsiegle is the admin. 
@jsiegle Can you add Sue to the repo? Additionally, can you make Han, Xinxin, and me admin so we dont need to bug you?

hanhou:
@ZhixiaoSu Feel free to weigh in on how you would like the optogenetics control to be!

PR #253, opened 2024-02-14
only stopping photometry in the d box

alexpiet:
  
### Describe changes:
resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/255
This weeks updates stop photometry excitation at GUI open/close to control the state of photometry in the even of a gui crash. However my code did not restrict this to the photometry boxes, so when the C box gui opened or closed it stopped excitation in the D box. This is a quick fix

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2308,6 +2308,8 @@ class Window(QMainWindow):
         '''
             Stop either bleaching or photometry
         '''
+        if self.box_letter != "D":
+            return
         logging.info('Checking that photometry is not running')
         try:
             ser = serial.Serial(self.Teensy_COM, 9600, timeout=1)


XX-Yin:
It seems that this is not a long-term solution as we could run photometry in any rigs, not only in D.

alexpiet:
> It seems that this is not a long-term solution as we could run photometry in any rigs, not only in D.

Yes, thats true. Do you have a better solution to merge in now?

XX-Yin:
> > It seems that this is not a long-term solution as we could run photometry in any rigs, not only in D.
> 
> Yes, thats true. Do you have a better solution to merge in now?
> 
> Probably the better solution is to move the Teensy_COM setting. Right now the Teensy_COM is a setting in the computer specific ForagingSettings.json, maybe we should move it to the box specific Settings_box4.csv

We can also add a `PhotometryBox`  parameter in `ForagingSettings.json`.

XX-Yin:
Another solution is Don't run `_StopPhotometry` in non photometry session.

alexpiet:
@XX-Yin Can we merge this is as a quick fix, and deal with longer term solutions later?

XX-Yin (approved):

alexpiet:
Issue for long term solution: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/255

PR #279, opened 2024-03-05
removing SaveTraining from GUI

### LINKED ISSUE #229: training stage parameters are getting saved
long term - the stage parameters should not be changed when the individual sessions parameters are changed
short term - I can set up an auto-job to just copy the training stage parameters from the repo periodically, which will wipe any changes from the individual session parameters

alexpiet:
  
### Describe changes:
- SaveTraining() saves the definition of the training stages. We do not want this to happen, since the training stages are now centrally defined. I removed this function and the button in the GUI

- resolves #229

- The button that said "Save Training Stage" no longer exists. 

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -2374,8 +2374,7 @@ class AutoTrainDialog(QDialog):
                 return
                 
             self.widgets_locked_by_auto_train.extend(
-                [self.MainWindow.TrainingStage,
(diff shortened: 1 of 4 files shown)


alexpiet:
@hanhou Is there anything special I need to do with respect to AutoTrain?

hanhou:
Indeed, it causes an error in AutoTrain. Let me fix it quickly.
image

hanhou (approved):

Issue #186, opened 2024-01-23
New session does not clear the plots


alexpiet:
Starting a new session should clear all the visualizations

Issue #87, opened 2023-12-06
Visualization:Visualization.py:_PlotMatching:line 265:Cannot calculate a linear regression if all x values are identical - 165 


alexpiet:
Error occurred 165 times in the last week

Issue #86, opened 2023-12-06
Visualization:Visualization.py:_PlotMatching:line 265:Axis limits cannot be NaN or Inf - 81 


alexpiet:
Error occurred 81 times in the last week

Issue #85, opened 2023-12-06
Visualization:Visualization.py:_PlotMatching:line 265:Inputs must not be empty. - 1399 


alexpiet:
Error occurred 1399 times in the last week

PR #286, opened 2024-03-06
Visualization fixes

alexpiet:
  
### Describe changes:
- Only runs the regression on the matching plot if we have enough data, thus avoiding several errors. 
- Removes an unused `PlotV.finish` flag
- General code clean up
- Clear the plots when NewSession is pressed
- Adds some logic to check for training parameter versus behavior parameter when loading session

- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/85
- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/86
- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/87
- resolves https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/186

If you press `NewSession` the plots will be cleared. This will also happen briefly when loading a mouse

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2577,6 +2577,9 @@ class Window(QMainWindow):
         self.PhotometryRun=0
         self.unsaved_data=False
         self.ManualWaterVolume=[0,0]       
+    
+        # Clear Plots
+        self.PlotLick._Update(GeneratedTrials=None)
 
         # Add note to log
         logging.info('New Session complete')
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2759,7 +2759,7 @@ class Window(QMainWindow):
             self.GeneratedTrials=GeneratedTrials
             self.StartANewSession=0
             PlotM=PlotV(win=self,GeneratedTrials=GeneratedTrials,width=5, height=4)
-            PlotM.finish=1
+            #PlotM.finish=1
             self.PlotM=PlotM
             #generate the first trial outside the loop, only for new session
             self.ToReceiveLicks=1
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2579,7 +2579,7 @@ class Window(QMainWindow):
         self.ManualWaterVolume=[0,0]       
     
         # Clear Plots
-        self.PlotLick._Update(GeneratedTrials=None)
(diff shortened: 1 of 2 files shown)


src/foraging_gui/Visualization.py
```diff
@@ -61,34 +73,43 @@ def _Update(self,GeneratedTrials=None,Channel=None):
...
+        try:
             self._PlotLicks()
         except Exception as e:
             logging.error(str(e))
+
         try:
             self._PlotMatching()
         except Exception as e:
             logging.error(str(e))
-        self.finish=1
```
hanhou, line 72: @XX-Yin what was the original purpose of setting a `finish` flag? Is it related to threading? Is it safe to remove?

hanhou (approved):

src/foraging_gui/Visualization.py
```diff
@@ -61,34 +73,43 @@ def _Update(self,GeneratedTrials=None,Channel=None):
...
+
     def _PlotBlockStructure(self):
```
XX-Yin, line 94: Why do you replace all ax2 (ax3, ax1) with self.ax2 (ax3, ax1)? Is there a specific consideration?

src/foraging_gui/Foraging.py
```diff
@@ -2577,6 +2578,9 @@ def _NewSession(self):
         self.PhotometryRun=0
         self.unsaved_data=False
         self.ManualWaterVolume=[0,0]       
+    
+        # Clear Plots
+        self.PlotM._Update(GeneratedTrials=None,Channel=None)
```
XX-Yin, line 2583: This is a bug when we don't have PlotM. https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/206

src/foraging_gui/Visualization.py
alexpiet: because its simpler. There is no reason to do:
```
ax = self.ax1
ax.plot()
```
when we can use:
```
self.ax1.plot()
```

src/foraging_gui/Foraging.py
alexpiet: Ok, good catch. I will add a check. Under what conditions does PlotM not exist?

alexpiet: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/287

PR #288, opened 2024-03-07
Use schedule to make suggestion when loading mice

### LINKED ISSUE #281: Load mouse should auto-populate from the schedule
Quick load could auto-populate based on the schedule
- Figure out programmatic access to smartsheet schedule
- Turn into df, then find scheduled mouse.        
- User can change what schedule slot to look at to load the mouse scheduled at a different time (does not change schedule)
- user can override schedule       
- Quick load only does this on boxes on the schedule (otherwise defaults to loading based on user input)
- Could also add experimenter name #294 
- Related to #466 (need same access to schedule)

alexpiet:
  
### Describe changes:
- Uses the published behavior schedule to make a suggestion of which mouse to load. 
    - We determine the current time slot by looking at the most recent start time. If it's within 30 minutes of a start time, then we use that suggestion.  
- The user must accept or reject the suggestion
- If the user accepts the suggestion, then the user either verifies the mouse/last session (like before), or verifies that this is a new mouse (like before). 
- If the user rejects the suggestion, then we use the existing prompt for the user to select the mouse from a dropdown menu, or type in the ID
- If there is any problem loading the schedule, then we use the existing prompt for the user to select the mouse from a dropdown menu, or type in the ID
- Potential problems that are handled:
    - The schedule file is missing
    - An error loading the schedule
    - A formatting error in the schedule
    - The schedule is older than 1 week
    - Its more than 30 minutes before the first scheduled time of the day
    - Its after 4pm (so past the last schedule of the day)
    - The current box is not listed in the schedule at this time
    - More than one mouse listed in this time/box
- The schedule needs to have the following format:
    - the first column/second row contains the date in Month/Date/Year format
    - The "Box" column should have a format like "1A"
    - The "Time Slot" column should have format H:MIN-H:MIN
    - The "Mouse ID" column should have the mouse id
    - One "sheet", not multiple pages

- Typing in the mouse ID is cumbersome, slow, and error prone
- resolves #281 

- If we are able to make a suggestion based on the schedule, you will see this prompt: 

MicrosoftTeams-image3c5ed85059848185023763c2e3b1cb2206c9625d1136f67de0d929e88588b4eb<br>

- [ ] tested in 447
- [ ] Feedback from RAs
- [ ] Handle late afternoon sessions (4-6, 6:15, make sure AM/PM computed correctly, change cutoff time)
- [ ] We need a way to place the schedule on each rig computer
    - [ ] by monday morning 6m
    - [ ] as a `.csv` file
    - [ ] https://github.com/smartsheet/smartsheet-python-sdk

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -29,7 +29,7 @@ codebase_curriculum_schema_version = DynamicForagingCurriculum.model_fields['cur
 logger = logging.getLogger(__name__)
 
 class MouseScheduleDialog(QDialog):
-    def __init__(self, MainWindow, schedule_mouse, parent=None):
+    def __init__(self, MainWindow, scheduled_mouse, parent=None):
         super().__init__(parent)
         self.MainWindow = MainWindow
         self.setWindowTitle('Box {}, Load Mouse'.format(self.MainWindow.box_letter))
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -46,17 +46,17 @@ class MouseScheduleDialog(QDialog):
         font.setPointSize(12)
         msg0.setFont(font)
 
-        msg = QLabel('mouse: <span style="color:purple;font-weight:bold">{}</span>'.format(scheduled_mouse['Mouse ID']))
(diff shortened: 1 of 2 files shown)


alexpiet:
closing this PR because I cannot get programmatic access to the schedule.

hanhou:
Is there an API for smartsheet?

alexpiet:
it wont let me generate an access token

hanhou:
Not sure if @macarenasa could help

PR #292, opened 2024-03-11
Remove outdated menu options

### LINKED ISSUE #290: Open Recent button does nothing
File > Open Recent does nothing
Its defined in the `.ui` files, but not connected to any slots in `Foraging.py` I suggest we remove it, since the recent updates to load seem to cover this use case. @XX-Yin any thoughts?

### LINKED ISSUE #291: multiple "save" options in the File menu
Under the "File" button we have:
- Save (calls `_Save()`)
- Force Save (calls `_ForceSave()`)
- Save (without restarting a session) (calls `_SaveContinue()`)
- Save As (calls `_SaveAs()`)

I propose:
- removing `Save (without restarting a session)`, because `_Save()` now defaults to `SaveContinue=1`, so its completely redundant. 
- I think we should keep `Force Save`. 
- I'm 50/50 on keeping `SaveAs()`, since I would like to encourage always saving in the standard spot. @XX-Yin Is this valuable in the ephys room?

alexpiet:
  
### Describe changes:
- Removes the `Open Recent` option in the file menu, that was never implemented
- Removes the `Save (without restarting)` option in the file menu, since this is redundant with `Save`
- Removes code mention of `TP_log_folder` in MyFunctions, because this attribute was never used, and generated errors when starting a new session

- Resolves #291 
- Resolves #290 

Removes outdated options in the file menu

- [x] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -176,7 +176,6 @@ class Window(QMainWindow):
         self.action_Open.triggered.connect(self._Open)
         self.action_Save.triggered.connect(self._Save)
         self.actionForce_save.triggered.connect(self._ForceSave)
-        self.SaveContinue.triggered.connect(self._SaveContinue)
         self.SaveAs.triggered.connect(self._SaveAs)
         self.action_Exit.triggered.connect(self._Exit)
         self.action_New.triggered.connect(self._New)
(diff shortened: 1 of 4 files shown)


alexpiet:

XX-Yin (approved):

Issue #190, opened 2024-01-24
Check if high speed camera is enabled


alexpiet:
- Currently in the training room, there is no high speed camera. In the ephys rig there is.
- We need to modify the bonsai workflow to gracefully check if a high speed camera is enabled or not. If a camera is present, enable the node. If a camera is not present, disable the node. 
- this is required before automatic updates in the ephys rigs

alexpiet:
@bruno-f-cruz Can you point us in the right direction?

bruno-f-cruz:
I would advise solving this at the level of the schema tbh. If there is a camera entry try to connect if not don't subscribe to the camera. I have shown xin xin how to do this with the nidaq already. Use the same pattern. Otherwise you could simply try to connect to the camera and if it doesn't exist catch the error and ignore it. Which one would you prefer?

XX-Yin:
Hi @bruno-f-cruz, do you have time to show me the `catch the error and ignore`? The excel schema method does not work for the nidaq.

bruno-f-cruz:
Pick a time from my next week s calendar. Meanwhile have you tried the docs?

https://bonsai-rx.org/docs/api/Bonsai.Reactive.Catch.html
https://bonsai-rx.org/docs/api/Bonsai.Reactive.OnErrorResumeNext.html

alexpiet:
![IMG_20240321_092033181](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/assets/7605170/a083bed5-ecc6-427b-af1f-c132923dc91a)

@bruno-f-cruz Showed @XX-Yin and I how to do this. The idea is we put a value in the rig `.csv` file, and then set the workflow to only subscribe to the high speed camera node if the appropriate value is in the rig file.

I will update the setting files to contain an entry `HighSpeedCamera`. It will default to `0`, and if its `0`, the camera node will not be subscribed to. If you want to use a high speed camera, then you set the node to `1`

bruno-f-cruz:
For completeness sake:

![image](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/assets/7049351/6cfa5a2d-9b97-4913-8544-4eb55a80c2a2)


<details>
    <summary>Workflow (Copy+Paste me)</summary>
    
```xml

<?xml version="1.0" encoding="utf-8"?>
<WorkflowBuilder Version="2.8.1"
[...]

PR #318, opened 2024-03-21
Update camera node

XX-Yin:
  
### Describe changes:
1. Subscribe the` HighSpeedCamera` node based on the `HighSpeedCamera` (0 does not subscribe; otherwise subscribe) field in the `Settings_box#.csv`. 
image
2. Read the camera Serial Number (BottomCamera and SideCamera) from the `Settings_box#.csv`. 
image

4. Updated the template `Settings_box#.csv`

Resolves: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/190#issuecomment-2013007954

1. We need to add `HighSpeedCamera` field to the Settings_box#.csv, and set the value to `0` in the training rig. 
2. In the ephys rig we should add the `HighSpeedCamera` field to `Settings_box#.csv` and set the value to `1` if we want to use the camera node. The serial numbers for `BottomCamera` and `SideCamera` should also be set in Settings_box#.csv.
Tested in the ephys rig1.

### DIFF
--- a/src/setting_templates/Settings_box1.csv
+++ b/src/setting_templates/Settings_box1.csv
@@ -4,3 +4,6 @@ BonsaiOsc1,4002
 BonsaiOsc2,4003
 BonsaiOsc3,4004
 BonsaiOsc4,4005
+HighSpeedCamera,0
+SideCamera,23022724
+BottomCamera,23022709
--- a/src/setting_templates/Settings_box1.csv
+++ b/src/setting_templates/Settings_box1.csv
@@ -5,5 +5,5 @@ BonsaiOsc2,4003
 BonsaiOsc3,4004
 BonsaiOsc4,4005
 HighSpeedCamera,0
-SideCamera,23022724
-BottomCamera,23022709
+SideCamera,12345678
+BottomCamera,12345678
--- a/src/workflows/foraging.bonsai
+++ b/src/workflows/foraging.bonsai
@@ -6865,308 +6865,361 @@
                 </Workflow>
               </Builder>
             </Expression>
-            <Expression xsi:type="Disable">
-              <Builder xsi:type="GroupWorkflow">
-                <Name>HighSpeedCamera</Name>
-                <Workflow>
-                  <Nodes>
-                    <Expression xsi:type="SubscribeSubject">
-                      <Name>BottomCameraFrames</Name>
-                    </Expression>
-                    <Expression xsi:type="MemberSelector">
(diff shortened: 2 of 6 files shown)


src/setting_templates/Settings_box1.csv
```diff
@@ -4,3 +4,6 @@ BonsaiOsc1,4002
 BonsaiOsc2,4003
 BonsaiOsc3,4004
 BonsaiOsc4,4005
+HighSpeedCamera,0
+SideCamera,23022724
```
alexpiet, line 8: The SideCamera and BottomCamera values should default to 0 (1) for clarity, and (2) the values will differ in different ephys rigs, right? So it doesnt make to default to a specific rig

src/workflows/foraging.bonsai
```diff
@@ -6865,308 +6865,361 @@
...
+                    <Name>CameraFrequency</Name>
+                  </Expression>
+                  <Expression xsi:type="PropertyMapping">
+                    <PropertyMappings>
+                      <Property Name="FrameRate" />
+                    </PropertyMappings>
+                  </Expression>
+                  <Expression xsi:type="Combinator">
+                    <Combinator xsi:type="ffmpeg:VideoWriter">
+                      <ffmpeg:FileName>C:\Users\svc_aind_ephys\Documents\temporaryvideo\2024-03-21_09-52-50\HarpFolder/../VideoFolder/side_camera.avi</ffmpeg:FileName>
```
alexpiet, line 7130: Why is Filename hardcoded here?

alexpiet (changes requested):

src/setting_templates/Settings_box1.csv
XX-Yin: I changed it to 12345678 to differentiate between the Boolean and the serial number.

src/workflows/foraging.bonsai
XX-Yin: Every time we run the bonsai workflow it will change the video file (and also the COM and so on). These paths will be overwritten in each new session.

alexpiet:
@XX-Yin It does not work on my laptop. I get a system.IO.IOException: The port "COM14" does not exist. I do not specify that com port in any of my setting files. I'm wondering if something ephys specific from the bonsai-workflow got saved in the workflow?

alexpiet:
Ok, now its working!

alexpiet (approved):

PR #342, opened 2024-03-27
Fix water volume

### LINKED ISSUE #341: Small differences between water valve opening time and the water volume.
The water volume is changed with the valve open time based on the water calibration. We only keep 3 decimal places for valve open time, so the water volume may show small differences when we set it based on the valve open time. 

For example, in the case shown below, where R(ul) is set to 3, R(s) will be set to 0.034. If we set R(s) to 0.034, R(ul) is not exactly 3, but 3.01.

image

XX-Yin:
  
### Describe changes:
Skip `LeftValue`, `RightValue`, `GiveWaterL`, `GiveWaterR` if `WaterCalibrationResults` is not empty as they will be set by the corresponding volume when loading the json file. 

The water volume is changed with the valve open time based on the water calibration. We only keep 3 decimal places for valve open time, so the water volume may show small differences when we set it based on the valve open time. Such small differences do not matter for tracking water consumption, but it's good to make it consistent to avoid confusion. Changing of water valve open time will change the water volume. Here, I skipped loading water valve open time. In this way, the water volume will keep consistent across sessions. 

Resolves #341
No

Small changes. Only tested on my own computer.

### DIFF
--- a/.github/workflows/production_merge_action.yml
+++ b/.github/workflows/production_merge_action.yml
@@ -1,0 +1,24 @@
+name: production-merge-action
+run-name: create production merge 
+on:
+    schedule:
+    - cron: '0 8 * * 5'
+jobs:
+    create-production-merge:
+        runs-on: ubuntu-latest
+        steps:
+            - name: checkout repository
+              uses: actions/checkout@v4
+              with:
+                fetch-depth: 0
+            - name: get current date
+              id: date
+              run: |
+                echo "date=$(date +'%Y-%m-%d')" >> "$GITHUB_ENV"
+            - run: echo $date
+            - run: echo "testing string ${{ env.date }}"
+            - name: create pull request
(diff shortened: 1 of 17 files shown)


alexpiet:
> The water volume is changed with the valve open time based on the water calibration. We only keep 3 decimal places for valve open time, so the water volume may show small differences when we set it based on the valve open time. Such small differences do not matter for tracking water consumption, but it's good to make it consistent to avoid confusion.

Thats not really the issue being fixed by this PR. The issue being fixed here is that the `_Open()` function loads the water calibration data from the old session, which may differ from the current session. Can you fix the description?

Does this mean that old data might have been collected using out-dated water calibration information?

Is there an equivalent issue for the laser calibration?

Can you please test this in 447?

XX-Yin:
I tested in 447. It works as expected.

alexpiet (approved):

PR #369, opened 2024-04-07
Update nwb

XX-Yin:
  
### Describe changes:
Updated the NWB file. 
- Fixed https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/325#event-12329588489. 
- Added opto fields (`session_wide_control`, `fraction_of_session`,`session_start_with`,`session_alternation`) to `add_trial` in the NWB. 
- Added metadata `laser_1_calibration_power`, `laser_2_calibration_power`, `laser_1_target_areas` and `laser_2_target_areas`. 
- Added metadata `commit_ID`, `repo_url` and `current_branch` to track the software version. 
- Cleaned the code.
- Added a test function `test_bonsai_json_to_nwb` to enforce a comprehensive test. If we change json file format in the future, be sure to add a new example json to #377 and `test_json_urls`. 

    If all successful:
    
    If some tests failed:
    

Resolved https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/325#event-12329588489. 
No
Tested on my own computer


ZhixiaoSu (approved):

src/foraging_gui/TransferToNWB.py
```diff
@@ -38,8 +38,17 @@ def _get_field(obj, field_list, reject_list=[], index=None, default=np.nan):
         field_list = [field_list]
         
     for f in field_list:
-        if hasattr(obj, f):
-            value = getattr(obj, f)
+            tag=0
```
hanhou, line 41: Can you use more readable variable names?

hanhou (changes requested):

src/foraging_gui/TransferToNWB.py
```diff
@@ -38,8 +38,17 @@ def _get_field(obj, field_list, reject_list=[], index=None, default=np.nan):
...
+            if type(obj) is type:
+                if hasattr(obj, f):
+                    value = getattr(obj, f)
+                    tag=1
+            elif type(obj) is dict:
+                if f in obj:
+                    value = obj[f]
+                    tag=1
```
hanhou, line 49: can you add doc string for this part?

src/foraging_gui/TransferToNWB.py
```diff
@@ -38,8 +38,17 @@ def _get_field(obj, field_list, reject_list=[], index=None, default=np.nan):
         field_list = [field_list]
         
     for f in field_list:
-        if hasattr(obj, f):
-            value = getattr(obj, f)
+            tag=0
+            if type(obj) is type:
```
hanhou, line 42: what does "is type" mean?

hanhou (changes requested):

src/foraging_gui/TransferToNWB.py
```diff
@@ -171,6 +180,12 @@ class obj:
         # Performance
         'foraging_efficiency': _get_field(obj, 'B_for_eff_optimal'),
         'foraging_efficiency_with_actual_random_seed': _get_field(obj, 'B_for_eff_optimal_random_seed'),
+
+        # Optogenetics
+        'laser_1_calibration_power': float(_get_field(obj.Opto_dialog, 'laser_1_calibration_power')),
+        'laser_2_calibration_power': float(_get_field(obj.Opto_dialog, 'laser_2_calibration_power')),
+        'laser_1_target_areas': _get_field(obj.Opto_dialog, 'laser_1_target',default='None') or 'None',
+        'laser_2_target_areas': _get_field(obj.Opto_dialog, 'laser_2_target',default='None') or 'None',
```
hanhou, line 188: Yon don't need `or 'None'` if your default value is 'None' because it will not take effect.

hanhou (changes requested):
Please also test with an old json file without photostimulation and let me know if it passes.

src/foraging_gui/TransferToNWB.py
XX-Yin: I changed the name to `has_field`. 

Hi Han, thanks for reviewing this over the weekend. I don't expect you to respond so quickly.

src/foraging_gui/TransferToNWB.py
XX-Yin: Updated

# the obj.Opto_dialog is a dictionary

XX-Yin:
> Please also test with an old json file without photostimulation and let me know if it passes.

I tested both the old and new json files.

src/foraging_gui/TransferToNWB.py
XX-Yin: Updated

src/foraging_gui/TransferToNWB.py
```diff
@@ -272,41 +293,52 @@ class obj:
...
+            LaserStartC = '0'
             LaserStartOffsetC = 0
-            LaserEndC = 0
+            LaserEndC = '0'
             LaserEndOffsetC = 0
-            LaserProtocolC = 0
+            LaserProtocolC = '0'
             LaserFrequencyC = 0
             LaserRampingDownC = 0
             LaserPulseDurC = 0
```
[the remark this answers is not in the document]
XX-Yin: I updated the empty string to 'None'. There is a bug if we directly use None or np.nan.

hanhou:
It fails with this file during the final writing nwb stage. Please check
[704151_2024-02-27_09-59-17.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14925453/704151_2024-02-27_09-59-17.json)

`TypeError: Can't implicitly convert non-string objects to strings`

src/foraging_gui/TransferToNWB.py
```diff
@@ -163,14 +164,25 @@ class obj:
...
         
         # Performance
         'foraging_efficiency': _get_field(obj, 'B_for_eff_optimal'),
         'foraging_efficiency_with_actual_random_seed': _get_field(obj, 'B_for_eff_optimal_random_seed'),
+
+        # Optogenetics
+        'laser_1_calibration_power': float(_get_field(obj.Opto_dialog, 'laser_1_calibration_power')),
+        'laser_2_calibration_power': float(_get_field(obj.Opto_dialog, 'laser_2_calibration_power')),
+        'laser_1_target_areas': _get_field(obj.Opto_dialog, 'laser_1_target',default='None'),
+        'laser_2_target_areas': _get_field(obj.Opto_dialog, 'laser_2_target',default='None'),
```
hanhou, line 180: A tip to reduce backward compatibility issues is to be mindful when you call something like `obj.xxx` you always catch the case where we didn't have that field in the past.

hanhou (changes requested):

src/foraging_gui/TransferToNWB.py
XX-Yin: Fixed.

XX-Yin:
> I tested the code with two json files from training rig, one newer and one older. Both of them failed. Please fix.
> 
> [704151_2024-02-27_09-59-17.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14925453/704151_2024-02-27_09-59-17.json) `TypeError: Can't implicitly convert non-string objects to strings`
> 
> [668551_2023-06-16.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14925528/668551_2023-06-16.json) `type object 'obj' has no attribute 'Opto_dialog'`
> 
> Whenever we change `TransferToNWB.py`, we should make sure it is backwards compatible with all the old json files, because we will always redo the transfer for all the old files. And since we also have different versions of json files before, we should at least test one of each historical format.
> 
> We may need to set up an automatic testing using [unitest ](https://www.dataquest.io/blog/unit-tests-python/)for this.

They were fixed. 

1. The first error is a version issue. The very old version doesn't have `Opto_dialog`. 
2. The second error is due to there are different data types, `null` and `False` (boolean), in the `auto_train_stage_overridden`. We transferred `null` to `None (string)` in the nwb, so there were `string` and `boolean` in the field `TP_auto_train_stage_overridden`. And the nwb can't handle this situation. I transferred the `null` to `False` to solve this issue.

src/foraging_gui/TransferToNWB.py
```diff
@@ -18,7 +18,7 @@
 
 logger = logging.getLogger(__name__)
 
-def _get_field(obj, field_list, reject_list=[], index=None, default=np.nan):
+def _get_field(obj, field_list, reject_list=[None, np.nan,''], index=None, default=np.nan):
```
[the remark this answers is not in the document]
XX-Yin: I found that `None, np.nan and ''` could easily cause errors if mixed with other data types in `add_Trial_column`, so I rejected all of them (a lot of errors come from this type). Users can still set it via `default`.

src/foraging_gui/TransferToNWB.py
```diff
@@ -129,9 +131,9 @@ class obj:
     # need to get more subject information through subject_id
     nwbfile.subject = Subject(
         subject_id=obj.ID,
-        description='Animal name:'+obj.ID+'  Weight after(g):'+obj.WeightAfter,
+        description='Animal name:'+obj.ID,
         species="Mus musculus",
-        weight=obj.WeightBefore,
+        weight=_get_field(obj, 'WeightAfter'),
```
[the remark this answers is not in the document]
XX-Yin: "WeightBefore" existed in the old data type, and is gone now.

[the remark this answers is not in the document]
hanhou: So we decided to use weightafter as subject.weight for all data, leaving the field of old data np.nan, right? Because we should be consistent about the naming.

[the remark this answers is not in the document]
XX-Yin: I remember that for the old data we also have the `WeightAfter` (I checked session 668551_2023-06-16.json).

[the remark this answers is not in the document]
XX-Yin: I set the default as np.nan

src/foraging_gui/TransferToNWB.py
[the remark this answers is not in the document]
hanhou: That is exactly the problem. Say, you have a variable of string type obj.str_field = ['string1', 'string2', '', 'string3'], if you call `_get_field(obj, 'str_field')` by default, you'll get `['string1', 'string2', np.nan, 'string3']`, which has incompatible data types (np.nan is a float). Yes, you can still use `_get_field(obj, 'str_field', default='')`, but it's weird that you reject `''` and then set it as a default.

Anyway, your code works too. It's just a matter of style where I prefer to explicitly specify reject_list on every function call.

hanhou:
> 2. The second error is due to there are different data types, `null` and `False` (boolean), in the `auto_train_stage_overridden`. We transferred `null` to `None (string)` in the nwb, so there were `string` and `boolean` in the field `TP_auto_train_stage_overridden`. And the nwb can't handle this situation. I transferred the `null` to `False` to solve this issue.

Okay, your current code should work, but it actually changed my original intention.

If AutoTrain is not engaged, you'll set it to `False`
```python
auto_train_stage_overridden=_get_field(obj, 'TP_auto_train_stage_overridden', index=i, default=False)
``` 
in my original code, I set it to `np.nan` (which is compatible with True and False) to indicate `stage_overridden` doesn't have a definition if autotrain is not engaged at all. (essentially `False` and `not relevant` are the same)
```python
auto_train_stage_overridden=_get_field(obj, 'TP_auto_train_stage_overridden', index=i, default=None) or np.nan
```

XX-Yin:
> > 2. The second error is due to there are different data types, `null` and `False` (boolean), in the `auto_train_stage_overridden`. We transferred `null` to `None (string)` in the nwb, so there were `string` and `boolean` in the field `TP_auto_train_stage_overridden`. And the nwb can't handle this situation. I transferred the `null` to `False` to solve this issue.
> 
> Okay, your current code should work, but it actually changed my original intention.
> 
> If AutoTrain is not engaged, you'll set it to `False`
> 
> ```python
> auto_train_stage_overridden=_get_field(obj, 'TP_auto_train_stage_overridden', index=i, default=False)
> ```
> 
> in my original code, I set it to `np.nan` (which is compatible with True and False) to indicate `stage_overridden` doesn't have a definition if autotrain is not engaged at all. (essentially `False` and `not relevant` are not the same)
> 
> ```python
> auto_train_stage_overridden=_get_field(obj, 'TP_auto_train_stage_overridden', index=i, default=None) or np.nan
> ```

Yes, the np.nan can work, but not the None. I updated the default to np.nan

hanhou:
I added a test function `test_bonsai_json_to_nwb` to enforce a comprehensive test. If we change json file format in the future, be sure to add the new format to `test_json_urls`. 

image

- see #377

hanhou:
@XX-Yin  Let me know if the test function works good on your side. If yes, we can approve this PR.

hanhou:
@XX-Yin @ZhixiaoSu If you would like to extract some photostim meta data from nwb to fill in in the master session table of the Streamlit app, you can issue an PR to add some stuff [here](https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-basic/blob/d173153e316d31164f8453d2324b89e4c357b687/code/process_nwbs.py#L182). Then you'll be able to do filtering and further visualiation in the app.

hanhou (approved):

XX-Yin:
> @XX-Yin Let me know if the test function works good on your side. If yes, we can merge this PR.

It works good on my side.

src/foraging_gui/TransferToNWB.py
```diff
@@ -271,42 +289,45 @@ class obj:
...
-                LaserRampingDownC = getattr(obj, f'TP_RD_{Sc}')[i]
-                LaserPulseDurC = getattr(obj, f'TP_PulseDur_{Sc}')[i]
+            if getattr(obj, f'TP_Laser_{Sc}')[i] == 'Blue':
+                LaserWavelengthC = float(473)
+            elif getattr(obj, f'TP_Laser_{Sc}')[i] == 'Red':
+                LaserWavelengthC = float(647)
+            elif getattr(obj, f'TP_Laser_{Sc}')[i] == 'Green':
+                LaserWavelengthC = float(547)
+            LaserLocationC = str(getattr(obj, f'TP_Location_{Sc}')[i])
+            Laser1Power=float(eval(_get_field(obj, field_list=[f'TP_Laser1_power_{Sc}',f'TP_LaserPowerLeft_{Sc}'],index=i,default='[np.nan,np.nan]'))[1])
```
hanhou, line 317: I have a new question about this. Is this backward compatible? The previous code is
```python
LaserPowerC = getattr(obj, f'TP_LaserPower_{Sc}')[i]
```
but now your `field_list` becomes `field_list=[f'TP_Laser1_power_{Sc}', f'TP_LaserPowerLeft_{Sc}']`. Shouldn't the second option be `f'TP_LaserPower_{Sc}'`?

src/foraging_gui/TransferToNWB.py
```diff
@@ -317,91 +338,88 @@ class obj:
...
+                        
+                        session_wide_control=_get_field(obj, 'TP_SessionWideControl', index=i, default='None'),
+                        fraction_of_session=float(_get_field(obj, 'TP_FractionOfSession', index=i, default=np.nan)),
+                        session_start_with=_get_field(obj, 'TP_SessionStartWith', index=i, default='None'),
+                        session_alternation=_get_field(obj, 'TP_SessionAlternating', index=i, default='None'),
+                        # add all auto training parameters (eventually should be in session.json)
+                        auto_train_engaged=_get_field(obj, 'TP_auto_train_engaged', index=i,default='None'),
+                        auto_train_curriculum_name=_get_field(obj, 'TP_auto_train_curriculum_name', index=i, default='None'),
+                        auto_train_curriculum_version=_get_field(obj, 'TP_auto_train_curriculum_version', index=i, default='None'),
+                        auto_train_curriculum_schema_version=_get_field(obj, 'TP_auto_train_curriculum_schema_version', index=i, default='None'),
```
hanhou, line 400: @XX-Yin  Well, you changed the default from 'none' to 'None', which unfortunately broke the AutoTrain pipeline...

Next time, please be careful not to change the result of a code when you refactor it.

hanhou: To make sure the whole system is up and running before the next morning, I just spent another hour fixing everything, which is really frustrating...

I should have reviewed the code more carefully. At the same time, we should be 200% careful when our code has a huge impact.

src/foraging_gui/TransferToNWB.py
XX-Yin: Hi Han, the second field should be `f'TP_LaserPowerLeft_{Sc}'` instead of  `f'TP_LaserPower_{Sc}'`. In the initial code, I did not differentiate Left and Right laser, so it was `f'TP_LaserPower_{Sc}'`. We don't have opto sessions with `f'TP_LaserPower_{Sc}'`.

hanhou: I don't understand. You mean previously it has always been f'TP_LaserPowerLeft_{Sc}' in the json file, and the one you used f'TP_LaserPower_{Sc}' was actually wrong?

XX-Yin: In the very beginning version of the GUI, the left and right lasers were combined in f'TP_LaserPower_{Sc}' just like your previous GUI. I then separated them to  f'TP_LaserPowerLeft_{Sc}' and f'TP_LaserPowerRight_{Sc}'.

hanhou: My question is: has the current code correctly captured all historical field names of laser power?

From what you're describing:

Phase 1. You start with f'TP_LaserPower_{Sc}' in both json and convertNWB code

Phase 2. Then at some point you split it to left and right in bonsai json, but you didn't change in the convertNWB code (that is why I asked whether the convert code was "wrong" during that period of time)

Phase 3. In the recent update of the GUI, you changed the laser power name from left/right to 1/2. So in this PR, you add the new name with 1/2 in field_list, together with a left/right for backward compatibility, which finally fixes the mismatch in phase 2.

Are I understanding correctly?

If so, have you ever collected photostim data during phase 1? If you did, then the current PR is not compatible with those data.

Issue #370, opened 2024-04-07
Settings/version are not saved in the session Json


XX-Yin:
- The settings from the `ForagingSettings.json`  and  `Settings_box#.csv` are not explicitly saved in the session json. 
- The commit ID to track versions is not saved.

PR #371, opened 2024-04-07
Save settings and commit ID

XX-Yin:
  
### Describe changes:
- Save the `ForagingSettings.json` to the session Json
- Save the `Settings_box#.csv` to the session Json
- Save the commit ID, repo url and the branch name to the session Json
- Save folder information. 
image

Resolved https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/370

No

Tested in 323_EPHYS3 and 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -843,7 +843,7 @@ class Window(QMainWindow):
         self.window_title = window_title
 
         # Get the commit hash of the current version of this Python file
-        self.get_file_git_infor(__file__)
+        self.commit_ID, self.current_branch, self.result_url=log_git_hash()
 
     def _InitializeBonsai(self):
         '''
@@ -3433,9 +3433,12 @@ def log_git_hash():
     try:
         git_hash = subprocess.check_output(['git','rev-parse','--short', 'HEAD']).decode('ascii').strip()
         git_branch = subprocess.check_output(['git','branch','--show-current']).decode('ascii').strip()
+        result_url = subprocess.check_outputn(['git', 'remote', 'get-url', 'origin']).decode('ascii').strip()
         logging.info('Current git commit branch, hash: {}, {}'.format(git_branch,git_hash))
+        return git_hash, git_branch, result_url
     except Exception as e:
         logging.error('Could not log git branch and hash: {}'.format(str(e)))
+        return None, None, None
 
 def show_exception_box(log_msg):
     '''
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -737,12 +737,12 @@ class Window(QMainWindow):
         }
         
         # Try to load the settings file        
-        Settings = {}
+        self.Settings = {}
         try:
             if os.path.exists(self.SettingFile):
                 # Open the JSON settings file
                 with open(self.SettingFile, 'r') as f:
-                    Settings = json.load(f)
+                    self.Settings = json.load(f)
                 logging.info('Loaded settings file')
             else:
                 logging.error('Could not find settings file at: {}'.format(self.SettingFile))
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -736,13 +765,32 @@ def _GetSettings(self):
             'default_ui':'ForagingGUI.ui'
         }
         
+        # Try to load Settings_box#.csv
+        self.SettingsBox={}
+        try:
+            if os.path.exists(self.SettingsBoxFile):
```
alexpiet, line 771: Why load the csv as a dataframe only to convert to a dictionary? Lets just load directly as a dictionary, and avoid importing the pandas package

src/foraging_gui/Foraging.py
```diff
@@ -141,6 +143,33 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
             self._ReconnectBonsai()   
         logging.info('Start up complete')
     
+    def get_file_git_infor(self,file_path):
```
alexpiet, line 146: This information already gets generated, but only added to the log

src/foraging_gui/Foraging.py
```diff
@@ -796,7 +844,8 @@ def _GetSettings(self):
         window_title = '{}'.format(self.current_box)
         self.window_title = window_title
 
-
+        # Get the commit hash of the current version of this Python file
+        self.commit_ID,self.repo_url, self.current_branch = self.get_file_git_infor(__file__)
```
alexpiet, line 848: Why return the values from the function only to save them as object fields? Why not have the function save them as object fields directly?

alexpiet (changes requested):

src/foraging_gui/Foraging.py
alexpiet: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L3361

src/foraging_gui/Foraging.py
XX-Yin: I updated it

src/foraging_gui/Foraging.py
XX-Yin: We have already installed the pandas package. How to directly load the csv as a dictionary?

alexpiet: https://stackoverflow.com/questions/6740918/creating-a-dictionary-from-a-csv-file

XX-Yin: Ok, thanks for sharing. I think it's ok to use pandas here as we already have that package installed.

src/foraging_gui/Foraging.py
XX-Yin: I combined these two functions.

XX-Yin:
@alexpiet I installed the harp-python package in all of behavior computers.

src/foraging_gui/Foraging.py
```diff
@@ -3368,11 +3406,14 @@ def log_git_hash():
         Add a note to the GUI log about the current branch and hash. Assumes the local repo is clean
     '''
     try:
-        git_hash = subprocess.check_output(['git','rev-parse','--short', 'HEAD']).decode('ascii').strip()
```
alexpiet, line 3371: Using `git rev-parse --short HEAD` takes 50ms on my computer compared to 870ms for `git log -n 1`

XX-Yin: "git rev-parse --short HEAD" only gets the head 7 digits.

XX-Yin: I updated the command to `git rev-parse HEAD`.

src/foraging_gui/Foraging.py
alexpiet: Why not update the code to use csv straight to a dictionary for readability

src/foraging_gui/Foraging.py
```diff
@@ -869,7 +890,8 @@ def _GetSettings(self):
         window_title = '{}'.format(self.current_box)
         self.window_title = window_title
 
-
+        # Get the commit hash of the current version of this Python file
+        self.commit_ID, self.current_branch, self.repo_url=log_git_hash()
```
alexpiet, line 894: This results in `log_git_hash` getting called twice, once at start up, and once here. We should avoid calling the exact function twice

XX-Yin: It's more elegant to call it only once. I updated the code

src/foraging_gui/Foraging.py
XX-Yin: I think the pandas is doing the same thing.

XX-Yin: Hey Alex, you prefer the csv.reader() instead of the pandas.read_csv()?

alexpiet: Or at least doing something like this: 
```
SettingsBox = {row[0]:row[1] for rows in df}
```
instead of 

```                
for idx, row in df.iterrows():
    key = row.iloc[0]  # First column
    value = row.iloc[1]  # Second column
    self.SettingsBox[key] = value
```

XX-Yin: I replaced it with `self.SettingsBox = {row[0]: row[1] for _, row in df.iterrows()}`

alexpiet (approved):

PR #373, opened 2024-04-08
Update delay

XX-Yin:
  
### Describe changes:
The previous task logic was that if a lick was detected, the delay would be restarted.

The new logic is the ITI and delay will restart if a lick is detected during the delay period. The duration of the new ITI and delay is the same as in the current trial. 
Old:
image
New:
image
image
 

The animal shows very high early licking rate. The new delay logic (similar to what Sue used before. The difference is that in Sue's case the new ITI and delay were redrew from exponential distribution) may help reduce the early licking rate. 

Related to discussion 
https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/309
https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/323

1. The task structure is changed. Licking during the delay will restart the entire ITI and delay.
2.  The format of B_DelayStartTime and B_DelayStartTimeHarp is changed from a matrix to a list, and  each trial may have multiple values due to repeated delays. `TransferToNWB.py needs to be updated to make it compatible with this change.`  

- [x] Tested on my own computer
- [ ] Tested in 447


XX-Yin:
@ZhixiaoSu @hanhou I am updating the delay period. The proposed logic is when we detect early licks during the delay period, the entire ITI and delay will be restarted with the same parameters of the current trial. Here, I don't plan to redraw the exponential distribution of the ITI and delay. Otherwise, it will make the code more complex. Do you have any concerns?

XX-Yin:
@ZhixiaoSu @alexpiet This is ready for review. Hi Sue, can we find a time to test it together?

ZhixiaoSu:
Yes, let's test it together! I think it may make a big change in how some animals do the task. When testing the lick sensors, I was touching the lick tubes constantly to get more 'licks' and I feel every time I stop for around 1s, I hear a go cue. This is probably because of the 1s fixed punishment. If a mouse also feels this way, it can explain why they have more iti lickings: they thought go cue comes when they keep licking and then stop.

XX-Yin:
> "When testing the lick sensors, I was touching the lick tubes constantly to get more 'licks' and I feel every time I stop for around 1s, I hear a go cue. " 

You mean it's like a hallucination?

> Yes, let's test it together! I think it may make a big change in how some animals do the task. 

Yes, I think it's a big change.

alexpiet:
> The task structure is changed. Licking during the delay will restart the entire ITI and delay.

Why structure it this way? Why not simply extend the ITI after each lick?

> The format of B_DelayStartTime and B_DelayStartTimeHarp is changed from a matrix to a list, and each trial may have multiple values due to repeated delays. TransferToNWB.py needs to be updated to make it compatible with this change.

I dont understand the need to change the structure. Can you explain? Either way, you should update `TransferToNWB.py` before merging.

XX-Yin:
> > The task structure is changed. Licking during the delay will restart the entire ITI and delay.
> 
> Why structure it this way? Why not simply extend the ITI after each lick?
> 
> > The format of B_DelayStartTime and B_DelayStartTimeHarp is changed from a matrix to a list, and each trial may have multiple values due to repeated delays. TransferToNWB.py needs to be updated to make it compatible with this change.
> 
> I dont understand the need to change the structure. Can you explain? Either way, you should update `TransferToNWB.py` before merging.

Mice showed very high early licking rate (~50%), which is detrimental to experiments (ephys, ophys and optogenetics ...). Previously, we usually use small delay duration or fixed delay duration, and we think this is one of the reasons of high early licking rate. The punishment was minimal for small delay duration and some mice showed regular licking pattern (for example 2Hz) for fixed delay duration. 

Sue restarted the entire trial when detecting early licking. And mice trained with this task structure also showed low early licking rate (caveat here is there are other factor may also contribute to early licking, for example, sound decibel). This PR replicates Sue's task structure.

ZhixiaoSu:
The new delay structure has two advantages compared to the current one: 
1. Time out punishment is longer, so the mouse got more punishment when it licks before go cue comes;
2. Time out length is unpredictable, so the mouse won't have the 'hallicination' that go cue comes in ~1s when it stops licking after continuous licking. 

> > The task structure is changed. Licking during the delay will restart the entire ITI and delay.
> 
> Why structure it this way? Why not simply extend the ITI after each lick?
> 
Because this requirement can be too strict for the mouse. It can be hard for them to hold the licks right after harvesting reward.

alexpiet:
> The new delay structure has two advantages compared to the current one:
> 
> 1. Time out punishment is longer, so the mouse got more punishment when it licks before go cue comes;
> 2. Time out length is unpredictable, so the mouse won't have the 'hallicination' that go cue comes in ~1s when it stops licking after continuous licking.
> 
> > > The task structure is changed. Licking during the delay will restart the entire ITI and delay.
> > 
> > 
> > Why structure it this way? Why not simply extend the ITI after each lick?
> 
> Because this requirement can be too strict for the mouse. It can be hard for them to hold the licks right after harvesting reward.

Can you draw a diagram of how timing changes? 

I'm ok with whatever the distribution of timing you want, I'm mostly concerned about how you implemented it and breaking downstream code (like TransferToNWB)

ZhixiaoSu:
> > The new delay structure has two advantages compared to the current one:
> > 
> > 1. Time out punishment is longer, so the mouse got more punishment when it licks before go cue comes;
> > 2. Time out length is unpredictable, so the mouse won't have the 'hallicination' that go cue comes in ~1s when it stops licking after continuous licking.
> > 
> > > > The task structure is changed. Licking during the delay will restart the entire ITI and delay.
> > > 
> > > 
> > > Why structure it this way? Why not simply extend the ITI after each lick?
> > 
> > 
> > Because this requirement can be too strict for the mouse. It can be hard for them to hold the licks right after harvesting reward.
> 
> Can you draw a diagram of how timing changes?
Borrowing Han's slide:
image
The shaded area is time out punishment for early licks. In current version. it is a fixed value, 1s (another nolick DelayPeriod). We want to change it to exponential distribution + DelayPerod, so that the punishment is longer and timing of go cue is less predictable. 
 
> I'm ok with whatever the distribution of timing you want, I'm mostly concerned about how you implemented it and breaking downstream code (like TransferToNWB)

I agree. And I think that depends on how trials are defined in this task.

hanhou:
I also have some questions about the implementation

1. Is there a simpler way to implement the change without modifying the bonsai workflow too much? For example, can you add a general "early_lick_punishment" timeout node with a flexible length? This node is triggered by an early lick and always goes back to the start of the delay period. By doing this, you don't need to combine the "ITI" and "delay" into a new "ITIandDelay" node. There are at least two advantages: 
    1) after all, the early lick punishment should not be equated with restarting an ITI, given that we are not technically initiating a new trial and thus it doesn't make sense to create a new inter-_trial_-interval, 
    2) by not locking the early lick punishment to any existing trial structure like ITI or delay, we have the full control of the early lick punishment length on the trial-by-trial basis. For example, we can predefined its length in bonsai and pass it via the OSC message at the start of each trial like you do with other parameters. Then we still have the flexibility to set it to be either a new delay or a new ITI or any other values. (btw, did we generate exponential distribution for ITI and delay inside bonsai or in the GUI?)

2. If what I describe above is not feasible, is there any room for improvement in your current implementation?

image

I'm not a bonsai expert, so can you explain the operators in the yellow circles? What is GoCue2, DelayStart2, Pass2? What are Licks and Licks2? The left and right lick spouts? Why do we have an additional purple Pass for the upper branch of Licks?

3. A general comment: it would be great to always ask our bonsai expert @bruno-f-cruz to review out bonsai workflow.

XX-Yin:
> after all, the early lick punishment should not be equated with restarting an ITI, given that we are not technically initiating a new trial and thus it doesn't make sense to create a new inter-trial-interval,

You mean just restart the delay, not the ITI+delay? 

> by not locking the early lick punishment to any existing trial structure like ITI or delay, we have the full control of the early lick punishment length on the trial-by-trial basis. For example, we can predefined its length in the GUI and pass it via the OSC message at the start of each trial like you do with other parameters. Then we still have the flexibility to set it to be either a new delay or a new ITI or any other values. (btw, did we generate exponential distribution for ITI and delay inside bonsai or in the GUI?)

This is related to my initial question "I am updating the delay period. The proposed logic is when we detect early licks during the delay period, the entire ITI and delay will be restarted with the same parameters of the current trial. Here, I don't plan to redraw the exponential distribution of the ITI and delay. Otherwise, it will make the code more complex. Do you have any concerns?"

If we redrew the exponential distribution for ITI and delay for early licking, it would make the code much more complex. Right now, we actually have few changes.

> yellow circles? What is GoCue2, DelayStart2, Pass2? What are Licks and Licks2? 

The Pass node (yellow) is controlling if Licks can go through it to restart the ITI+Delay. The Pass node is controlled by the GoCue, DelayStart and purple Pass nodes. The original GoCue is a behavior subject, which will be triggered every time the ITIAndDelay (SelectMany) node is triggered, and this is not desired. So, I transferred the original GoCue, DelayStart and purple Pass nodes to publish subject to get the GoCue2, DelayStart2 and Pass2. GoCue2 will close the yellow Pass node, DelayStart2 will open the yellow Pass node and the Pass2 will close the Pass node. The Licks1 and Licks2 are corresponding to old Janelia lick detector and the new AIND lick detector.  

I'm happy to go through with you more details. 

That would be great if @bruno-f-cruz could help review the workflow.

XX-Yin:
> This change of trial structure is significant and we should be very cautious. I suggest we go through these steps:
> 
> 1. test the task logic with hand thoroughly
> 2. test with at least two mice and for at least one week of training, preferably one naive and one well-trained. This means only two days on production_test as usual may not be enough and we may need to make a special case for testing this change.
> 3. As Alex said, before even merging to develop, please also make corresponding changes in TransferToNWB. After some test data are collected at step 2, test NWB conversion and make sure all downstream pipeline still works, ie, the session should show up in the app. If not, I'll have to make further modifications to the pipeline.
> 4. Importantly, we should also make sure the modified TransferToNWB and downstream pipeline are still fully compatible with previous data format.

1. Sue and I will test it in the behavior rig. 
2. Let's see how we implement this. I think Sue already have a lot of sessions trained with this task structure. 
3 and 4. I've updated the GUI code so that we don't need to update the TransferNWB.py. I realize it's best to leave the TransferNWB.py file intact if possible.

XX-Yin:
> soon

The current licking pattern is terrible. We cannot make a meaningful interpretation from this.

XX-Yin:
> > soon
> 
> The current licking pattern is awful. We cannot make a meaningful interpretation from this.

I mean for later experiments (optogenetics, ephys ...)

ZhixiaoSu (commented):
Code has been tested with both finger play and mouse. And the nwb data generated from new code is normal.

ZhixiaoSu (approved):

alexpiet:
@XX-Yin @ZhixiaoSu This PR was a big update that @hanhou and I had both expressed some concerns about. In the future, before merging a PR, please get explicit approval from anyone who expresses concerns.

alexpiet:
> one is only merging to development, but not to production (main) yet.

FYI, It was pushed to `develop` last week, so its part of this week's testing (today), and will be merged into `main` tomorrow unless we raise an issue. 

I understand the urgency of fixing the mouse motivation. I think @hanhou and I mostly had concerns about how the data was being organized internally. Han was out last week, and I was deferring to his expertise. So we had a communication breakdown.

The current situation is fine, I just wanted to raise the issue for future updates.

XX-Yin:
@alexpiet I don't change the NWB file, so it should maintain the downstream analysis. I agree we need more agreements to merge cases like this in the future. Let's talk about it more tomorrow.

PR #381, opened 2024-04-13
Check drop frames

XX-Yin:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "aind-auto-train@git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main",
     "pynwb",
     "requests",
+    "harp-python",
 ]
 
 [tool.setuptools.packages.find]
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target
 git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main
 pynwb
 requests
+harp-python
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1978,7 +1978,7 @@ class Window(QMainWindow):
 
         if SaveContinue==0:
             # force to start a new session; Logging will stop and users cannot run new behaviors, but can still modify GUI parameters and save them.                 
-            self._NewSession()
+            self._NewSession(dont_ask=True)
             # do not create a new folder
             self.CreateNewFolder=0
         # check drop of frames
@@ -2720,21 +2720,22 @@ class Window(QMainWindow):
         except Exception as e:
             logging.error(str(e))
 
-    def _NewSession(self):
+    def _NewSession(self,dont_ask=False):
         logging.info('New Session pressed')
         self._StopCurrentSession() 
 
-        # If we have unsaved data, prompt to save
-        if (self.ToInitializeVisual==0) and (self.unsaved_data): 
-            reply = QMessageBox.critical(self, 
-                'Box {}, New Session:'.format(self.box_letter), 
-                'Start new session without saving?',
-                QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
-            if reply == QMessageBox.No:
-                self.NewSession.setStyleSheet("background-color : none")
-                self.NewSession.setChecked(False)
-                logging.info('New Session declined')
-                return False
+        if dont_ask==False:
+            # If we have unsaved data, prompt to save
+            if (self.ToInitializeVisual==0) and (self.unsaved_data): 
+                reply = QMessageBox.critical(self, 
+                    'Box {}, New Session:'.format(self.box_letter), 
+                    'Start new session without saving?',
+                    QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
+                if reply == QMessageBox.No:
+                    self.NewSession.setStyleSheet("background-color : none")
+                    self.NewSession.setChecked(False)
+                    logging.info('New Session declined')
+                    return False
         
         # stop the camera 
         self._stop_camera()
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -327,16 +327,25 @@ class Window(QMainWindow):
         self.Sessionlist.addItems(sorted_dates)
         self._connect_Sessionlist(connect=True)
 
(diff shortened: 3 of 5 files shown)


ZhixiaoSu (approved):
Please test with the main branch to make sure there's no conflict with the behavior boxes.

alexpiet:
@ZhixiaoSu @XX-Yin @hanhou This PR requires a manual installation step of the Harp package on every computer. We need to do this manual step before merging the PR

alexpiet:
What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.

XX-Yin:
> What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.


> @ZhixiaoSu @XX-Yin @hanhou This PR requires a manual installation step of the Harp package on every computer. We need to do this manual step before merging the PR. You also need to provide specific steps on how to install the Harp package

Yes, we need to install the harp-python package. We can directly do it through running `pip install harp-python` in the correct environment. I can do it for all behavior computers this afternoon.

XX-Yin:
> What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.

The logic is to check if the trigger number matches the camera frame number. I am working on a separate branch to correct the mismatched timestamps of the dropped frames session. We can't recover dropped frames.

alexpiet:
> > What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.
> 
> The logic is to check if the trigger number matches the camera frame number. I am working on a separate branch to correct the mismatched timestamps of the dropped frames session. We can't recover dropped frames.

Since we can't recover dropped frames, we should do this monitoring and correction in an analysis step, not in the GUI

XX-Yin:
> > > What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.
> > 
> > 
> > The logic is to check if the trigger number matches the camera frame number. I am working on a separate branch to correct the mismatched timestamps of the dropped frames session. We can't recover dropped frames.
> 
> Since we can't recover dropped frames, we should do this monitoring and correction in an analysis step, not in the GUI

We can correct the mismatched timestamps if we detect dropped frames. The analysis pipeline may take very long to happen. I think it's important to check the camera is correctly running by checking dropping of frames (e.g. camera (connection, settings ...), computer (disk space ...) and other parts (code) ).

alexpiet:
> > > > What is the logic of checking for dropped frames in the GUI? Can the user take any action to recover the dropped frames? I dont think so. This seems like an analysis issue, not a data collection issue.
> > > 
> > > 
> > > The logic is to check if the trigger number matches the camera frame number. I am working on a separate branch to correct the mismatched timestamps of the dropped frames session. We can't recover dropped frames.
> > 
> > 
> > Since we can't recover dropped frames, we should do this monitoring and correction in an analysis step, not in the GUI
> 
> We can correct the misaligned timestamps if we detect dropped frames. The analysis pipeline may take very long to happen. I think it's important to check the camera is correctly running by checking dropping of frames (e.g. camera (connection, settings ...), computer (disk space ...) and other parts (code) ).

- Can these checks be done at the start of the session?
- I still disagree this should be part of the GUI, but I'll let it get merged in.

XX-Yin:
"Can these checks be done at the start of the session?"

It’s much more difficult and laborious to check all of this before we start each session. There are also reasons that we couldn't predict before.

ZhixiaoSu:
Hi Alex, I think it is a quality check to help us know whether the frame started to drop as soon as it happens. Now, some videos drop frames and In the post-session analysis, it has been hard to infer what happened at that moment that caused the frame dropping. With this function, we might know if high cpu demand or memory limit or even certain human operations caused frame dropping.  Maybe it will not be useful in the future when we can infer the reasons for dropping offline. But now, it's nice to have.

alexpiet:
> Hi Alex, I think it is a quality check to help us know whether the frame started to drop as soon as it happens. Now, some videos drop frames and in the post-session analysis, it has been hard to infer what happened at that moment that caused the frame dropping. With this function, we might know if high cpu demand or memory limit or even certain human operations caused frame dropping. Maybe it will not be useful in the future when we can infer the reasons for dropping offline. But now, it's nice to have.

Can we check for dropped frames as the video is collected?

XX-Yin:
> > Hi Alex, I think it is a quality check to help us know whether the frame started to drop as soon as it happens. Now, some videos drop frames and in the post-session analysis, it has been hard to infer what happened at that moment that caused the frame dropping. With this function, we might know if high cpu demand or memory limit or even certain human operations caused frame dropping. Maybe it will not be useful in the future when we can infer the reasons for dropping offline. But now, it's nice to have.
> 
> Can we check for dropped frames as the video is collected?

We have IDs for each frame. The IDs will be missing for dropped frames.

ZhixiaoSu:
No we don’t check the video itself. We are checking the saved timestamps
for a video that usually equals frame number against the number of trigger.
And yes, as Xinxin said, for most of the frame-dropping session, frame ID
was not able to help realigning frame-dropping videos.

hanhou:
Although I see its diagnostic value, I also disagree that this should be a part of the GUI. This could be a standalone script running in the terminal after acquisition. We already put lots of things to the GUI and we have been thinking about simplifying it. 

Also, figuring out what factors contribute to frame loss should be a benchmarking step outside data acquisition, where you can do whatever stress tests without risking real data. If you rely on troubleshooting the system during real data collection, you may never know the limits of the system since you're not likely to push the system to its limits during data collection.

Btw, can you count the number of frames of a video stream directly in Bonsai? I think you should be able to do so?

XX-Yin:
> Although I see its diagnostic value, I also disagree that this should be a part of the GUI. This could be a standalone script running in the terminal after acquisition. We already put lots of things to the GUI and we have been thinking about simplifying it.
> 
> Also, figuring out what factors contribute to frame loss should be a benchmarking step outside data acquisition, where you can do whatever stress tests without risking real data. If you rely on troubleshooting the system during real data collection, you may never know the limits of the system since you're not likely to push the system to its limits during data collection.
> 
> Btw, can you count the number of frames of a video stream directly in Bonsai? I think you should be able to do so? I think it's even possible to plot the frame rate on line?

The bonsai can also be used to check drop of frames, but the visualization is not straightforward, and it's also inflexible to store the error related information to the json.

I think it's still worth integrating this function into the GUI until video capture is very stable.

PR #385, opened 2024-04-15
Revert "Check drop frames"

alexpiet:
Reverts AllenNeuralDynamics/dynamic-foraging-task#381
- This PR has a breaking change that needs to be addressed
- I disagree that the GUI should check for dropped frames at all. This PR adds two additional packages in order to implement an unneeded function


XX-Yin:
With this PR, we can immediately know if there are dropped frames at the end of the session. If there are no dropped frames, it means everything is ok. If we detect dropped frames, we can check the camera (connection, settings ...), computer (disk space ...) and other parts (code) immediately.

PR #391, opened 2024-04-17
Don’t delay 5 seconds for a non-high-speed video session

XX-Yin:
  
### Describe changes:
Previously, regardless of whether the session captured video, the code would sleep for 5 seconds waiting for the high-speed video data save to complete.

Now, only sleep 5 seconds for high-speed video session. 

Don't sleep 5s when saving data for non-high-speed video session.

For high-speed video session, the code will sleep 5s when saving the data. 
For non-high-speed video session, no influence. 

Tested in 323_EPHYS3.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -345,8 +345,6 @@ class Window(QMainWindow):
             use_default_folder_structure=0
             if save_tag==1:
                 # check the drop frames of the current session
-                # sleep some time to wait for the finish of saving video
-                time.sleep(5)
                 if hasattr(self,'HarpFolder'):
                     HarpFolder=self.HarpFolder
                     video_folder=self.VideoFolder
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -369,6 +367,8 @@ def _check_drop_frames(self,save_tag=1):
 
             camera_trigger_file=os.path.join(HarpFolder,'BehaviorEvents','Event_94.bin')
             if os.path.exists(camera_trigger_file):
+                # sleep some time to wait for the finish of saving video
+                time.sleep(5)
```
hanhou, line 371: How do you know that 5 seconds is enough? Is there a way to tell when the saving is done? (at least you can check the file size until it stops growing)

XX-Yin: From what I observed, when we stop logging for the current codec, the save completes almost immediately (within 2 seconds). I set it long (5s) to make it safer.

PR #395, opened 2024-04-17
Adding error tracking for untracked local files, and being on the wrong branch

### LINKED ISSUE #383: Generate Error in GUI log if local repo is dirty, or off main/production_testing
- Just for logging purposes

alexpiet:
  
### Describe changes:
- Right now, if a user has untracked local changes to the repository, or is running the GUI off the main branch we do not alert the user, or log that information as an error. 
- Now we add a note to the log on startup if there are untracked changes
- We log an error if the python version is not 3.9
- If a user starts a session AND the mouse is not "0", then:
    - we check for branch and untracked local changes. 
    - If the branch is not main or production testing, then the user is prompted. If they continue, an error is logged.
    - If the branch has untracked local changes, then the user is prompted. If they continue, an error is logged. 

- resolves #383 

- If you are running the code off pipeline, you will be alerted.
- If you are running the code with untracked changes to the code, you will be alerted

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3685,11 +3685,6 @@ if __name__ == "__main__":
     win.current_branch=current_branch
     win.repo_url=repo_url
     win.repo_dirty_flag=repo_dirty_flag
-    win.commit_ID=None
-    win.current_branch=None
-    win.repo_url=None
-    win.repo_dirty_flag=None
-
     win.show()
    
      # Run your application's event loop and stop after closing all windows
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3685,6 +3685,11 @@ if __name__ == "__main__":
     win.current_branch=current_branch
     win.repo_url=repo_url
     win.repo_dirty_flag=repo_dirty_flag
+    win.commit_ID=None
+    win.current_branch=None
+    win.repo_url=None
+    win.repo_dirty_flag=None
+
     win.show()
    
      # Run your application's event loop and stop after closing all windows
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3519,6 +3519,12 @@ def log_git_hash():
         git_branch = subprocess.check_output(['git','branch','--show-current']).decode('ascii').strip()
         repo_url = subprocess.check_output(['git', 'remote', 'get-url', 'origin']).decode('ascii').strip()
         logging.info('Current git commit branch, hash: {}, {}'.format(git_branch,git_hash))
+        print('Current git commit branch, hash: {}, {}'.format(git_branch,git_hash))
+        
+        py_version = sys.version
+        logging.info('Python version: {}'.format(py_version))
+        print('Python version: {}'.format(py_version))       
+
         return git_hash, git_branch, repo_url
     except Exception as e:
         logging.error('Could not log git branch and hash: {}'.format(str(e)))
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2987,7 +2987,7 @@ class Window(QMainWindow):
                     return                
                 else:
                     # Allow the session to continue, but log error
-                    logging.error('Starting session with untracked local changes: {}'.format(self.dirty_files)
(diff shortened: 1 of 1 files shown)


alexpiet:

src/foraging_gui/Foraging.py
```diff
@@ -3625,6 +3684,7 @@ def exception_hook(self, exc_type, exc_value, exc_traceback):
     win.commit_ID=commit_ID
     win.current_branch=current_branch
     win.repo_url=repo_url
+    win.repo_dirty_flag=repo_dirty_flag
```
XX-Yin, line 3687: Do you want to save the repo_dirty_flag to the json?

alexpiet: Yes. This is critical information

XX-Yin: It seems it's not saved?

alexpiet: Good catch, I've added it to the saved data

src/foraging_gui/Foraging.py
```diff
@@ -3514,15 +3549,40 @@ def log_git_hash():
...
+        logging.error('Incorrect version of python! Should be 3.9, got {}'.format(py_version[0:3]))
+
     try:
+        # Get information about task repository
         git_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip()
         git_branch = subprocess.check_output(['git','branch','--show-current']).decode('ascii').strip()
         repo_url = subprocess.check_output(['git', 'remote', 'get-url', 'origin']).decode('ascii').strip()
-        logging.info('Current git commit branch, hash: {}, {}'.format(git_branch,git_hash))
-        return git_hash, git_branch, repo_url
+        repo_dirty = subprocess.check_output(['git','diff-index','--name-only', 'HEAD']).decode('ascii').strip()
```
XX-Yin, line 3565: The foraging.bonsai or foraging.bonsai.layout will be changed every session if we save it (the harp COM ports; log folder ). The layout could also be different in different computers (e.g. visualizing the camera could be different). Some changes are automatic. I don't know if there is a good solution for this. I suggest we skip checking dirty for foraging.bonsai and foraging.bonsai.layout or just showing warning (instead of a popup) if we detect dirty for these two files.

alexpiet: I don't understand why the `foraging.bonsai` or `foraging.bonsai.layout` files would ever change:
- Changing the harp COM ports should be done through `ForagingSettings.json`, not the bonsai workflow
- Why would you change the log folder?

XX-Yin: The COM ports in the foraging.bonsai will be overwritten in each computer with different COM ports (one example shown below). 
image

alexpiet: But the COM ports are defined in the `Settings_Box.csv` files. Why do you change them in the bonsai workflow?

XX-Yin: We don't need to change the foraging.bonsai, but the bonsai changes it by default. If we save the bonsai workflow, these changes will be saved. We don't see the dirty information in the behavior rig because we don't save the changed bonsai workflow. 

If we hope to use a different layout to visualize the video capture and hope to save the layout, these changes in the foraging.bonsai (including com ports and log folder path) will also be saved automatically. 

I also suggest we save the list of dirty files.

alexpiet: @XX-Yin I've updated the code to save which files were dirty, both in the log, and in the `json`
- I don't understand why you would change the layout of the video capture. 
- If you change the com ports and the log folder path, this will be done through the `.csv` file, and wont result in the workflow changing.

alexpiet: Note that the user can always just press "ok" to continue running with untracked local changes

XX-Yin: It's ok for me to click extra "ok" button if there are some changes.

XX-Yin:
Do you why is Foraging.py recognized as a dirty file even though I haven't modified it?

image

XX-Yin:
A weird case happened on my side. If the foraging.bonsai or foraging.bonsai.layout is changed, the Foraging.py file will be labeled as dirty even no changes. It's normal if we change other python files.

XX-Yin (approved):

PR #434, opened 2024-05-09
FIP updates

### LINKED ISSUE #433: FIP workflow updates
- As we cannot close FIP bonsai window from the GUI, the button on the GUI doesn't need status indication(ie color chage). It's rather confusing.
- `Start Excitation` should be decoupled from `Start workflow` since the RAs should check for coupling before starting FIP workflow

### LINKED ISSUE #435: Ability to change FIP power levels
@ZhixiaoSu and @hagikent would like to be able to change the power level of the FIP excitation. This is a short term change based. Long term we would like to generalize the power level, but this is not necessary for the moment. (see #437 )
- How/where will this parameter be set?
   - It will default to low power, and a toggle window on the GUI will allow the user to save the power level which will get loaded for each session. One mouse will always use the same power level.  
   - We will use the terms `normal` and `axon` for the power level. We decided its better to discuss it at a high level, rather than `low` and `high` which feel like loaded terms
- How will this parameter be sent to Teensy? 
   - Right now `Start Excitation` sends a command `c` to the Teensy board. If we want to send high power, we will send `e` instead
   - Stopping excitation will be the same as now.

alexpiet:
  
### Describe changes:
- Previously, when a user pressed `Start FIP workflow` the button toggled colors. But this was confusing because the workflow could not be stopped from the GUI. Now the color will not change
- Previously, `Start Excitation` could not be pressed unless the FIP workflow had already been started. However, @hagikent requests that users check the cables (using excitation) before starting the workflow. So I removed the requirement that the FIP workflow had been started before allowing for excitation. The `Start` button still checks that the FIP workflow has been started. 
- Users can now the FIP mode between `Normal` and `Axon`. The only difference is the excitation power. `Normal` sends the Teensy command `c`, whereas `Axon` sends the teensy command `e` (@hagikent can you double check I got that mapping correct)
- Previously when QComboBox fields were modified by the user, the slot was connected to keyPressEvent which took no action. I added a function that simply logs the new value for debugging purposes. 
- Also resolves a small issue where loading a mouse does not "accept" the ID or experimenter fields (the text remain purple). I added a keyPressEvent which accepts them

- resolves #433 
- resolves #435 

- The `Start FIP Workflow` button will not change color after you press it
- You can press `Start Excitation` without starting the workflow first, to allow for checking cables. 
- When running an FIP mouse you need to set the `FIP mode`. This will default to `Normal`. Some mice for @ZhixiaoSu will have FIP mode = `Axon` for axonal imaging. An individual mouse will always have the same FIP mode. 

Standard UI
MicrosoftTeams-image3364726d096d2306de4288df6c20b2dcd57fa989631de8d5e32431472fb46a03

Xinxin's UI
MicrosoftTeams-image863c161a49bff4edff2a3921ab24d1b6971627c7796dbaed2d3bd880ebce4ac4

- none

- [x] tested in 428
- [x] Make sure it accurately changes the teensy command, and FIP power
- [x] Make sure the parameter is saved and loaded correctly

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2803,14 +2803,14 @@ class Window(QMainWindow):
             return 0 
 
         if self.StartExcitation.isChecked():
-            logging.info('StartExcitation is checked, photometry mode: {}'.format(self.FIPMode.CurrentText()))
+            logging.info('StartExcitation is checked, photometry mode: {}'.format(self.FIPMode.currentText()))
             self.StartExcitation.setStyleSheet("background-color : green;")
             try:
                 ser = serial.Serial(self.Teensy_COM, 9600, timeout=1)
                 # Trigger Teensy with the above specified exp mode
-                if self.FIPMode.CurrentText() == "Normal":
+                if self.FIPMode.currentText() == "Normal":
                     ser.write(b'c')
-                elif self.FIPMode.CurrentText() == "Axon":
+                elif self.FIPMode.currentText() == "Axon":
                     ser.write(b'e')       
                 ser.close()
                 self.TeensyWarning.setText('Started FIP excitation')
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1472,6 +1472,9 @@ class Window(QMainWindow):
             self.UpdateParameters=1 # Changes are allowed
             # change color to black
             for container in [self.TrainingParameters, self.centralwidget, self.Opto_dialog]:
+                for child in container.findChildren((QtWidegets.QComboBox)):
+                    print(child.objectName())
+
                 # Iterate over each child of the container that is a QLineEdit or QDoubleSpinBox
                 for child in container.findChildren((QtWidgets.QLineEdit,QtWidgets.QDoubleSpinBox,QtWidgets.QSpinBox)):
                     if child.objectName()=='qt_spinbox_lineedit':
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2782,16 +2782,6 @@ class Window(QMainWindow):
                QMessageBox.Ok )               
 
     def _StartExcitation(self):
-
-        if not self.FIP_started:
-            logging.warning('FIP workflow is not running, cannot start excitation')
-            reply = QMessageBox.information(self, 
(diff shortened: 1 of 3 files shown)


alexpiet:
@hagikent This is ready for review, and you should test to make sure it works with the updated Teensy 
@XX-Yin @hanhou This is mostly FIP updates, but I made a couple other changes you should review

src/foraging_gui/Foraging.py
```diff
@@ -205,8 +205,9 @@ def connectSignalsSlots(self):
         self.StartBleaching.clicked.connect(self._StartBleaching)
         self.NextBlock.clicked.connect(self._NextBlock)
         self.OptogeneticsB.activated.connect(self._OptogeneticsB) # turn on/off optogenetics
-        self.OptogeneticsB.currentIndexChanged.connect(self._keyPressEvent)
-        self.PhotometryB.currentIndexChanged.connect(self._keyPressEvent)
+        self.OptogeneticsB.currentIndexChanged.connect(lambda: self._QComboBoxUpdate('Optogenetics',self.OptogeneticsB.currentText()))
+        self.PhotometryB.currentIndexChanged.connect(lambda: self._QComboBoxUpdate('Photometry',self.PhotometryB.currentText()))
+        self.FIPMode.currentIndexChanged.connect(lambda: self._QComboBoxUpdate('FIPMode', self.FIPMode.currentText()))
```
XX-Yin, line 210: We may log the change for all `QComboBox` in a more general way. It doesn’t have to be done here.

        for container in [self.TrainingParameters, self.centralwidget, self.Opto_dialog]:
            # Iterate over each child of the container that is a QLineEdit or QDoubleSpinBox
            for child in container.findChildren((QtWidgets.QComboBox)):        
                child.currentIndexChanged.connect(lambda: self._QComboBoxUpdate())

XX-Yin (approved):
They are reasonable to me.

hagikent (approved):
Looks good. Seemingly working on 428-9-A with the updated Teensy software.

src/foraging_gui/Foraging.py
alexpiet: Well, we dont want to do it for all QComboboxes, because some of them we map to specific functions. I agree we should find a better solution

XX-Yin: I think one change can be mapped to multiple functions in parallel.

PR #447, opened 2024-05-14
Build rig json

### LINKED ISSUE #345: rig.json discussion
- We will add a template for the rig.json that will live in the `src/settings_templates/rig.json`
   - This template will have information for FIP, Opto, and behavior. 
- When we configure each behavior box, we will need to edit the `rig.json`, and then move it to `ForagingSettings`
   - The user will need to add/remove the relevant modalities (FIP, Opto)
   - There will be a rig.json for each behavior box, so we can name them `rig_box1.json`, etc
- The specific fields that will need to be modified for all behavior boxes:
   - modification_date
   - cameras/camera assembly/computer_name (if there are multiple cameras, need to change in multiple places)
   - cameras/camera_assembly/lens/serial_number (if available)
   - daqs/HarpDevice/computer_name
- The specific fields that will need to be modified for FIB rigs:
   - modalities (add Modality.FIB )
   - detectors/Detector/serial_number (these are CMOS FIB cameras, will be two Green/Red)
   - objectives/Objective/serial_number
- The specific fields that will need to be modified for Opto rigs:
   - ??
- The specific fields that need to be modified on Ephys rigs:
   - Lick detectors are AINDLickDetector, not Janelia_Lick_Detector
   - Cameras, hardware/drivers
- As part of the FIB excitation LED calibration, the `rig.json` needs to be updated
   - [ ] Make a note in the SOP for the FIB laser calibration (@hagikent)
- [ ] Add opto device information (@XX-Yin) to KH's draft
   - NiDAQ hardware/driver
   - LED/laser? 
- [x] Add Harp driver (@hagikent )
   - https://ftdichip.com/drivers/
   - "[FTDI driver](https://ftdichip.com/drivers/) (for Harp devices)"
   - "Install the version marked: "Click here to download the Windows 7 to Windows 11 and Windows Server driver installer"
- [x] Add newscale lick stages (@hagikent)
   - [x] travel, should be distance the newscale stage can move
   - [x] travel_unit
   - [x] serial number for newscale stage (can grab from ForagingSettings.json)
   - [x] firmware,  github repo/branch/commit
      - https://github.com/AllenNeuralDynamics/python-newscale, branch: axes-on-target, commit #7c17497
   - [x] USBXpress driver (https://www.silabs.com/documents/public/software/install_USBXpress_SDK.exe)
- [x] Update Computer configuration instructions based on this `rig.json` (@alexpiet )
- [x] Update GUI code (@alexpiet)
    - [x] The GUI will either (1) load the existing `rig.json` saved in `ForagingSettings`, or (2) generate a new `rig.json`
    - [x] The `rig.json` will need to be regenerated if we perform calibration on the laser, FIB, or water; if any part with a serial number changes.
    - [x] The GUI can check these things by loading the existing json, checking against the calibration files/task files.
    - [ ] After getting a valid `rig.json` it will be copied to the metadata directory of the saved data folder

TODO after PR #447
- [x] Need to pin version of aind-data-schema  
- [ ] fill in what details are specified in the rig_specification file
- [x] add instructions for how/when `rig_specification_<box_number>.json` should be updated
- [x] Determine what data modalities are included (video? Opto?)
- [x] Camera entries are fixed, unclear if they are for 428 or 447 @hagikent 
- [x] AIND lick spout motor stage details are missing
- [x] Speaker details are missing
- [x] Laser calibrations are missing - unclear if its for opto or FIP
- [x] FIP entries are fixed, and dont load from rig_specification
- [x] Harp devices are missing, including sound card, expander, synchronization board, etc
- [x] Optogenetic details are missing

alexpiet:
  
### Describe changes:
- Adds `Foraging._LoadRigJson` that ensures there is an up to date rig.json
    - If no rig.json file exists locally, then one will be created
    - If a rig.json file already exists, we check to see if it is up to date
    - `ForagingSettings.json` includes a toggle for not making a json: `create_rig_metadata`
    - "up to date" is determined by rig specific details that are written in `ForagingSettings.json` and `Settings_box#.csv`
    - Does NOT include this rig.json metadata file in the saved data. 
    - Checks if this is an FIP rig by checking if a Teensy COM is set for this box
    - Does not include any information about ephys rigs. Ephys rigs should set `create_rig_metdata` to False, and supply their own rig.json
- Uses naming scheme of `rig_<rig_name>_YYYY-MM-DD_HH_MM_SS.json`
    - rig_name is `current_box` in ForagingSettings, except for 447-1/2/3, which have `-A/B/C/D` appended 
- The following fields are now in `ForagingSettings.json`:
    - `rig_metadata_folder`, optional, defaults to `ForagingSettings/rig_metadata`
    - `create_rig_metadata`, optional, defaults to True
    - `FIP_settings`, optional, defaults to `<user>/Documents/FIPSettings/` only used for FIP serial numbers
- The following fields are now in `Settings_box#.csv`:
    - `HasOpto`, optional,  If set to "1", then opto metadata will be included
    - `FipObjectiveSerialNumber`, optional. If FIP rig, then must include 
- Future proofing:
    - There has been discussion about hosting rig.json files in a central database (like SLIMS). If that happens, then we just change where we load the rig.json file from (SLIMS, instead of the rig_metadata folder), and everything else works the same
    - If we move rig.json generation to a separate repository so we can align with ephys rigs (See discussion in #460), then we just replace the function call `RigJsonBuilder.build_rig_json_core()` with a call to the separate repository, and everything else works the same

- resolves #345 

- No changes in daily operation. The rig.json generation should happen on startup in the background
- If a piece of hardware that is tracked in `Settings_box#.csv` is replaced, then the updated information needs to be provided. 

- Need to install three code packages
    - `conda activate Foraging`
    - `pip install deepdiff`
    - `pip install aind-data-schema==0.36.0`
    - `pip install aind-data-schema-models==0.1.5`
-  Need to add entries to `ForagingSettings.json` and `Settings_box#.csv` as needed
- [ ] 446/447 @alexpiet
- [ ] Ephys1 @KanghoonJ 
- [ ] Ephys3 @XX-Yin 
- [ ] 428 @hagikent 

- [x] tested in 447


XX-Yin:
@alexpiet Thanks for making this rig.json. I think there are repeated efforts. We should talk about how to combine them. 

I also developed code to flexibly generate the rig.json (linked below). The logic is devices are shared in the `local_devices.py`, and we need to list what devices to use in the `fields_for_generating_rig_metadata.json`, then the `generate_rig_metadata.py` will generate rig metadata. I have used this code to generate the rig.json for ephys 3. It could be extended to other rigs.  

https://github.com/AllenNeuralDynamics/CTLUT-metadata-gen/blob/generate_rig_metadata/src/CTLUT_metadata_gen/generate_rig_metadata.py
https://github.com/AllenNeuralDynamics/CTLUT-metadata-gen/blob/generate_rig_metadata/src/CTLUT_metadata_gen/local_devices.py
https://github.com/AllenNeuralDynamics/CTLUT-metadata-gen/blob/generate_rig_metadata/src/CTLUT_metadata_gen/fields_for_generating_rig_metadata.json

XX-Yin:
Can you explain more how do you ensure the rig.json is up to date? The rig.json could be very different in different rigs. 

My general opinion is the rig.json should not be created from the behavior GUI, but from an independent source. Reasons are 1) rig.json is independent from the behavior. 2) There are a lot of variations across rigs, which requires providing extra information to create rig.json (my solution is to define local devices, fields to select which devices to use and a script to define rules).

alexpiet:
> Can you explain more how do you ensure the rig.json is up to date? The rig.json could be very different in different rigs.
> 
> My general opinion is the rig.json should not be created from the behavior GUI, but from an independent source. Reasons are 1) rig.json is independent from the behavior. 2) There are a lot of variations across rigs, which requires providing extra information to create rig.json (my solution is to define local devices, fields to select which devices to use and a script to define rules).

I ensure the rig.json is up to date by generating a new rig.json and seeing if anything changed other than the date. 

I generate the new rig.json in a very similar manner to how you generate them. Each rig has details provided in `ForagingSettings.json`, `Settings_Box.csv`, and then any additional details can be added to `rig_specification.json`. Those settings, along with any calibration files, are equivalent to the "fields" in your set up, then I have a function that generates the rig.json (equivalent to your local devices and script). 

There are two big advantages of generating the rig.json in the GUI. (1) It happens automatically without a user having to generate a new rig.json on every rig. (2) When a user updates the rig, most of those changes are already going to be reflected in ForagingSettings.json, or Settings_Box.csv, so it makes sense to also log them in rig_specification.json. 

Any rig can opt-out of this process, and generate their own rig.json. I imagine the Ephys rigs will want to do this since the rig.json will be much more complicated.

XX-Yin:
I understand you want the rig.json to be automatically generated when changes occur. I agree with this idea. However, this can be an independent process. It will make the maintenance of the code very hard if we integrate "generating rig.json" to the behavior GUI. The main reason is the devices and the code to generate the rig.json are very different across different rigs. If they were all hard-coded in the behavioral GUI, we would need to constantly update the code. For example, there are more fields that need to be added to rig.json (e.g. LED for optogenetics, IR light, patch cable, serial numbers of devices).

Currently, everyone writes their own code to generate rig.json. I think we should have a unified repository to standardize this process. My idea is to create a shared device repository and write standardized code to flexibly generate the rig.json based on these local devices along with an additional file to select the device we wish to use. We don't need to do it now.

hagikent:
My understanding is that this function is "Opto-out". So, ephys rigs can be simply excluded.
For 24 boxes in the 446/447, I think it makes sense to automatize this as they are virtually identical except for Opto+/- and FIP+/-.

hanhou:
I agree that we can merge this for now. From Alex's teams message, my understanding is that this PR is still a work in progress for rig.json but can move things forward. Let's think about how to deal with ephys rigs later (similar question applies to session.json).

alexpiet:
> The main reason is the devices and the code to generate the rig.json are very different across different rigs. If they were all hard-coded in the behavioral GUI, we would need to constantly update the code. For example, there are more fields that need to be added to rig.json (e.g. LED for optogenetics, IR light, patch cable, serial numbers of devices).

In 446/447/428 there are effectively three types of rigs: behavior, opto, FIP. For each of those rig types, the only thing we need to handle is the exactly serial number for each part. This can easily be included in the `rig_specification.json`. This hardware gets changed very infrequently, so we don't need to update the code often. We will only need to update the code when a new type of device is added (and maybe we can reorganize my code so we dont need to do that). 

The ephys rigs can opt out of this process and generate their own rig.json files

> Currently, everyone writes their own code to generate rig.json. I think we should have a unified repository to standardize this process. My idea is to create a shared device repository and write standardized code to flexibly generate the rig.json based on these local devices along with an additional file to select the device we wish to use. We don't need to do it now.

I think this is the start of that process. Its a work in progress. However, we need to start generating rig.jsons asap so we can move to the automatic data transfer process.

XX-Yin:
> I agree that we can merge this for now. From Alex's teams message, my understanding is that this PR is still a work in progress for rig.json but can move things forward. Let's think about how to deal with ephys rigs later (similar question applies to session.json).

Most of the information in the session json requires fields from the behavior GUI, but the rig.json is independent. My main concern is it will make the maintenance very difficult, and most likely we will have a centralized repo to create rig.json.

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,510 @@
...
+            camera_target=d.CameraTarget.FACE_SIDE_RIGHT,
+            camera=d.Camera(
+                name="Side face camera",
+                detector_type="Camera",
+                serial_number="TBD",
+                manufacturer=d.Organization.AILIPU,
+                model="ELP-USBFHD05MT-KL170IR",
+                notes="The light intensity sensor was removed; IR illumination is constantly on",
+                data_interface="USB",
+                computer_name="W10DTJK7N0M3",
```
XX-Yin, line 90: computer name is varied.

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,510 @@
...
+            lens=d.Lens(
+                name="Xenocam 1",
+                model="XC0922LENS",
+                serial_number="unknown",
+                manufacturer=d.Organization.OTHER,
+                max_aperture="f/1.4",
+                notes='Focal Length 9-22mm 1/3" IR F1.4',
+            ),
+        ),
+        d.CameraAssembly(
```
XX-Yin, line 108: The camera assembly could be varied (e.g. high-speed camera assembly and the current camera assembly used in 447). For high-speed camera assembly, the side and bottom cameras used different lenses and filters.

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,510 @@
...
+                manufacturer=d.Organization.THORLABS,
+                model="M565F3",
+                wavelength=565,
+            ),
+        ]
+
+        components['detectors']=[
+            d.Detector(
+                name="Green CMOS",
+                serial_number="21396991",
```
XX-Yin, line 277: The serial_number could be different.

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,510 @@
...
+    ###########################################################################
+    # Assemble rig schema
+    rig = r.Rig(
+        rig_id="{}_{}".format(settings['rig_name'],datetime.now().strftime('%Y-%m-%d')), 
+        modification_date=date.today(),
+        **components 
+        )
+    logging.info('finished building rig json')
+    return rig
+
```
XX-Yin, line 485: There are other devices (e.g. harp devices, USB daq for optogenetics and IR light) we need to add.

XX-Yin:
> > The main reason is the devices and the code to generate the rig.json are very different across different rigs. If they were all hard-coded in the behavioral GUI, we would need to constantly update the code. For example, there are more fields that need to be added to rig.json (e.g. LED for optogenetics, IR light, patch cable, serial numbers of devices).
> 
> In 446/447/428 there are effectively three types of rigs: behavior, opto, FIP. For each of those rig types, the only thing we need to handle is the exactly serial number for each part. This can easily be included in the `rig_specification.json`. This hardware gets changed very infrequently, so we don't need to update the code often. We will only need to update the code when a new type of device is added (and maybe we can reorganize my code so we dont need to do that).
> 
> The ephys rigs can opt out of this process and generate their own rig.json files
> 
> > Currently, everyone writes their own code to generate rig.json. I think we should have a unified repository to standardize this process. My idea is to create a shared device repository and write standardized code to flexibly generate the rig.json based on these local devices along with an additional file to select the device we wish to use. We don't need to do it now.
> 
> I think this is the start of that process. Its a work in progress. However, we need to start generating rig.jsons asap so we can move to the automatic data transfer process.


I understand the urgency to generate the rig.json. The rig.json can be generated independent (e.g. using the code I wrote). My suggestion is that the behavior GUI should still determine the update of rig.json (not necessarily through comparing the new and old rig.json), but the code to generate the rig.json should be kept independently to enable flexibility.     

If we don't need to change the rig.json very frequently. Why not create a static rig.json? And we only need to update the calibration part every time when there is a new calibration.

src/foraging_gui/RigJsonBuilder.py
hagikent: (just copy pasting a post in Teams chat) I believe I have communicated with you that my template lacks optogenetics-related stuff and need to be added.

alexpiet:
@XX-Yin @hagikent @hanhou This is ready for review again. It now fills out all the details of the rig.json

src/foraging_gui/RigJsonBuilder.py
alexpiet: I've added these additional devices

src/foraging_gui/RigJsonBuilder.py
alexpiet: This is resolved now

src/foraging_gui/RigJsonBuilder.py
alexpiet: This is resolved now

src/foraging_gui/RigJsonBuilder.py
alexpiet: This is resolved now

hanhou (approved):

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,681 @@
...
+        d.HarpDevice(
+            name="Harp Behavior",
+            harp_device_type=d.HarpDeviceType.BEHAVIOR,
+            manufacturer=d.Organization.CHAMPALIMAUD,
+            core_version="2.1",
+            firmware_version="FTDI version:",
+            computer_name=settings['computer_name'], 
+            is_clock_generator=False,
+            channels=[
+                d.DAQChannel(channel_name="DO0", device_name="Solenoid Left", channel_type="Digital Output"),
```
XX-Yin, line 290: This is not correct. Both Solenoid and lick spouts are both via Ethernet cable.

src/foraging_gui/RigJsonBuilder.py
```diff
@@ -0,0 +1,681 @@
...
+
+    # DAQS
+    ###########################################################################
+
+    components['daqs']=[
+        d.HarpDevice(
+            name="Harp Behavior",
+            harp_device_type=d.HarpDeviceType.BEHAVIOR,
+            manufacturer=d.Organization.CHAMPALIMAUD,
+            core_version="2.1",
```
XX-Yin, line 285: the `core_version` is not necessarily "2.1". Most of them are 2.0. We should check for each behavior board.

alexpiet: They aren't all the same!? That seems like something we should fix

src/foraging_gui/RigJsonBuilder.py
alexpiet: I think @hagikent wrote this part. What is the correct way to document how the solenoid and lick spouts are connected? I could write `data_interface=d.DataInterface.ETH` which corrects the issue you mentioned, but then its not clear how the solenoids and lick spouts are connected to the behavior board

src/foraging_gui/RigJsonBuilder.py
hagikent: this is also new to me...thought all the same. I can check ones in 446/447/428 tomorrow. (and maybe ask Helio)

src/foraging_gui/RigJsonBuilder.py
hagikent: I remember I wrote this with Saskia's input during the Ephys-turned-metadata Hackathon.
Why Ethernet cables need to be documented? The relevant info would be port/channel number?

XX-Yin: DO0 and DO1 are other channels in the behavior board, not from the Ethernet cables.

src/foraging_gui/RigJsonBuilder.py
alexpiet: Do you know how to check the version in code?

alexpiet: I removed core_version, and firmware_version. They are listed as optional, and since they might be different I wont set them unless we can set them automatically

XX-Yin: I don't know how to check the version through bonsai code. It's ok to remove the core_version at this moment.

src/foraging_gui/RigJsonBuilder.py
alexpiet: Yes, Xinxin is correct. DO0 and DO1 are separate ports from the ethernet ports.

How about:

`
   d.HarpDevice(
            name="Harp Behavior",
            harp_device_type=d.HarpDeviceType.BEHAVIOR,
            manufacturer=d.Organization.CHAMPALIMAUD,
            computer_name=settings['computer_name'], 
            is_clock_generator=False,
            data_interface = d.DataInterface.ETH,
            notes="Lick detectors, and reward delivery solenoids are connected via ethernet cables"
            )
 `

src/foraging_gui/RigJsonBuilder.py
alexpiet: Bruno gave me some tips on how to get the core version, firmware version, and hardware version in code. I'll make a separate issue

src/foraging_gui/RigJsonBuilder.py
XX-Yin: yeah, looks good.

alexpiet: Updated

hagikent: I don't know what's the official name of those Eth ports but can we make it consistent with how we call it in software? In bonsai, they are called `DigitalOutput/SupplyPort0(or1)` and Saskia and I decided to define DO0or1.
Otherwise super confusing especially when making the updated diagram.

Screenshot 2024-05-22 at 16 00 23

alexpiet:
@XX-Yin @hanhou @hagikent 
I tested this on 1C and 1D. I think we are ready to merge

src/foraging_gui/RigJsonBuilder.py
hagikent: And to be a bit more informative for @alexpiet , two ethernet cables are, in parallel, connecting HarpBehavior and Janelia-lick-detector(+solenoid activator). 0 and 1 are corresponding to L and R solenoids, respectively.

alexpiet: I dont have any idea what we should call it. All I know is that DO0 and DO1 are physically marked on the boards at a different port

hagikent: >DO0, DI1, 

I see, seems yes I confirmed in the below link....

According to this, hardware-wise they are apparently called `Poke` as intended usage is to connect NosePoke device?
https://github.com/harp-tech/device.behavior

@XX-Yin 
Why at the software level they are SupplyPort0? Shouldn't be matching? Too confusing.

XX-Yin: This is real.  Different naming can be confusing.

 I think the `Poke` is more related to the tasks they originally design for. They are multiple connections integrated in the Ethernet connector, and `SupplyPort0` is more related to the general function, from one of the ethernet connections, to `give 12v voltage output`.

XX-Yin: > This is real. Different naming can be confusing.
> 
> I think the `Poke` is more related to the tasks they are originally designed for. They are multiple connections integrated in the Ethernet connector, and `SupplyPort0` is more related to the general function, from one of the ethernet connections, to `give 12v voltage output`.

XX-Yin:
> @XX-Yin @hanhou @hagikent I tested this on 1C and 1D. I think we are ready to merge

Okay, let's merge it and test the session metadata generation tomorrow.

alexpiet:
> > @XX-Yin @hanhou @hagikent I tested this on 1C and 1D. I think we are ready to merge
> 
> Okay, let's merge it and test the session metadata generation tomorrow.

I made one quick change that I need to test in 447 to automatically load the FIP serial numbers. Then I will merge and test the session.json generation

PR #451, opened 2024-05-15
Stopping excitation before saving

### LINKED ISSUE #450: Photometry end time is not specified
When photometry is running, we should either stop excitation when trials stop, or when the user tries to save. @hagikent any opinion? I think this is related to https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/380

alexpiet:
  
### Describe changes:
- Forces photo excitation to stop before saving. 
- I believe when excitation is not stopped before saving a race condition can happen when excitation messages continue to be generated and saving can stall. Forcing the excitation to stop before saving may prevent this, its also just better to control the excitation process. 

- resolves #450 
- may resolve https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/380

- none

- none

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2037,6 +2037,12 @@ class Window(QMainWindow):
                 self.WarningLabel.setStyleSheet(self.default_warning_color)
                 return
 
+        # Stop Excitation if its running
+        if self.StartExcitation.isChecked():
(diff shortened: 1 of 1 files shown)


hagikent (approved):
nice fool-proof. There might be a better design but would work as a great quick fix. thanks.

PR #452, opened 2024-05-16
Add rwd delay to nwb (test PR)

ZhixiaoSu:
The nwb file can be a bit misleading when the session has a '**reward delay**', which is a fixed delay between response and reward. 

In current version, '**reward_outome_time**' is the time when reward is determined. If a choice is made, it's time of first lick, if not, it's time of end of lick window. It makes sense in sessions where there's no reward delay. But when we introduce a reward delay, this time should be called '**response_time**' or 'decision_time' and additional '**reward_delay**' information (not saved in nwb yet) should be added to calculate time of outcome. 
 
Given how big impact this file may have downstream, minimum change is made here to enable analysis. One column called '**reward_delay**'  was added to NWB. 

Btw, one questions about our current behavior pipeline: if we update the generate nwb function, do previous sessions get re-processed?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,8 @@ dependencies = [
     "requests",
     "harp-python",
     "deepdiff",
+    "aind-data-schema==0.36.0",
+    "aind-data-schema-models==0.1.1"
 ]
 
 [tool.setuptools.packages.find]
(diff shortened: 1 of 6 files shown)


ZhixiaoSu:
Re-do starting from develop branch.

PR #454, opened 2024-05-16
Add reward delivery to dev

ZhixiaoSu:
  
### Describe changes:
1. Added 'reward delay' field to nwb.trials. It is a delay between choice time and reward delivery.
2. Additional description about 'reward_outcome_time': It is time when choice is registered. 
This columns makes it possible to calculate actual reward delivery time when there's a reward delay in session.
There will be no change in behavior. 
There will be no manual update,
No, it is tested locally on my laptop and desktop. 

Additional info:

In current version, 'reward_outome_time' is the time when reward is determined, not when it is delivered. If a choice is made, it's time of first lick, if not, it's time of end of lick window. It makes sense in sessions where there's no reward delay. But when we introduce a reward delay, this time should be called 'response_time' or 'decision_time' and additional 'reward_delay' information (not saved in nwb yet) should be added to calculate time of outcome.

Given how big impact this file may have downstream, minimum change is made here to enable analysis. One column called 'reward_delay' was added to NWB.

### DIFF
--- a/src/foraging_gui/TransferToNWB.py
+++ b/src/foraging_gui/TransferToNWB.py
@@ -204,7 +204,7 @@ def bonsai_to_nwb(fname, save_folder=save_folder):
     nwbfile.add_trial_column(name='rewarded_historyR', description=f'The reward history of right lick port')
     nwbfile.add_trial_column(name='delay_start_time', description=f'The delay start time')
     nwbfile.add_trial_column(name='goCue_start_time', description=f'The go cue start time')
-    nwbfile.add_trial_column(name='reward_outcome_time', description=f'The reward outcome time (reward/no reward/no response)')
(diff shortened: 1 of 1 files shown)


alexpiet (approved):

XX-Yin:
@ZhixiaoSu There is an edge case. The old session does not have `obj.TP_RewardDelay`.

ZhixiaoSu:
> @ZhixiaoSu There is an edge case. The old session does not have `obj.TP_RewardDelay`.

@XX-Yin Can I fix it by first checking if it has the field? If it doesn't, I'll put it as 0.0.

XX-Yin:
> 0.0

Yes, it will work. You can run the file directly to pass the test.

ZhixiaoSu:
Update: set default to 0 to take care of earlier sessions without reward delay parameter by setting it to 0.

XX-Yin (approved):

PR #475, opened 2024-05-29
Sound intensity calibration function

hagikent:
### Describe changes:
Attenuation function is added to GoCueChecker.bonsai in src\soundcard_setup

image

If you navigate to `AttenuationPayload`,

image

you will be able to attenuate sound intensity for both L/R channels separately. (larger attenuation value= smaller sound)

This snippet should be used for both checking if sound upload properly worked and sound intensity calibration (e.g. GoCue~60dB)

With this, precise calibration of sound intensity per rig can be possible.

Also, automatically solving this:
https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/396

No. Only when calibrating sound intensity.

No.
(main foraging.bonsai will be updated accordingly to integrate attenuation and then externalize the values in local setting files)

Tested in 428-9-A


alexpiet:
@hagikent After we determine the appropriate attenuation, how/where do we save that value?

hagikent:
https://teams.microsoft.com/l/message/19:3f4fca20f4d34720a0ae8b90a6a75b22@thread.v2/1716915637194?context=%7B%22contextType%22%3A%22chat%22%7D

I was thinking of "_...and storing the attenuation value in Setting.json or Box.csv?_" and Xinxin seems to be agreeing. Or could be a different calibration file like the solenoid calibration file.

XX-Yin:
> calibration



> @alexpiet https://teams.microsoft.com/l/message/19:3f4fca20f4d34720a0ae8b90a6a75b22@thread.v2/1716915637194?context=%7B%22contextType%22%3A%22chat%22%7D
> 
> I was thinking of "_...and storing the attenuation value in Setting.json or Box.csv?_" and Xinxin seems to be agreeing. Or could be a different calibration file like the solenoid calibration file.

I think these values can go to the setting csv file

hagikent:
I don't have any preference. The most intuitive, easiest to maintain location would be the best.
Anyway, this PR is agnostic to how to actually store calibrated values.

alexpiet:
@hagikent Can you go ahead an implement the system for saving the calibration values?

hagikent:
@alexpiet 
As we only need a value corresponding to 60dB with a fixed frequency, I was just thinking of manually adding it to Box.cvs after calibration. Are you expecting something more?

alexpiet:
Yes, that seems great. And just figuring out how to get bonsai to read that value

hagikent:
Attenuation node can externalize the value (see below). 
Screen Shot 2024-05-29 at 13 29 43

So, just sending value directly from CLI.
Or adding csvreader to make Bonsai read it.
Screen Shot 2024-05-29 at 13 38 32

I don't know which one goes smoothly with other "parameter externalization" you two already implemented, though.

alexpiet:
Add it to Settings_box#.csv

hagikent:
Merged into #481

PR #481, opened 2024-05-30
Sound intensity calibration function and corresponding setting file

hagikent:
### Describe changes:
1.Dedicated setting file to specify attenuation value - `SoundAttenuation.csv` in the templates folder
2. `GoCueChecker.bonsai` is updated to read this file.
Screen Shot 2024-05-30 at 13 58 25

Main foraging-bonsai should follow this change.
@XX-Yin @alexpiet 

Sound Intensity control/standardization across rigs.

No. Only during deploying.

`SoundAttenuation.csv` template should be copy pasted to `Documents\ForagingSettings` folder.
Tested in 428


XX-Yin:
@hagikent Is the purpose of this PR to help perform calibration? 

To make changes to the existing bonsai workflow, we can directly save the `AttenuationLeft` to the `Settings_box#.csv`. The bonsai workflow already reads the `Settings_box#.csv`. no need to create another csv file for each rig.

hagikent:
@XX-Yin 
>Is the purpose of this PR to help perform calibration?

Yes, then the value will be copy pasted to `Settings_box#.csv`.
As Scis/RAs will be changing the value multiple times during calibration, keeping this file separate from main setting file would be safer to avoid erroneous change in other entries. Please add the `SoundAttenuation` entry when you update the main foraging.bonsai. Default should be 0 for now (at least while "0.08" file stay in 447).

hagikent:
please approve if this makes sense and we can go forward (so that we can start calibration in `main` from next Wed)

XX-Yin (approved):

PR #485, opened 2024-05-30
Backup data

XX-Yin:
  
### Describe changes:
1. Back up the data each trial. 
2. The code for saving backup data runs in a different thread and does not affect the operation of behavior propagation in the main thread. 
3. The data format is the same as normal saving process. 

Back up the data in case the software crashes

No

The backup is turned on by default. If you want to turn it off, set `save_each_trial` in the `ForagingSettings.json` to `False`.

No

- [x] Tested on my own computer
- [x] Tested in 447
- [x] Tested in 323_EPHYS3

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2302,16 +2302,14 @@ class Window(QMainWindow):
                     else:
                         Value=getattr(self.GeneratedTrials, attr_name)
                         try:
-                            if math.isnan(Value):
-                                Obj[attr_name]='nan'
+                            if isinstance(Value, float) or isinstance(Value, int):                                
+                                if math.isnan(Value):
+                                    Obj[attr_name]='nan'    
+                                else:
+                                    Obj[attr_name]=Value   
                             else:
-                                Obj[attr_name]=Value
+                                Obj[attr_name]=Value        
                         except Exception as e:
-                            # Lots of B_xxx data are not real scalars and thus math.isnan(Value) will fail
-                            # e.g. B_AnimalResponseHistory. We just save them as they are. 
-                            # This is expected and no need to log an error.
-                            # I don't know the necessity of turning nan values into 'nan', 
-                            # but for backward compatibility, I keep it above.
                             logging.info(f'{attr_name} is not a real scalar, save it as it is.')
                             Obj[attr_name]=Value
         # save other events, e.g. session start time
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -94,7 +94,7 @@ class Window(QMainWindow):
         self.ToInitializeVisual = 1 # Should we visualize performance
         self.FigureUpdateTooSlow = 0# if the FigureUpdateTooSlow is true, using different process to update figures
         self.ANewTrial = 1          # permission to start a new trial
-        self.to_save = 1            # permission to save backup data
+        self.save_backup_allowed = 1   # permission to save backup data; 0, the previous saving has not finished, and it will not trigger the next saving; 1, it is allowed to save backup data
         self.UpdateParameters = 1   # permission to update parameters
         self.loggingstarted = -1    # Have we started trial logging
         self.unsaved_data = False   # Setting unsaved data to False 
@@ -3302,7 +3302,7 @@ class Window(QMainWindow):
     
     def _thread_complete6(self):
         '''complete of save data'''
-        self.to_save=1
+        self.save_backup_allowed=1
 
     def _thread_complete_timer(self):
         '''complete of _Timer'''
@@ -3692,8 +3692,8 @@ class Window(QMainWindow):
                 if self.actionLicks_sta.isChecked():
                     self.PlotLick._Update(GeneratedTrials=GeneratedTrials)
                 # save the data everytrial
-                if GeneratedTrials.B_CurrentTrialN>0 and self.to_save==1 and self.save_each_trial:
-                    self.to_save=0
+                if GeneratedTrials.B_CurrentTrialN>0 and self.save_backup_allowed==1 and self.save_each_trial:
+                    self.save_backup_allowed=0
                     self.threadpool6.start(worker_save)
 
                 if GeneratedTrials.CurrentSimulation==True:
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2189,6 +2189,11 @@ class Window(QMainWindow):
             ForceSave=1
             SaveAs=0
             SaveContinue=1
+            saving_type_label = 'backup saving'
+        elif ForceSave==1:
+            saving_type_label = 'force saving'
+        else:
+            saving_type_label = 'normal saving'
 
         logging.info('Saving current session, ForceSave={}'.format(ForceSave))
         if ForceSave==0:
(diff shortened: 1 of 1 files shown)


XX-Yin:
@yonibrowning @alexpiet @hanhou @galenlynch 

I think we should merge this first to use this backup function. If it's really necessary to save the single trial, we can update it later.

src/foraging_gui/Foraging.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         self.ToInitializeVisual = 1 # Should we visualize performance
         self.FigureUpdateTooSlow = 0# if the FigureUpdateTooSlow is true, using different process to update figures
         self.ANewTrial = 1          # permission to start a new trial
+        self.to_save = 1            # permission to save backup data
```
hanhou, line 97: please use more explicit name such as `save_backup_allowed`, although I don't quite understand the function of this flag. Could you add more documentation?

src/foraging_gui/Foraging.py
```diff
@@ -3549,12 +3572,15 @@ def _Start(self):
             workerGenerateAtrial.signals.finished.connect(self._thread_complete4)
             workerStartTrialLoop = Worker(self._StartTrialLoop,GeneratedTrials,worker1,workerPlot,workerGenerateAtrial)
             workerStartTrialLoop1 = Worker(self._StartTrialLoop1,GeneratedTrials)
+            worker_save = Worker(self._Save,BackupSave=1)
+            worker_save.signals.finished.connect(self._thread_complete5)
```
hanhou, line 3576: should this be `_thread_complete6` for consistency?

hanhou (commented):
Please also make sure that this backup saving mechanism does not interfere with the normal saving process at the end of the session.

XX-Yin:
> Please also make sure that this backup saving mechanism does not interfere with the normal saving process at the end of the session.
> 
> Specifically:
> 
> * If the session crashes, what part of data will be saved and what part will be missing (at least the post weight)? How about meta data?
If the session crashes, the data not entered to the GUI will not be saved (including the post weight). The meta data will be saved. The backup saving triggers the same saving function, so all fields provided to the GUI should be saved. 

> * If the session runs fine, will the final saving process overwrite the backup file with all correct information included (weight info, e.g.)?
Yes, the final saving will overwrite the backup file with all correct information.

> * What if the user decides not to save the session in the end? Would you delete the backup file?
If the user decides not to save the session, I think it's better to save the entire session folder. 

They are all good questions. The backup saving will not interfere the normal saving steps. 

Please see inline one-to-one response

src/foraging_gui/Foraging.py
XX-Yin: Updated naming for consistency.

src/foraging_gui/Foraging.py
XX-Yin: I updated the name to `save_backup_allowed`. 

The backup saving is running in a different thread. When  `save_backup_allowed == 0`, indicating that the previous saving has not finished, and it will not trigger the next saving; `save_backup_allowed == 1`, it is allowed to save backup data.

hanhou:
> > Please also make sure that this backup saving mechanism does not interfere with the normal saving process at the end of the session.
> > Specifically:
> > 
> > * If the session crashes, what part of data will be saved and what part will be missing (at least the post weight)? How about meta data?
> >   If the session crashes, the data not entered to the GUI will not be saved (including the post weight). The meta data will be saved. The backup saving triggers the same saving function, so all fields provided to the GUI should be saved.
> 
> > * If the session runs fine, will the final saving process overwrite the backup file with all correct information included (weight info, e.g.)?
> >   Yes, the final saving will overwrite the backup file with all correct information.
> 
> > * What if the user decides not to save the session in the end? Would you delete the backup file?
> >   If the user decides not to save the session, I think it's better to delete the entire session folder.
> 
> They are all good questions. The backup saving will not interfere the normal saving steps.
> 
> Please see inline one-to-one response

Sounds good. 

> I think it's better to delete the entire session folder.

Does this need to be executed by the code?

Or shall we explicitly separate the normally saved file and the backup file like this:

1. The backup thread always saves the data to `xxx.backup`
2. At the end of the session, if the user saves the data _or_ choose to restart without saving, always delete the backup file.
3. If the session crashes, the backup file remains

src/foraging_gui/Foraging.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         self.ToInitializeVisual = 1 # Should we visualize performance
         self.FigureUpdateTooSlow = 0# if the FigureUpdateTooSlow is true, using different process to update figures
         self.ANewTrial = 1          # permission to start a new trial
-        self.to_save = 1            # permission to save backup data
+        self.save_backup_allowed = 1   # permission to save backup data; 0, the previous saving has not finished, and it will not trigger the next saving; 1, it is allowed to save backup data
```
hanhou, line 97: In this case, I think `previous_backup_completed` sounds even clearer.

XX-Yin: Sounds good! Updated!

XX-Yin:
> > > Please also make sure that this backup saving mechanism does not interfere with the normal saving process at the end of the session.
> > > Specifically:
> > > 
> > > * If the session crashes, what part of data will be saved and what part will be missing (at least the post weight)? How about meta data?
> > >   If the session crashes, the data not entered to the GUI will not be saved (including the post weight). The meta data will be saved. The backup saving triggers the same saving function, so all fields provided to the GUI should be saved.
> > 
> > 
> > > * If the session runs fine, will the final saving process overwrite the backup file with all correct information included (weight info, e.g.)?
> > >   Yes, the final saving will overwrite the backup file with all correct information.
> > 
> > 
> > > * What if the user decides not to save the session in the end? Would you delete the backup file?
> > >   If the user decides not to save the session, I think it's better to delete the entire session folder.
> > 
> > 
> > They are all good questions. The backup saving will not interfere the normal saving steps.
> > Please see inline one-to-one response
> 
> Sounds good.
> 
> > I think it's better to delete the entire session folder.
> 
> Does this need to be executed by the code?
> 
> Or shall we explicitly separate the normally saved file and the backup file like this:
> 
> 1. The backup thread always saves the data to `xxx.backup`
> 2. At the end of the session, if the user saves the data _or_ choose to restart without saving, always delete the backup file.
> 3. If the session crashes, the backup file remains
> 
> This would give them even less opportunity to interfere with each other, and we'll easily know which sessions have crashed.

To simplify steps, I just added a tag `saving_type_label` to differentiate `normal saving` and `backup saving`. If the user clicks the Save button, it will overwrite the existing file with label changed from `backup saving` to `normal saving`.

XX-Yin:
@yonibrowning has agreed to merge this for use in the ephys rig, although he thought we need more discussions in the future. 

Can we approve it so that we can merge it this week?

alexpiet:
Have we tested this with FIP running?

XX-Yin:
> Have we tested this with FIP running?

Not yet, I don't stop the fib photometry for backup saving. It shouldn't influence the fiber photometry, but I will test it. 

image

XX-Yin:
@alexpiet I tested the fiber photometry, it worked.

alexpiet:
@XX-Yin I will review tomorrow, and we can merge before friday

alexpiet (approved):

Issue #474, opened 2024-05-28
Resize GUI windows to fit on half screen


alexpiet:
The water calibration dialog window is too wide to fit in the monitors on 446. It either needs to be resized, or configured to allow scrolling when its smaller (I dont know how @hanhou did this)

this is a list of all the window uis. 
- [x] Water calibration (@alexpiet done)
- [x] Metadata (@alexpiet done)
- [ ] Laser calibration (@XX-Yin )
- [ ] Optogenetics (@XX-Yin )
- [x] Licks distribution
- [x] AutoTrain
- [x] Camera
- [x] TimeDistribution

hanhou:
I was using the [Scroll Area](https://www.pythonguis.com/tutorials/qscrollarea/) of Qt.

I also used the [Layout](https://www.youtube.com/watch?v=h_DVfsD9PKI) to make the GUI resizable.

alexpiet:
@XX-Yin 
- If you plan on doing optogenetics in 446, please resize the laser calibration and optogenetics panels to fit on a half screen. 
- I think its a good idea even if you aren't going to do optogenetics in 447.

XX-Yin:
> @XX-Yin
> 
> * If you plan on doing optogenetics in 446, please resize the laser calibration and optogenetics panels to fit on a half screen.
> * I think its a good idea even if you aren't going to do optogenetics in 447.

@alexpiet 

This seems to be a low priority. I can make changes when we need to monitor two windows. 1) Optogenetics is not monitored all the time, it is usually only open when the GUI starts, and we do not need to monitor two windows at the same time. 2) There is only one opto setup each computer.

alexpiet:
Soon the 447 computers will be half screen sized, and the opto dialog and laser calibration dialog will be unusable in 447. @XX-Yin you should take care of this soon

ellahiltonvano:
@XX-Yin is this something that you can implement? Or do you never expect RAs to perform opto experiments in the behavior rooms?

XX-Yin:
@ellahiltonvano I don't expect the RAs to run opto experiments at this stage.

ellahiltonvano:
@alexpiet given this info does it make sense to close this issue or should we keep it open?

alexpiet:
Its fine with me to close it. Lets just be aware that it will be very difficult to do either laser calibration or optogenetics in 446 or 447

ellahiltonvano:
[...]

PR #502, opened 2024-06-06
Water Calibration overhaul

### LINKED ISSUE #297: Water calibration updates
- Use small step size below a certain value
- Then go to a bigger step size
- .03 is smallest reward size currently
- Is water calibration data being saved? (yes)
- Determine what values we should use for calibration
- Have single value calibration results saved separately -> If result deviates by some amount, then it triggers calibration

Updates on 05/28:
- Right now we use an interval time of 0.5s, which means between each water release we wait 500ms. Some interval here is necessary to ensure the valve closes. I spot checked at an open valve time of 0.060s, using an interval time from .1s to .5s resulted in the same measurement after 200 cycles. We should think a little more carefully about this, but I bet we can use 0.1s intervals
- "Enter" has weird behavior - leads to endless water delievry (because auto locks to "Open left"
- saved water calibration parameters are not loaded? 
- proposal - calibration parameters are defined in file, not in the GUI. there is only one calibration type. 
- Open Left, Keep left open, etc should be locked when "Start calibrating left" is activated
- Should add input checks for alphabetical inputs
- Why is there a keep left open, and keep right open? presumably to flush lines, but they shouldn't be in "Single value calibration"
- start calibration should check for empty tube field
- single calibration should have parallel structure to multi value calibration
- proposal 
   - use valve open times of .015, .02, .025, .03, .04, .045
   - use open interval time of .2s
 - fix annoying press enter
- current: 0.01, to 0.05, with stride of 0.005, maybe trim a pint on each side
- make pop up of when done
- RAs want to know when things are done

### LINKED ISSUE #456: could not convert string to float: '6.97a', Dialogs.py, _StartCalibratingLeft, line 586, 6B
Happened 175k times last week. Looks like we need to make sure there is input checking for some of the water calibration dialog fields

Steps: 
- Figure out a solution for ensuring text input for calibrations are numeric. One possibility is to check and then reprompt the user. In the main window GUI we alternatively just check character by character and remove non-accepted characters.

alexpiet:
  
### Describe changes:
- Resizes the water calibration dialog to fit on a half screen
- Allows spot checks to be saved without interfering with full calibrations
- Changes default calibration settings to a narrower range of values
- Increases the calibration speed by reducing the interval time
- Streamlines logic for calibration, by removing fields and removing notion of saving calibration parameters
- Makes it easier to repeat a measurement
- Emergency stop closes the valve and cancels the current measurement

- partially addresses #474 
- resolves #456, by only using QInputDialog for inputs
- resolves #297 
- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/347
- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/337

- calibration should be significantly faster and easier
- Users should never need to manually delete data

When you open the water calibration dialog, it will look like this:
MicrosoftTeams-imagef0a6eee9a98f5fc5be41f31683ebefb2c48f9149acb48497197222caedb2187e

After pressing Calibrate Left, you will be prompted to enter the before weight:
MicrosoftTeams-imagec32d4f56aafb9bcfb84a8746652a3560f33bda23123ae99b0bc2747547e6ba64

The calibration will run and provide progress reports in purple text. When its finished, you will be prompted to enter the after weight. 
MicrosoftTeams-image6e2e3fb1a8061e81f3d0cff9e63396aa69be31f67eebfae8a7975da8b02d7d6a

You can then press "Continue" to perform the next measurement. "Repeat" to repeat the measurement of the valve open time currently selected. Or "Finish" to stop measurements. You can perform repeats immediately, or do all measurements and then repeat any measurements you like. 
MicrosoftTeams-image500609bbec3c8ab9d571f6c80238fb0fb3214c024a596e541c2667477f064bc5

- You may delete the `WaterCalibrationPar_#.json` file in `ForagingSettings` on each rig. This is not required
- [ ] 446/447
- [ ] 428
- [ ] Ephys1
- [ ] Ephys3
- [ ] To decide: how often to do full calibration, and spot checks

- [x] tested in 447


alexpiet:
calibration_test

In order to test the new calibration system. I calibrated 1A several times, the results are plotted here. The labels are confusing:
- 2024-06-06_left valve, using the new system, with 0.01s intervals
- 2024-06-07_left valve, using the new system, with 0.10s intervals
- 2024-06-08_left valve, using the new system, with 0.25s intervals
- 2024-06-09_left valve, using the new system, with 0.5s intervals
- 2024-06-10_left valve, using the OLD system, with 0.5s intervals
All other labels reflect previous calibrations

We observe a difference in slope between the calibrations (all done today, in a row), and previous calibrations. But all of today's calibrations on old and new system are in agreement. I conclude the new system works, and 0.1s intervals are fine

alexpiet:
To assess what threshold we should use for valve drift, I performed 8 spot checks in a row targetting 2uL volumes. I got back: 2.1 +/- 0.0299 mg (mean +/- std). None of the samples would have been flagged by my proposed tolerance of 10% error ( < 1.8, or > 2.2 mg).

If we had used the old calibration settings to target the 2uL volumes, then the results would have been flagged. 

I conclude the spot checks are valuable, and the tolerance I used is ok. 

I attach to this comment the results of the water calibration test

[calibration_notes.txt](https://github.com/user-attachments/files/15687295/calibration_notes.txt)

[Test_1_WaterCalibration_1.json](https://github.com/user-attachments/files/15687296/Test_1_WaterCalibration_1.json)

alexpiet:
@XX-Yin @hanhou This is ready for review. See two validation test results in the comments

XX-Yin:
@alexpiet Overall it looks good. I don't have time to review it this week. I will review it next week so we can merge it next week.

XX-Yin:
@alexpiet It's better the GUI can be open without the `WaterCalibrationPar_#.json`. We can set some default parameters.

alexpiet:
> @alexpiet It's better the GUI can be open without the `WaterCalibrationPar_#.json`. We can set some default parameters.

Good call. I updated the code to:
- set default fields for `WaterCalibrationPar_#.json`
- No longer log an error if `WaterCalibrationPar_#.json` is not provided, or cannot load
- Only add additional calibration types from `WaterCalibrationPar_#.json` if they have the required fields.

XX-Yin (approved):
I checked and tested the logic. It looks reasonable to me.

PR #503, opened 2024-06-06
390 input checking for settings files

rachelstephlee:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:

Error checking for Setting_box csvs: 
1. Check that there is a newline at the end of the csv file
2. Check that mandatory fields are in csv file. 

### What issues or discussions does this update address?

[Issue 390.](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/390#issuecomment-2153534019) 

### Describe the expected change in behavior from the perspective of the experimenter

Better error checking so Bonsai doesn't crash immediately when the settings_box.csv file is not formatted correctly. 

### Describe any manual update steps for task computers

None. 

### Was this update tested in 446/447?

No.

PR #509, opened 2024-06-12
Update sound

XX-Yin:
  
### Describe changes:
Updated the bonsai workflow to load the decibel parameter from the `settings_box#.csv`.
Previous:
image

Now:
image

Set the decibel from the attenuation parameter, which is more flexible and no need to change the sound waveform.

No

set `AttenuationLeft` to `0` in the `settings_box#.csv`. 
set `AttenuationRight` to `0` in the `settings_box#.csv`. 

- [x] 447 @XX-Yin 
- [ ] 323_EPHYS1 @KanghoonJ 
- [ ] 323_EPHYS3 @XX-Yin 
- [x] 447


alexpiet:
Is `AttenuationLeft` required? Or does the workflow have a default value? 

If it doesn't have a default value, can you add it to the mandatory fields? (https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/503)

Is the process for setting this parameter clearly documented somewhere? Does a positive number mean more attenuation?

hagikent:
@alexpiet 
What @XX-Yin writes are all correct, thanks @XX-Yin . (although the value seemingly not linearly- or log- correlated with decibel..)

>Is the process for setting this parameter clearly documented somewhere?

For the current 447, where already quiet sound is uploaded, it should be always 0.
For 446 (and future 447), where round sound is uploaded and attenuated when being played, setting this value would be basically a calibration process, which we are now working on.

alexpiet:
Thanks @hagikent, @XX-Yin for getting this figured out and implemented!

XX-Yin:
@hagikent @alexpiet During testing in 447, I found some rigs are using the right channel of the soundcard instead of the left channel, so I added both `AttenuationLeft` and `AttenuationRight` to separately control the left and right channels.

hagikent:
sounds good. ideally, connections should be made identical, though!
(yes, I saw connections from soundcard to amplifier was somewhat random too...)

alexpiet:
@XX-Yin @hagikent The next step is to make sure all the rigs are calibrated correctly. Perhaps during that process we can also move all the rigs to use the left sound channel?

hagikent:
@alexpiet  sounds like a plan

PR #517, opened 2024-06-13
Stall timer resets with any Harp message

### LINKED ISSUE #483: Stall timer should be reset when there is early licking
RAs reported a mouse with extremely high early licking that caused the stall timer to trigger. We should reset the timer when this happens. Or maybe we can just ping to see if bonsai is connected before alerting the user?

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/400

alexpiet:
  
### Describe changes:
Since we changed the punishment for early licking, trial durations have sometimes gotten long enough to trigger the stall timer. The stall timer was checking the elapsed time since the last trial start as a way of checking if bonsai had crashed. Now I reset the stall timer based on any harp message from bonsai. Therefore early licking cannot trigger the stall timer, since the licks will show up as harp messages. 

- resolves #483 

- Should see the stall timer only if bonsai crashes

- none

- [x] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3752,8 +3752,8 @@ class Window(QMainWindow):
                     # User continues, wait another stall_duration and prompt again
                     logging.error('trial stalled {} minutes, user continued trials'.format(elapsed_time))
                     stall_iteration +=1
-                else:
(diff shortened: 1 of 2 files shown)


alexpiet:

XX-Yin (approved):

PR #528, opened 2024-06-18
Automatically set IACUC protocol and Project Name from schedule

alexpiet:
  
### Describe changes:
- Uses a cached copy of the training schedule to set the IACUC protocol and Project Name when starting a new mouse
- The cached copy is refreshed every 2 minutes (process running on my computer), and saved at `/allen/aind/scratch/dynamic_foraging/DynamicForagingSchedule.csv`
- If the mouse is not on the schedule, no IACUC protocol is set, and the Project Name defaults to "Behavior Platform"
- If the project description file is not present, then the Project Name defaults to "Behavior Platform"
- Since the ephys rigs and 428 are not on the schedule, they will not have these fields set. But if the mouse went through training in 446/447, then it will already be set by the time the mouse goes to ephys/428

- none

- none

- [x] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -959,7 +959,7 @@ class Window(QMainWindow):
     
     def _LoadSchedule(self):
         if os.path.exists(self.Settings['schedule_path']):
-            schedule = pd.read_csv(self.Settings['schedule_path']
+            schedule = pd.read_csv(self.Settings['schedule_path'])
             
         else:
             logging.error('Could not find schedule at {}'.format(self.Settings['schedule_path']))
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -960,7 +960,12 @@ class Window(QMainWindow):
     def _LoadSchedule(self):
         if os.path.exists(self.Settings['schedule_path']):
             schedule = pd.read_csv(self.Settings['schedule_path'])
-            
+            box_list = []
+            for tower, box in itertools.product([1,2,3,6,7,8,9],['A','B','C','D']):
+                box_list.append(str(tower)+box)
+
+            self.schedule = schedule.query('Box in @box_list').copy() 
+            print(self.schedule)
         else:
             logging.error('Could not find schedule at {}'.format(self.Settings['schedule_path']))
             return
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2636,6 +2636,7 @@ class Window(QMainWindow):
         self.BaseWeight.setText('')
         self.WeightAfter.setText('')
         self.TargetRatio.setText('0.85')
+        self.keyPressEvent(allow_reset=True) 
 
         # Set IACUC protocol in metadata based on schedule
         protocol = self._GetInfoFromSchedule(mouse_id,'Protocol')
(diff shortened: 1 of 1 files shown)


alexpiet:

src/foraging_gui/Foraging.py
```diff
@@ -985,6 +1008,7 @@ def _GetSettings(self):
             'metadata_dialog_folder':os.path.join(self.SettingFolder,"metadata_dialog")+'\\',
             'rig_metadata_folder':os.path.join(self.SettingFolder,"rig_metadata")+'\\',
             'project_info_file':os.path.join(self.SettingFolder,"Project Name and Funding Source v2.csv"),
+            'schedule_path':os.path.join('Z:\\','dynamic_foraging','DynamicForagingSchedule.csv'),
```
XX-Yin, line 1011: We should also enable setting 'schedule_path' from the ForagingSettings.json. The path may be different in different computers.

src/foraging_gui/Foraging.py
```diff
@@ -955,6 +957,27 @@ def _check_line_terminator(self, file_path):
                 return True
             else:
                 return False
+    
+    def _LoadSchedule(self):
+        if os.path.exists(self.Settings['schedule_path']):
+            schedule = pd.read_csv(self.Settings['schedule_path'])
+            box_list = []
+            for tower, box in itertools.product([1,2,3,6,7,8,9],['A','B','C','D']):
```
XX-Yin, line 965: Why do you constrain tower and box?

src/foraging_gui/Foraging.py
alexpiet: The value can be modified by adding an entry in `ForagingSettings.json`. This code just sets the default value. This value is only used as `self.Settings['schedule_path']`

src/foraging_gui/Foraging.py
alexpiet: Because there are rows in the schedule that are blank, have non-mouse information like dividers between 446 and 447 on the schedule. This just filters for rows that contain actual schedule information

alexpiet: It also filters out boxes 4,5,6 which are for non-dynamic foraging tasks

src/foraging_gui/Foraging.py
XX-Yin: That makes sense.

src/foraging_gui/Foraging.py
XX-Yin: Can we just get all rows with a mouse ID? We will not load mouse IDs for non-dynamic foraging tasks. 

 Do you think we can load all the previous schedule information from somewhere? in this way, we can also use it in other rigs (e.g ephys rig).

alexpiet: The problem is that `Mouse ID` also contains junk values from rows that are used for headers/dividers. Including entries like: "behcoverage6.17.png", "6/17/2024", "nan", "446", "447"

We can add boxes 4,5,6 if you want, but since those boxes never run the task it wont make any difference. 

Keeping a database of mice information is a nice idea, but lets make that a separate PR

XX-Yin: The first column contains the mouse IDs and other information, and we can easily extract rows with mouse IDs. Filtering on the mouse ID makes the code more generic, regardless of the device.

image

XX-Yin: > Keeping a database of mice information is a nice idea, but lets make that a separate PR

If it's very complex, it may not worth doing it recently.

XX-Yin: > The problem is that `Mouse ID` also contains junk values from rows that are used for headers/dividers. Including entries like: "behcoverage6.17.png", "6/17/2024", "nan", "446", "447"
> 
> We can add boxes 4,5,6 if you want, but since those boxes never run the task it wont make any difference.
> 
> Keeping a database of mice information is a nice idea, but lets make that a separate PR

There is no need to even fetch the rows with the mouse ID, we can just read all the rows and match the rows with the query mouse ID.

alexpiet: @XX-Yin I've changed the section so now I just drop rows where either "Mouse ID" or "Box" are NaN.

XX-Yin (approved):

PR #535, opened 2024-06-20
Automatically bump foraging_gui version

### LINKED ISSUE #218: Enable version autobump on merges to main via github action
We only merge updates to main once a week. We should configure this update to automatically bump the version of the GUI code, and log this version in the session.json, providing semantic labels for when data was collected.

alexpiet:
  
### Describe changes:
- Increments the version of the foraging gui package with every merge to main. This means each week's release will have a unique tag. Right now we have `0.0.0`, this will increment to `0.0.1`, then `0.0.2` and so on. I propose we manually bump the minor version (eg `0.1.0`) when we have updates that requirement manual intervention
- Then in python `foraging_gui.__version__` will reflect this version number. This will improve clarity of metadata and issue tracking
- Logs this version in the GUI log, and in session metadata

- resolves #218 

- none

- none


alexpiet:

hanhou:
Should we start to add release type to the commit message?

https://github.com/AllenNeuralDynamics/aind-library-template?tab=readme-ov-file#semantic-release

image

alexpiet:
@hanhou I'm not sure how to integrate that into the automatic PRs for the weekly merges to main

hanhou:
You are calling `tag.yml` in `aind-github-actions`, which in turn uses `mathieudutour/github-tag-action@v6.1` ([this line](https://github.com/AllenNeuralDynamics/aind-github-actions/blob/70e7d14addc584e1d661b5eda8489963daa2e3a9/.github/workflows/tag.yml#L41)), right? 

I guess how it works is that it will analyze the `type` prefixes of all commit messages inside a PR and determine the highest level of change:

- If there are any commits with BREAKING CHANGE, it triggers a major version bump. (0.0.0 -> 1.0.0)
- If there are any feat commits but no BREAKING CHANGE commits, it triggers a minor version bump. (0.0.0 -> 0.1.0)
- If there are only fix commits and no feat or BREAKING CHANGE commits, it triggers a patch version bump. (0.0.0 -> 0.0.1)
- If we don't use any `type` prefix before, it will always trigger a patch version bump by default. (0.0.0 -> 0.0.1)

So there are two options for us:

Option 1 (what you described): we don't use `type` prefix in the commit message as before. In this case, weekly PRs always bump the patch version number by default. When we 

Option 2: We adopt the [Angular commit message conventions](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format). In this case, if we want to indicate the requirement of manual intervention, we can add "BREAKING CHANGE:" to the footer of the commit message to bump the major version number.

I think Option 2 may be a better practice in the long run.

alexpiet:
@hanhou Thanks for explaining that! Option 2 seems to be great.

alexpiet:
Do we need to make any changes to this PR? I think we just need to enforce standards on commit messages

hanhou (approved):

Issue #542, opened 2024-06-23
Set the randomness distribution separately for block and ITI/Delay


XX-Yin:
It's better to draw the block length from the uniform distribution and the ITI/delay from the exponential distribution. The current code sets the same randomness distribution for block length and ITI/delay.

alexpiet:
Why set the block length from the uniform distribution?

XX-Yin:
For uncoupled task: Yes
For coupled task: It's truncated exponential distribution.

XX-Yin:
To be more specific. 

For uncoupled task: 
Block: uniform
ITI/Delay: truncated exponential

For coupled task:
Block: truncated exponential
ITI/Delay: truncated exponential

ellahiltonvano:
@XX-Yin is this an update you still think should be implemented?

XX-Yin:
We has a PR that will probably not be merged soon. Please close this issue at this moment.

PR #543, opened 2024-06-23
Update random method

XX-Yin:
  
### Describe changes:
1. Enable to set the randomness methods (even and exponential distribution) separately for block length and ITI/Delay. 
2. Updated the NWB file to include the randomness fields.
3. The uncoupled task hard-coded the randomness methods for block length as `even`. The update allows us to control the randomness of the uncoupled tasks from the GUI.

https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/542

1. The curriculum may need updating in the future (@hanhou). The `Randomness` was moved and updated with two fields called `RandomnessBlock` and `RandomnessOther`. If not provided, `RandomnessBlock` and `RandomnessOther` will default to `Even` and `Exponential` respectively.

No

- [x] tested on my own computer

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -238,7 +238,8 @@ class Window(QMainWindow):
         self.TargetRatio.textChanged.connect(self._UpdateSuggestedWater)
         self.WeightAfter.textChanged.connect(self._PostWeightChange)
         self.BaseWeight.textChanged.connect(self._UpdateSuggestedWater)
-        self.Randomness.currentIndexChanged.connect(self._Randomness)
+        self.RandomnessBlock.currentIndexChanged.connect(self._RandomnessBlock)
+        self.RandomnessOther.currentIndexChanged.connect(self._RandomnessOther)
         self.actionTemporary_Logging.triggered.connect(self._startTemporaryLogging)
         self.actionFormal_logging.triggered.connect(self._startFormalLogging)
         self.actionOpen_logging_folder.triggered.connect(self._OpenLoggingFolder)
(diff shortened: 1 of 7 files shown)


XX-Yin:
@KanghoonJ @ZhixiaoSu 

The `block distribution` of the `uncoupled task` was set to `Even` by default (hardcoded by @hanhou). I updated the code to explicitly set the `block distribution` and `ITI/delay distribution` independently for all tasks.

XX-Yin:
@hanhou @alexpiet @ZhixiaoSu @KanghoonJ This is ready for review.

XX-Yin:
> 2\. hazard

Hi Han, the idea is to allow us to set the distribution of all tasks from the GUI. 

1. I agree to use the uniform distribution for block length in uncoupled task. 
2.  For coupled tasks, it may be better to use a uniform distribution, too. 1) It will let us get more longer blocks during analysis. 2) If it's left-skewed exponential distribution, animals may perceive the block more as `short blocks` and develop strategies good for short blocks.

hanhou:
If your main concern is that blocks are too short, you can increase beta to make them more even. I think this is a major change to the task, and we should discuss it more thoroughly before changing it. Could you add a discussion point to the next behavior meeting?

Also, changing the name will break lots of downstream pipeline, especially considering that during the production_test phase there are different rigs running on different codebases. Could you still use `Randomness` for blocks (default = `exponential`) and `RandomnessOther` for ITI etc (default = `exponential` too)?

XX-Yin:
@hanhou 
`downstream pipeline`, are you mainly meaning the `NWB` and `Curriculum`? 

For `NWB`, we don't have the `Randomness` field stored in the old version. I updated the NWB to add the `Randomness` to different versions of behavior json. 

For `Curriculum`, if the `Randomness` is not detected as a GUI field, will an error occur? Or just ignored.

hanhou:
> For Curriculum, if the Randomness is not detected as a GUI field, will an error occur? 

Looks like we're fine in this case https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/2fea58cc170beb75d6d92e31a9d7253e26c6bf57/src/foraging_gui/Dialogs.py#L2989

But please make sure that both are set to `exponential` by default for maximum backward compatibility.

XX-Yin:
> > For Curriculum, if the Randomness is not detected as a GUI field, will an error occur?
> 
> Looks like we're fine in this case
> 
> https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/2fea58cc170beb75d6d92e31a9d7253e26c6bf57/src/foraging_gui/Dialogs.py#L2989
> 
> But please make sure that both are set to `exponential` by default for maximum backward compatibility.

@XX-Yin 
I think the distribution of block length in the uncoupled task was hardcoded to `uniform`. 

I will set the `block distribution` for `uncoupled task` to `uniform`, and for `coupled task` to `exponential` by default. 
And set the `ITI and delay distribution` to `exponential` by default.

XX-Yin:
@hanhou We can hold off this pull request until the curriculum is updated. Setting default block lengths for different tasks would complicate the logic and be easy to make mistakes.

hanhou:
> I think the distribution of block length in the uncoupled task was hardcoded to uniform.

Oh, yes. But I still think we'll never use exponential block length in uncoupled task since it will further complicate the distribution of the "effective block length" in the uncoupled task. @ZhixiaoSu what do you think?

alexpiet:
Closing as stale

PR #550, opened 2024-06-24
Handles dropped bonsai connection when stopping logging

### LINKED ISSUE #496: ConnectionAbortedError' object is not subscriptable
ConnectionAbortedError' object is not subscriptable | Foraging.py | _stop_logging | line 3165 | 3B, ephys1

Note, that there are two errors here: The bonsai connection is being aborted, which could be from multiple issues. But then there is the secondary issue that the error is not subscriptable, meaning we aren't handling it correctly.

alexpiet:
  
### Describe changes:
Right now, if the user tries to stop logging and bonsai has closed (from either a crash, or user closing the workflow), we generate a new uncaught exception. Now we handle the closed connection and add a warning to the user

- resolves #496 

- Should be more stable when bonsai is closed

- none

- tested on my laptop

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3307,6 +3307,11 @@ class Window(QMainWindow):
         self.WarningLabel_2.setText('')
         self._set_metadata_enabled(True)
 
+        self._ConnectBonsai()
+        if self.InitializeBonsaiSuccessfully == 0:
+            self.WarningLabel.setText('Lost bonsai connection')
+            self.WarningLabel.setStyleSheet(self.default_warning_color)
+
         # Reset state variables
         self._StopPhotometry() # Make sure photoexcitation is stopped 
         self.StartANewSession=1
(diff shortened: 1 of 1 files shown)


alexpiet:
@XX-Yin @hanhou this is ready for review

src/foraging_gui/Foraging.py
```diff
@@ -3303,6 +3307,11 @@ def _NewSession(self):
         self.WarningLabel_2.setText('')
         self._set_metadata_enabled(True)
 
+        self._ConnectBonsai()
```
XX-Yin, line 3310: If we lose connection to bonsai, the error should also happen at the stop the camera (self._stop_camera()).  Can we move this before the self._stop_camera()?

alexpiet: The error is always being generated from the restart_logging function. So I suspect its due to the user manually closing bonsai, and not a crash. 

It looks like the stop_camera function has some code to check for bonsai connections already

XX-Yin (approved):

PR #563, opened 2024-06-25
[skip email] Han add test trunexp

hanhou:
### Describe changes:
- Add a script for testing distribution and hazard rate of truncated exponential distribution

- related to https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/442

NA

NA

NA


XX-Yin:
Should we add another folder called `analysis` and move this file to that folder?

hanhou:
I don't think it belongs to analysis. I think it should be a part of the task design, especially for block design. So I put it in the same folder as the block generators for convenience.

alexpiet (approved):

XX-Yin (approved):

PR #589, opened 2024-07-03
[skip email] matching tag in this branch with main

alexpiet:
For some reason back-merging didn't work. So I'm manually getting this version to 0.1.2, to match main. This should synchronize all the branches


alexpiet:

PR #590, opened 2024-07-03
Updates water calibration tolerance, and clarifies procedure for spot checks

### LINKED ISSUE #561: Water calibration spot checks - increase tolerance, and clarifying warning
I watched @NgD-Huy and @henryloeffler do a spot check today. About 4-5 of the rigs exceeded the 10% tolerance, but just barely. I think we should increase the tolerance so these warnings are less common. 

Then, when we do exceed tolerance, we should alert the user to repeat the spot check, and if its still above threshold, immediately perform a full calibration

alexpiet:
  
### Describe changes:
- Increases the tolerance for the spot checks from 10% to 15%
- Clarifies the text when spot checks are outside of the tolerance that the user should (1) double check the entered values, (2) repeat the spot check, (3) do a full calibration if the second spot check is still outside the bounds. 

- resolves #561 

- More clear instructions

- none

- tested on my laptop

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -1054,7 +1054,7 @@ class WaterCalibrationDialog(QDialog):
         TOLERANCE = float(self.SpotLeftVolume.text())*.15
         if np.abs(error) > TOLERANCE:
             reply = QMessageBox.critical(self, 'Spot check left', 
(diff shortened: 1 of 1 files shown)


alexpiet:

XX-Yin (approved):

PR #596, opened 2024-07-09
Add one trial for no response

XX-Yin:
  
### Describe changes:
1. Added a selection `add one trial for no response` to the UI. It is set to `Yes` by default. 
image
2. When the `add one trial for no response` is selected as `Yes`, it will add one trial to the block length on both sides. 

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/439

Select the `add one trial for no response` as `No` if you don't want to turn off this function.

No

Tested on my own computer.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3776,7 +3776,7 @@ class Window(QMainWindow):
                 print('Current trial: '+str(GeneratedTrials.B_CurrentTrialN+1))
                 logging.info('Current trial: '+str(GeneratedTrials.B_CurrentTrialN+1))
                 if (self.GeneratedTrials.TP_AutoReward  or int(self.GeneratedTrials.TP_BlockMinReward)>0
-                    or self.GeneratedTrials.TP_Task in ['Uncoupled Baiting','Uncoupled Without Baiting']):
+                    or self.GeneratedTrials.TP_Task in ['Uncoupled Baiting','Uncoupled Without Baiting']) or self.AddOneTrialForNoresponse.currentText()=='Yes':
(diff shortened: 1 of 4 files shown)


ZhixiaoSu:
Thank you for adding this feature! The related issue is here https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/439
So, with this change: 
Does it mean this choice is only manual? If we do that, it will put the animal off-curriculum. Or we will need to add a field to all auto-training curriculums And what's the reason of keeping it 'No' as default?

XX-Yin:
> Thank you for adding this feature! The related issue is here [AllenNeuralDynamics/aind-behavior-blog#439](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/439) So, with this change: Does it mean this choice is only manual? If we do that, it will put the animal off-curriculum. Or we will need to add a field to all auto-training curriculums And what's the reason of keeping it 'No' as default?

Ok, I can set it `Yes` by default. 

When I am doing optogenetics, some effects are "increased no response rate" and I don't want an additional trial. I can manually change this for my mice.

ZhixiaoSu:
Got it! This sounds more like something to be taken care of in the data acquisition period after being well-trained. If this is the case, maybe it makes sense to have an optogenetics curriculum so that the graduation+opto stage has this field to be 'No'?

XX-Yin:
@ZhixiaoSu Updated. The default was set to `Yes`

KanghoonJ (approved):

ZhixiaoSu:
7/9/2024: Branch tested on 446 6A. Session runs normally, and block length gets longer by 1 trial on both sides when animal does not respond. Tested task: uncoupled unbaited.

PR #600, opened 2024-07-13
confirm the experimenter name

XX-Yin:
  
### Describe changes:
Make it mandatory to check the experimenter's name. A window will pop up when the `Start` is clicked. 

image

- If you select "Yes", the session will start. If you select "No", the session will not start.

I frequently started sessions with the wrong default experimenter name, especially I trained the same mice with other people in different sessions. I think the same may happen to others. Having the experimenter check the experimenter's name will fix this problem.

A window will pop up asking for checking experimenter's name. 

No

Tested on my own computer

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3507,7 +3507,7 @@ class Window(QMainWindow):
             # check experimenter name
             reply = QMessageBox.critical(self,
                 'Box {}, Start'.format(self.box_letter),    
-                f'The experimenter is <span style="color:red;">{self.Experimenter.text()}</span>. Is it correct?',
+                f'The experimenter is <span style="color:red;">{self.Experimenter.text()}</span>. Is this correct?',
                 QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
             if reply == QMessageBox.No:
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -3503,17 +3503,16 @@ def _Start(self):
...
-                    'Box {}, Start'.format(self.box_letter),    
-                    'Experimenter field set to default, continue anyways?',
-                    QMessageBox.Yes | QMessageBox.No, QMessageBox.No)
-                if reply == QMessageBox.No:
-                    self.Start.setChecked(False)
-                    logging.info('User declines using default name')
-                    return                
+            reply = QMessageBox.critical(self,
+                'Box {}, Start'.format(self.box_letter),    
+                f'The experimenter is <span style="color:red;">{self.Experimenter.text()}</span>. Is it correct?',
```
alexpiet, line 3510: Change the text to "Is this correct?"

alexpiet (changes requested):

src/foraging_gui/Foraging.py
XX-Yin: updated!

alexpiet (approved):

Issue #377, opened 2024-04-10
Setting up different versions of bonsai json files for testing TransferToNWB


hanhou:
I'm uploading some json files here for testing the TransferToNWB script.

[668551_2023-06-16.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936281/668551_2023-06-16.json)
[662914_2023-09-22.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936313/662914_2023-09-22.json)
[684039_2023-12-01_08-22-32.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936315/684039_2023-12-01_08-22-32.json)
[704151_2024-02-27_09-59-17.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936331/704151_2024-02-27_09-59-17.json)
[1_2024-04-06_16-31-06.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936356/1_2024-04-06_16-31-06.json)
[706893_2024-04-09_14-27-56_ephys.json](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/files/14936359/706893_2024-04-09_14-27-56_ephys.json)

hanhou:
`list out of range` error
[746346_2025-01-02_10-16-10.json](https://github.com/user-attachments/files/18303936/746346_2025-01-02_10-16-10.json)

hanhou:
`too many indices` error
[746346_2024-12-02_13-16-12.json](https://github.com/user-attachments/files/18304104/746346_2024-12-02_13-16-12.json)

PR #602, opened 2024-07-15
fixed nwb

XX-Yin:
  
### Describe changes:
1. Fixed the NWB data type issue for `B_DelayStartTimeHarp`. The `B_DelayStartTimeHarp` is `Null` for optogenetics (no delay period for optogenetics condition), which will cause an error. This issue is solved by setting the default to `np.nan` for empty `B_DelayStartTimeHarp`.
2. Handle different names. `Laser_#` was previously updated to `LaserColor_#` for clarity. Updated here to handle both names. 

see Describe changes section. 

No

No

Tested on my own computer with multiple sessions.


src/foraging_gui/TransferToNWB.py
alexpiet, line 1: Is this file being used anymore, once the automatic pipeline starts running? @hanhou

hanhou: My temporary pipeline is using this file, while the automatiic pipeline under development is more complicated. They copy and paste our code and develop separately in `/util/behavior_to_nwb.py` file in [this capsule](https://codeocean.allenneuraldynamics.org/capsule/9216710/tree) of [this pipeline](https://codeocean.allenneuraldynamics.org/capsule/0995759/tree). We should definitely find a better place to put this file. @rachelstephlee  @Ahad-Allen

XX-Yin:
@alexpiet @hanhou I think we can merge this PR first and then migrate it to a different place where the downstream pipeline can easily get access to.

hanhou (approved):

hanhou (commented):
Please also add a new json test file [here](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/377) that this PR was supposed to fix, and include it to the test list at line 577.

XX-Yin:
Added, but that file is not added to the `test_json_urls` in the `TransferToNWB.py`. We can do it in the next update.

Issue #606, opened 2024-07-16
Swap top and bottom position of left and right licks in real time visualizer


alexpiet:
to be consistent with main GUI window

alexpiet:
@hagikent

hagikent:
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/609
should solve this.

PR #609, opened 2024-07-17
Lick visualization LR switch

hagikent:
### Describe changes:
L:top / R:bottom -> L:bottom / R:top

https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/606

You should the lick visualization it flipped

no

In 428.


hagikent:
@alexpiet addressing the request from @NgD-Huy that LR-licks should be flipped to match the bottom video.
As it's a just visualization fix, you don't need to carefully review.

alexpiet (approved):

PR #619, opened 2024-07-24
Engage autotrain by default

hanhou:
### Describe changes:
- Create the AutoTrain dialog in the background upon GUI initialization.
- Whenever the user changes the subject_id and hits Enter: 
    - if the mouse is already in auto-training, the AutoTrain dialog will automatically engage the curriculum and the suggested stage.
    - if the mouse doesn't exist, do nothing 

- Resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/355

- By default, AutoTrain will be engaged automatically if the mouse exists.
image

- When necessary, the user can still open the AutoTrain dialog to interact as before.
- If AWS credential is missing, the user will receive a warning message when the GUI opens, but the GUI will still run without AutoTrain.
image
image

None

- Tested on my PC.


XX-Yin:
@hanhou what will happen if the aws is not connected?

hanhou:
> @hanhou what will happen if the aws is not connected?

Good question!

I just added a commit to better handle the AWS connection issue. See the updated PR description.

XX-Yin:
Can we add an option to indicate whether auto-training should be used automatically when loading a session (it could be set as `enable automatically training` by default)?  This can handle the case where we are using the off-curriculum training, but the mouse is trained before with the autotrain.

hanhou:
> Can we add an option to indicate whether the auto-training should be used automatically when loading a session (it could be set as `enable automatically training` by default)? This can handle the case where we are using the off-curriculum training, but the mouse is trained before with the autotrain.

I stongly suggest that we keep on-curriculum even at ephys recording. If you'd like to use training parameters different than STAGE_GRADUATION, you should actually add a new stage to the curriculum to make it on-curriculum, such as STAGE_EPHYS.

That said, if you still would like to add a switch in Foraging_setting, just add your switch to [this line](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/034d934329f5bebe678a76b6033672c880060a47/src/foraging_gui/Foraging.py#L251).

XX-Yin:
Ok, I added an option to set the `auto_engage` from the settings file. It's set to `True` by default.

PR #628, opened 2024-07-29
Skip upload manifest for test mouse, clarify alert text for upload manifest failures

### LINKED ISSUE #626: Skip upload manifest if mouse ID is 0.
That way we don't need to worry about uploading testing data

### LINKED ISSUE #627: Alert for failed upload manifest should say to make an issue, instead of tell the mouse owner

alexpiet:
  
### Describe changes:
- Does not generate an upload manifest for the test mouse (any ID <= 10)
- Changes the alert text to: "Could not generate upload manifest. Please alert the mouse owner, and report on github."

- resolves #626 
- resolves #627 

- none

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4232,6 +4232,9 @@ class Window(QMainWindow):
         '''
             Generates a manifest.yml file for triggering data copy to VAST and upload to aws
         '''
+        if self.ID.text() == '0':
+            logging.info('Skipping upload manifest, because this is the test mouse')
+            return
         try: 
             if not hasattr(self, 'project_name'):
                 self.project_name = 'Behavior Platform'
@@ -4286,7 +4289,7 @@ class Window(QMainWindow):
             logging.error('Could not generate upload manifest: {}'.format(str(e)))
             QMessageBox.critical(self, 'Upload manifest', 
                 'Could not generate upload manifest. '+\
-                'Please alert the mouse owner that this session will not be uploaded.')
+                'Please alert the mouse owner, and report on github.')
             
 
 def start_gui_log_file(box_number):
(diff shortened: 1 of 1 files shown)


alexpiet:

XX-Yin:
Do we need to store the state of manifest file generation/uploading success? The general question is how to track which file is successfully uploaded or not.

alexpiet:
> Do we need to store the state of manifest file generation/uploading success? The general question is how to track which file is successfully uploaded or not.

@XX-Yin The aind-watchdog-service has its own log, and will generate an alert if a session is not uploaded correctly. But this alert only happens if the manifest is created. So as long as the GUI creates the manifest, all other tracking will be handled by the watchdog service

XX-Yin:
> > Do we need to store the state of manifest file generation/uploading success? The general question is how to track which file is successfully uploaded or not.
> 
> @XX-Yin The aind-watchdog-service has its own log, and will generate an alert if a session is not uploaded correctly. But this alert only happens if the manifest is created. So as long as the GUI creates the manifest, all other tracking will be handled by the watchdog service

That sounds good. Is it necessary to have a folder called `manifest_failure` to store the failure sessions every day? So, we can more easily extract the failure sessions.

alexpiet:
> > > Do we need to store the state of manifest file generation/uploading success? The general question is how to track which file is successfully uploaded or not.
> > 
> > 
> > @XX-Yin The aind-watchdog-service has its own log, and will generate an alert if a session is not uploaded correctly. But this alert only happens if the manifest is created. So as long as the GUI creates the manifest, all other tracking will be handled by the watchdog service
> 
> That sounds good. Is it necessary to have a folder called `manifest_failure` to store the failure sessions every day? So, we can more easily extract the failure sessions.

The watchdog service has a `manifest` folder, and then a `manifest_complete` folder. So any session in `manifest` has not yet been uploaded.

XX-Yin (approved):

Issue #653, opened 2024-08-09
Ensure schedule information is set properly


alexpiet:
- check for FIP on/off, and mode
- check for curriculum
- Good bridge issue before using SLIMS to set all of this information automatically.

alexpiet:
We can address https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/559 at the same time. 
Proposed solution:
- Move the logic that loads the schedule and sets project information from `Foraging._OpenNewMouse` to `Foraging._Start`
- Check if the curriculum matches
- Check if FIP status is correct
- Check if FIP mode is correct
- We dont need to check if its the correct box, trainer, or training time, since those can be flexible. 

When we move to loading this information from SLIMS, we should have the GUI set all of those fields automatically. But for now, lets just verify they are correct

micahwoodard:
@alexpiet for check if FIP on/off should I be checking self.FIP_started? I am unclear about which variable indicates if FIP is on or off

alexpiet:
@micahwoodard Good question. You should check if `self.PhotometryB.currentText()=='on'` This is the GUI toggle button that sets whether to run FIP (FIP = Fiber Photometry, fyi). `FIP_started` tells us whether the FIP process has started already for this session, but we already check that if PhotometryB is "on". Since you are checking the schedule, you dont care if the FIP has started or not, only that the toggle has been set to eventual start.

micahwoodard:
Oh okay I should've thought harder about FIP and photometry. If the schedule is blank in the FIP column, does that mean not photometry?

alexpiet:
Yes, if the schedule is blank, then no photometry

PR #669, opened 2024-08-28
Feat 653 schedule info

micahwoodard:
  
### Describe changes:
Schedule is checked to confirm FIP mode and settings are correct
#653 
Pop up will occur if gui detects discrepancy between schedule and settings 
None
- [x] 446


micahwoodard:
@alexpiet to get this to work locally, I had to make changes by creating the self.schedule attribute and changing the mouse_id input from a string to an int in _GetInfoFromSchedule. I couldn't find where in the code that self.schedule was being assigned so calling _GetInfoFromSchedule always returned None for me. Is that just because I don't have something on my computer or a bug in the code base?  For using an int instead of a str, I was wondering if that is just the schedule I'm using or another bug?

alexpiet (approved):

Issue #651, opened 2024-08-09
GUI does not alert for saving without closing


alexpiet:
- Perhaps auto saving messes with this?
- three times yesterday users closed the GUI without saving final weight

alexpiet:
Details:
- Its unclear to me why the GUI doesn't alert for saving without closing. The first step is trying to replicate this

micahwoodard:
- seperate out post wieght info vs unsaved data
- let ra overwrite

PR #670, opened 2024-08-29
Post weight save checks

micahwoodard:
  
### Describe changes:
More checks to make sure user enters post weight of mouse
#651 
Additional pop up when post weight is not entered and session has been run
None 
- [x] 446


src/foraging_gui/Foraging.py
```diff
@@ -2119,11 +2119,20 @@ def closeEvent(self, event):
         # stop the current session first
         self._StopCurrentSession() 
 
+        reply = None
```
alexpiet, line 2122: Why make a default for this here?

src/foraging_gui/Foraging.py
```diff
@@ -2119,11 +2119,20 @@ def closeEvent(self, event):
...
-                'Box {}, Foraging Close'.format(self.box_letter), 
+            reply = QMessageBox.critical(self,
+                'Box {}, Foraging Close'.format(self.box_letter),
                 'Exit without saving?',
-                QMessageBox.Yes | QMessageBox.No , QMessageBox.No)  
+                QMessageBox.Yes | QMessageBox.No , QMessageBox.No)
+            if reply == QMessageBox.No:
+                event.ignore()
+                return
+        elif self.WeightAfter.text() == '' and self.StartANewSession == 1 and reply != QMessageBox.Yes and not self.unsaved_data:  # post weight not entered and session ran
```
alexpiet, line 2131: How would reply have been set by this point?

src/foraging_gui/Foraging.py
```diff
@@ -3137,14 +3146,20 @@ def _LoadVisualization(self):
 
     def _Clear(self):
         # Stop current session first
-        self._StopCurrentSession()    
-    
+        self._StopCurrentSession()
+
+        reply = None
```
alexpiet, line 3151: Same logic questions here with reply

src/foraging_gui/Foraging.py
```diff
@@ -3425,17 +3440,28 @@ def _NewSession(self):
         logging.info('New Session pressed')
 
         # If we have unsaved data, prompt to save
-        if (self.ToInitializeVisual==0) and (self.unsaved_data): 
-            reply = QMessageBox.critical(self, 
-                'Box {}, New Session:'.format(self.box_letter), 
+        reply = None
```
alexpiet, line 3443: Again here. Maybe I am missing something

alexpiet (changes requested):

src/foraging_gui/Foraging.py
micahwoodard: I was trying to avoid the situation where if user said it was fine that there was unsaved data , the weight pop up wouldn't show since the user already said it was fine. I think that's taken care of though by saying not self.unsaved_data so I can remove the reply != QMessagebox.Yes

alexpiet (approved):

Issue #668, opened 2024-08-28
Making directions/orientations consistent across training-software modalities


hagikent:
@alexpiet @hanhou @XX-Yin @ZhixiaoSu @KanghoonJ @galenlynch @bwtan-allen 
See below for the feature request from RAs (discussed at this week's Tue9 meeting).
I will take care of 1, 2, and 4 this week. 
@alexpiet could you please take care of 3? (or delegate it to @XX-Yin or someone else? @micahwoodard maybe?)

See this RAs' group agreement summarized by @ellahiltonvano 
https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/596#discussioncomment-10478931

1.
![image](https://github.com/user-attachments/assets/36a4cc3f-20d1-4a46-a2f1-fe00a39bba50)
2.
![image](https://github.com/user-attachments/assets/c6ccfcbd-1e1a-4b8d-9afa-d7d6158384fc)
3.
![image](https://github.com/user-attachments/assets/7c6db129-dc0b-473f-ba54-bbe4d586d4aa)
4.
![image](https://github.com/user-attachments/assets/1ad82848-498a-42a2-9efa-78117bb290cb)

---
@galenlynch @KanghoonJ @ZhixiaoSu @XX-Yin 
I will try to make the flip functions conditional/optional so that you can make it off at ephys rigs just by changing a flag in the local setting json file.

hanhou:
Sorry that I missed this discussion, but this is in contrast to many downstream conventions that have been already in place. In all visualizations and simulations, we have been using LEFT=0 and RIGHT=1, which is also in line with papers published by the Cohen lab.

alexpiet:
Downstream plots can use whatever convention we like. This is just for the acquisition computers.

hagikent:
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/674
This PR will solve 1, 2, and 4. Could someone work on the GUI visualization?

hagikent:
Someone still needs to take care of 3.

PR #673, opened 2024-08-30
Add image visualizer that allows online rotation

bruno-f-cruz:
Partially address #668 

The rotation can be applied by opening an IplImage with the new visualizer, right clicking anywhere in the canvas and using the bottom UI.

This code will eventually live with the core bonsai repos in the Allen (https://github.com/AllenNeuralDynamics/Bonsai.AllenNeuralDynamics)

The alternative solution would be something like this: (I would advise against it as it is computationally inefficient since each flip copies the image)

image


hagikent (approved):
I will integrate this to high-speed camera preview/recording and test.

PR #682, opened 2024-09-10
fix logic for post weight pop up

micahwoodard:
  
### Describe changes:
- better logic for determining if a session has been run and if the weight of mouse has still not been entered
[#615](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/615)
- [ ] 446


micahwoodard:
@alexpiet  It seems like after you hit save , the UI is assuming the [next start click will be a new session](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/54d502e89323458a95297803dd3111d37041a666/src/foraging_gui/Foraging.py#L2501C13-L2507C39)? Is that right, and if so, should we check if the post weight has been entered in the case start is pressed after save?

alexpiet:
@micahwoodard This is one of the places where the logic in the code is very messy.

After `_Save()` has run with `SaveContinue==0`, if a user presses `Start` then a new session is started. 

I think we want to check for post weight in the case where (1) trials have been run (this session wasn't loaded from file), and (2) the user presses close/quit, or load (to load a new mouse), or Start (to start a new session).

src/foraging_gui/Foraging.py
```diff
@@ -3103,13 +3105,19 @@ def _Clear(self):
         if self.unsaved_data:
             reply = QMessageBox.critical(self,
                 'Box {}, Clear parameters:'.format(self.box_letter),
-                'Unsaved data exists! Do you want to clear training parameters?',
+                'Unsaved data exists!  n `  training parameters?',
```
alexpiet, line 3108: What is going on here with line 3108?

alexpiet (approved):
I think you accidentally deleted one message on line 3108, but looks good other than that

micahwoodard:
I've made changes for behavior when loading in a mouse.

alexpiet (approved):

PR #684, opened 2024-09-11
update data backup

XX-Yin:
  
### Describe changes:
Save data in the main thread. 

I speculate that the occasional crashes are related to the conflicting interaction between the different threads (between backup saving and `_GenerateATrial`).  Here, I moved the backup to after the `GeneratedTrials._GenerateATrial(self.Channel4)` to avoid simultaneous running of these two functions.

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/605#issuecomment-2342288067

No

No

Tested on my own computer.


alexpiet:
But if saving takes .5s, that means the GUI will be unresponsive for .5s after every trial. This is not ideal

XX-Yin:
> But if saving takes .5s, that means the GUI will be unresponsive for .5s after every trial. This is not ideal

You are right. I updated it to still use a separate thread. And moved the backup to after the `GeneratedTrials._GenerateATrial(self.Channel4) ` as I found the conflict is mainly caused by interaction with `_GenerateATrial`.

XX-Yin:
@alexpiet Do you have extra concerns about this PR?

src/foraging_gui/Foraging.py
```diff
@@ -3963,6 +3959,23 @@ def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
                 #generate a new trial
                 if self.NewTrialRewardOrder==1:
                     GeneratedTrials._GenerateATrial(self.Channel4)   
+                '''
+                ### Save data in the main thread. Keep it temporarily for testing ###
```
alexpiet, line 3963: Did you mean to remove this part, saving data in the main thread?

XX-Yin: Yes, I discarded this the strategy to 'save data in the main thread'. Instead, I moved the previous `saving code` to the after of `GeneratedTrials._GenerateATrial(self.Channel4)` to make sure they are not running simultaneously.

alexpiet: I'm confused. Line 3967 shows saving the data in the main thread, then line 3975 shows saving in a separate thread. Did you intend to remove the saving in the main thread?

XX-Yin: The code for "saving data in the main thread" has been commented out with ''' '''

alexpiet: Oh duh. I didn't see the multiline comment ```, I was looking for a # before each line

alexpiet (approved):

PR #694, opened 2024-09-20
Ensures project name is always set even if its not in the schedule

alexpiet:
  
### Describe changes:
Fixes a small bug where `project_name` is not set in the manifest if its not on the schedule. The default `Behavior Platform` was correctly set in the metadata, but not the upload manifest. 

none

none

- [ ] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3784,6 +3784,8 @@ class Window(QMainWindow):
                     logging.info('Setting Project name: {}'.format(project_name))
                     add_default = False
             if add_default:
+                project_name = 'Behavior Platform'
+                logging.info('Setting Project name: {}'.format('Behavior Platform'))
                 projects = [self.Metadata_dialog.ProjectName.itemText(i)
                             for i in range(self.Metadata_dialog.ProjectName.count())]
                 index = np.where(np.array(projects) == 'Behavior Platform')[0]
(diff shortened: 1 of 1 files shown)


alexpiet:
@XX-Yin @hanhou @micahwoodard This is ready for review

XX-Yin (approved):

src/foraging_gui/Foraging.py
```diff
@@ -3784,14 +3784,15 @@ def _Start(self):
                     logging.info('Setting Project name: {}'.format(project_name))
                     add_default = False
```
micahwoodard, line 3785: This add default logic seems a little convoluted to me but it could be I'm just not familiar with the code.  It seems like you could achieve something similar like this:

```
schedule_info= self._GetInfoFromSchedule(mouse_id, 'Project Name')
project_name = schedule_info if schedule_info is not None else 'Behavior Platform'
projects = [self.Metadata_dialog.ProjectName.itemText(i)
                            for i in range(self.Metadata_dialog.ProjectName.count())]
index = np.where(np.array(projects) == project_name)[0]
if len(index) > 0:
  index = index[0]
  self.Metadata_dialog.ProjectName.setCurrentIndex(index)
   self.Metadata_dialog._show_project_info()
  logging.info('Setting Project name: {}'.format(project_name))
else:
  project_info = {
                        'Funding Institution': ['Allen Institute'],
                        'Grant Number': ['nan'],
                        'Investigators': ['Jeremiah Cohen'],
                        'Fundee': ['nan'],
                    }
  self.Metadata_dialog.project_info = project_info
  self.Metadata_dialog.ProjectName.addItems([project_name])
  logging.info('Setting Project name: {}'.format(project_name))
```
Maybe not worth the effort though to simplify this logic if it's just for a small bug fix.

micahwoodard (commented):
Suggestion of simplified logic if I understand the functionality correctly. But as a quick bug fix looks good to me.

PR #695, opened 2024-09-20
Feat 595 slims waterlog

micahwoodard:
  
### Describe changes:
- gui will write waterlogs to slims when user enters post weight and clicks save
- [#595](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/595)
- None
- environment variables need to have SLIMS_PASSWORD and SLIMS_USERNAME (handeled by SIPE). 

- [ ] 446


micahwoodard:
There is some messiness associated with adding a mouse that is not found in slims. I was assuming that adding mice using the api will be temporary, which is why I thought it would be okay to reference the schedule and have pop up windows within the AddWaterlog function that will be removed later. Let me know if this is not the case, and I can rework the logic.

src/foraging_gui/Foraging.py
```diff
@@ -1216,6 +1223,105 @@ def _GetSettings(self):
...
+        '''
+        try:
+            self.slims_client = SlimsClient(username=os.environ['SLIMS_USERNAME'],
+                                            password=os.environ['SLIMS_PASSWORD'])
+        except KeyError:
+            raise KeyError('SLIMS_USERNAME and SLIMS_PASSWORD do not exist as '
+                         'environment variables on machine. Please add')
+
+        try:
+            self.slims_client.fetch_model(models.SlimsMouseContent, barcode='00000000')
```
alexpiet, line 1238: It looks like the logic here is you try to check SLIMS for mouse "0000000". If you get an exception saying "no record found" then you assume that means you connected to slims. If you get some other error, then you assume its a connection issue. 

Does the SLIMS API have a better way to check the connection? 

If fetch fails for other reasons (like the SLIMS server is down). Then the error message is a little confusing. I would say something like. "Exception trying to read from Slims: {e}.\n This may be due to credential errors. ..."

src/foraging_gui/Foraging.py
```diff
@@ -1216,6 +1223,105 @@ def _GetSettings(self):
...
+        '''
+
+        try:    # try and find mouse
+            mouse = self.slims_client.fetch_model(models.SlimsMouseContent, barcode=session.subject_id)
+        except Exception as e:
+            if str(e) == 'No record found.':    # if no mouse found or validation errors on mouse
+                # check schedule to make sure enough information is known about mouse to create model in slims
+                if not pd.isnull((curr_st := self._GetInfoFromSchedule(session.subject_id, 'Current State'))) \
+                        and not pd.isnull((point_of_contact := self._GetInfoFromSchedule(session.subject_id, 'PI'))):
+                    model = models.SlimsMouseContent(barcode=session.subject_id,
```
alexpiet, line 1263: Add a logging error here
logging.error('Mouse not in SLIMS, creating new entry....')

src/foraging_gui/Foraging.py
```diff
@@ -1216,6 +1223,105 @@ def _GetSettings(self):
...
+                    logging.error(f'Mouse {session.subject_id} does not exist in Slims, and schedule does not contain '
+                                  f'enough information to add mouse. Waterlog needs to be added manually to Slims')
+                    QMessageBox.critical(self,
+                                         'Adding Waterlog',
+                                         f'Mouse {session.subject_id} does not exist in Slims, and schedule does not contain '
+                                         f'enough information to add mouse. Waterlog needs to be added manually to Slims',
+                                         QMessageBox.Ok)
+                    return
+            else:
+                raise e
```
alexpiet, line 1293: I would add logging.info statements to all the steps of this logic tree. Its super messy, so lets leave a paper trail. I understand that adding the mouse is temporary

src/foraging_gui/Foraging.py
```diff
@@ -1216,6 +1223,105 @@ def _GetSettings(self):
...
+            test_pk=self.slims_client.fetch_pk("Test", test_name="test_waterlog"))
+
+        # check if mouse already has waterlog for at session time and if, so update model
+        waterlogs = self.slims_client.fetch_models(models.SlimsWaterlogResult, mouse_pk=mouse.pk)
+        if waterlogs != [] and waterlogs[0].date.strftime("%Y-%m-%d %H:%M:%S") == \
+                session.session_start_time.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S"):
+            model.pk = waterlogs[0].pk
+            self.slims_client.update_model(model=model)
+        else:
+            self.slims_client.add_model(model)
```
alexpiet, line 1324: just add more logging.info statements through here to help with future debugging

src/foraging_gui/Foraging.py
micahwoodard: Yeah that makes sense I'll ask Patrick if there is a better way to tell if the credentials are incorrect

src/foraging_gui/Foraging.py
```diff
@@ -241,6 +247,7 @@ def connectSignalsSlots(self):
         self.actionConnectBonsai.triggered.connect(self._ConnectBonsai)
         self.actionReconnect_bonsai.triggered.connect(self._ReconnectBonsai)
         self.Load.clicked.connect(self._OpenLast)
+        self.Save.setCheckable(True)
```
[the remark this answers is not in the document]
micahwoodard: It allows the Save button to be set to checked when pressed by a user. I needed the save button to be checkable so I could tell if the _Save function was called by a user pressing the save button (Save.isChecked() == True) or if it was called by somewhere else in the code. I think some of the other buttons like Start are set checkable somewhere in the ui files since Start.isChecked() is used throughout the code, but I'm assuming the Save button is not because I couldn't get it to stay checked without adding this line.

alexpiet (approved):

alexpiet:
@micahwoodard If a computer doesn't have the SLIMS credentials, then the GUI will raise an error. I therefore think we shouldn't merge this in until we know the environment credentials have been set. Does that make sense?

micahwoodard:
Yeah that makes perfect sense to me. I'll hold off on merging this until Jessy finishes adding environment variables

alexpiet:
Merging because SLIMS credentials are set up

PR #701, opened 2024-09-24
Feat 668 switch graph axis

micahwoodard:
  
### Describe changes:
- switch position of L and R on graph
- #668 
- position of L and R are flipped on graph
- None
- [ ] 446

### DIFF
--- a/src/foraging_gui/Visualization.py
+++ b/src/foraging_gui/Visualization.py
@@ -354,11 +354,15 @@ class PlotV(FigureCanvas):
 
     def _UpdateAxis(self):
         self.ax1.set_yticks([0,1])
-        self.ax1.set_yticklabels(['R', 'L'])
+        self.ax1.set_yticklabels(['L', 'R'])
         self.ax1.set_ylim(-0.6, 1.6)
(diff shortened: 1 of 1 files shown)


micahwoodard:
image
Here's what it looks like now which is flipped from current implementation on main

alexpiet:
This switches the labels, but not the plot itself

micahwoodard:
Ahh okay my bad I thought it was just switching axis labels. I'll work more

hagikent:
@micahwoodard 
Yes, the content to be visualized has to be flipped, not only labels.
Note, the data to be saved for analysis should be kept as is! Down-stream analysis convention shouldn't be changed.

alexpiet:
@micahwoodard You might be able to do something as simple as `plt.gca().invert_yaxis()`

micahwoodard:
@alexpiet  and @hagikent I think it should be fixed now. Per Alex's suggestion, I just flipped the y axis orientation instead. Here is what it looks like now. 
image

alexpiet (approved):

Issue #296, opened 2024-03-12
Add indicator for |bias| > 0.7 


alexpiet:
Steps:
- Calculate bias online (@hanhou has a definition)
- Check for bias outside of a tolerance range at the start of each trial
- Alert the user through either a pop-up, mark on the trial graph, or a text warning (seek RA feedback here)
    - remove the matching plot
    - Moving the legends to the space where the matching plot is now
    - design some indicator for side bias
    - Can then remove the "hide legend" button

alexpiet:
@hanhou Can you provide @micahwoodard a definition of how to compute bias online?

alexpiet:
These two issues might be redundant:
I think this issue is one method for computing bias:
https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/314

I think the resolution of this issue is to compute bias, and remove the matching plot:
https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/313

micahwoodard:
@alexpiet @hanhou the issue says calculate bias online. Are there instances of online calculations elsewhere in the repo or other places? I have not done that before so example code or direction would be helpful

alexpiet:
@micahwoodard By "online" we mean after each trial in the GUI as opposed to calculating the bias "offline" after the session is over. So you will just use a normal function in the code repo that executes either in the main thread, or in a separate thread if its slow.

micahwoodard:
Ohhhhhh okay thank you for that clarification

micahwoodard:
@alexpiet @hanhou Do you want me to import the logistic regression bias functions in the aind-foraging-behavior-bonsai or just implement the functionality within the foraging repo? I see the the logistic regression function linked takes a nwb file as an argument, and  I see that there is a TrasferToNWB py file within the foraging repo, but I'm not seeing where a nwb file is generated or used. Is the idea to generate a nwb based on the GenerateTrials class and then feed that into the logistic regression bias function?

alexpiet:
@micahwoodard We want to implement the functionality into the task repo. 

TransfertoNWB.py is going to be removed soon.

micahwoodard:
[...]

PR #741, opened 2024-10-02
Bias Indicator

micahwoodard:
  
### Describe changes:
- added widget that calculates and graphs logistic regression bias of licks
- [#296](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/296) 
- graph tracking bias will be visible
- [x] 446

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "newscale@git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target",
     "aind-auto-train@git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main",
     "aind-slims-api@git+https://github.com/AllenNeuralDynamics/aind-slims-api@main",
+    "aind-dynamic-foraging-models@git+https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models@main",
     "pynwb",
     "requests",
     "harp-python",
@@ -35,7 +36,9 @@ dependencies = [
     "aind-data-schema==0.36.0",
     "aind-data-schema-models==0.1.10",
     #"pydantic==2.8.2",
-    "pyyaml"
+    "pyyaml",
+    "pyqtgraph"
+
 ]
 
 [tool.setuptools.packages.find]
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ pyserial
 git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target
 git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main
 git+https://github.com/AllenNeuralDynamics/aind-slims-api@main
+git+https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models@main
 pynwb
 requests
 harp-python
@@ -18,3 +19,4 @@ deepdiff
 aind-data-schema==0.36.0
 aind-data-schema-models==0.1.10
 pyyaml
+pyqtgraph
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -126,7 +126,8 @@ class Window(QMainWindow):
         self.threadpool_workertimer=QThreadPool() # for timing
 
         # create bias indicator
-        self.bias_indicator = BiasIndicator()  # TODO: Where to store bias_threshold parameter? self.Settings?
+        self.bias_indicator = BiasIndicator(x_range=50)  # TODO: Where to store bias_threshold parameter? self.Settings?
+        self.bias_n_size = 500
         self.bias_indicator.setSizePolicy(QSizePolicy.Maximum, QSizePolicy.Ignored)
 
         # Set up more parameters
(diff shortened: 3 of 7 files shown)


alexpiet:
Wow, looks great!

My immediate reaction is that the bias plot is taking up too much space. Can you make it smaller horizontally? Like maybe 2/3rds or 1/2 of the current width?

micahwoodard:
> Wow, looks great!
> 
> My immediate reaction is that the bias plot is taking up too much space. Can you make it smaller horizontally? Like maybe 2/3rds or 1/2 of the current width?
> 
> Right now I think you are computing the bias on every trial, and using the last 15 trials to compute the bias. And then only plotting the bias from the last 15 trials. @hanhou What do you think of that?
> 
> I think we generally want Left to be the top, and Right to be the bottom

Oh I thought we were switching the right and left on the other graph so that's why I had it configured like this. I can definitely swap those though. Just to double check, a positive bias value correlates to right: negative, left?

hanhou:
>  I'm just using randomly generated numbers to calculate bias so it would be nice to test maybe with an actual mouse?

In the drop down menu "Tools" you can find a simulator which can simulate animal behavior (use the win-stay-lose-shift one)

alexpiet:
> > I think we generally want Left to be the top, and Right to be the bottom
> 
> Oh I thought we were switching the right and left on the other graph so that's why I had it configured like this. I can definitely swap those though. Just to double check, a positive bias value correlates to right: negative, left?

Well, I can't remember what should be on top. I just noticed the bias is plotted the opposite of the plots on the left, but I guess you haven't flipped those yet. Sorry for the confusion.

The sign of the bias depends on how you called the logistic regression function. When you call it, if right choices are = 1, then positive bias will mean right bias. If left choices are = 1, then positive bias will mean left bias. I think, and @hanhou can confirm, that the code always uses right=1 by convention

hanhou:
> Right now I think you are computing the bias on every trial, and using the last 15 trials to compute the bias. And then only plotting the bias from the last 15 trials. @hanhou What do you think of that?

I would keep track two biases: a global bias computed from trial 1 to trial N of a session, and a local bias from trial N-M to trial N. In my proposed algorithm, the lickspout movement is driven by the local bias, and I use M = 500 (which could span to previous session). I don't know the best hyperparameters, and we have to try. But in general, M should be larger than ~ 200 trials so that the bias estimate is more accurate

image

hanhou:
> We could take some existing data and compute the bias from it (maybe @hanhou already has this somewhere)

Yes, we should do some simulation with existing data. I only computed global bias for each session, as shown in the streamlit app:

image

The overall distribution is like this:

image

It looks like logistic regression sometimes failed ending up with very large biases.

micahwoodard:
@hanhou and @alexpiet do have a minute today to chat about steps forward for this? It would be nice to get some clarity implementation of the global and local biases and testing the best parameters for the regression. I am free anytime today if either of you have a minute.

alexpiet:
Conversation with @micahwoodard, and we decided for the first pass implementation:
- don't compute global bias. @hanhou Its not clear why we need this in the GUI.
- use 0.7 as bias threshold, but seek feedback from RAs whether this matches their intuition
- use M=500 for number of trials to compute bias over. Seek feedback from RAs whether this is jittery or sluggish compared to their intuition
- Compute the bias every 10 trials, this is probably overkill, but just to get a sense of the data
- In the first 0 to M trials, compute the bias over just the trials in this session, since figuring out bias from the last session is (1) hard to implement, (2) unclear scientific value (do mice track bias over days?)
- Probably move bias computation to a separate thread if its slow to compute with M=500
- Long term we would like to store the bias along with other trial history information. I dont know how thats organized or implemented, so I advised @micahwoodard she might want to skip that for the first pass while we figure out the issues above. She should follow up with @XX-Yin about how to organize that information.

micahwoodard:
Okay @alexpiet and @hanhou  this is ready to be reviewed again. I've implemented the changes Alex mentioned above. One thing I did want to draw attention to is that I was having difficulties calculating bias if the trial count was too similar to the n_trial_back parameter. To handle this, I am setting n_trail_back to be 60% of the trial count before 500 trial. I tried to explain what I'm doing in [this comment](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/1fcccc1603707fd4a77191f4abe17171aa2cf8da/src/foraging_gui/Foraging.py#L4199C21-L4201C104). Let me know if there is a better way to handle this. 

Also, another this I wanted to note is that now I am plotting the bias as [bias, trial_count] instead of [bias, bias_count].  I also added logging statements noting bias and trial_count. I changed this so it would be easier to look back on the logs and see the bias trend through out the session. Let me know if this isn't what you envisioned. 

Thanks and hope you have a great weekend!

alexpiet:
@micahwoodard Looks great, a have a handful of cosmetic details to address. But I'll merge it and we can get feedback from the RAs and iterate from there

alexpiet (approved):

XX-Yin:
@micahwoodard There is an error information when using the ForagingGUI_Ephys.ui. Can you take a look?

micahwoodard:
@XX-Yin do you want me to come take a look right now at the Ephys rig for the error information? If so where are the ephys rigs, I've never been to those. Is it like a logging error?

XX-Yin:
@micahwoodard You don't need to test it on the Ephys rig. And this error can be replicated in the main branch if you open the `ForagingGUI_Ephys.ui` GUI by changing the `default_ui` to `ForagingGUI_Ephys.ui` in the `ForagingSettings.json` file. I'm also happy to show you on my own computer if you can't replicate this error.

PR #769, opened 2024-10-10
Added code to integrate stagewidget

jessy-liao:
  
### Describe changes:

- Added code to integrate stage widget (with newest version of stagewidget 1.0.3.dev4)

- Adding stage widget to dynamic foraging gui

- If newscale configuration are set in ForagingSettings.json it will utilize the new stage widget

- On the user side: None
- For SIPE system engineers: deploy stagewidget to systems using dynamicforaginggui. This will install the standalone version for debugging purposes but will also set environment variables needed for the stagewidget package to be installed correctly. 
    - Note: the environment variable to set is 
        - PIP_EXTRA_INDEX_URL=http://eng-tools:3141/aibs/packages (for latest release version)
        - PIP_EXTRA_INDEX_URL=http://eng-tools:3141/aibs/dev (for latest dev version)

- NO, only tested locally
- Tested:
    - Successful integration of stage widget with newscale settings set and not set
    - Successful integration of stage widget for ForagingGUI and ForagingGUI_Ephys
    - Successful logging (stage widget does not override currently existing log)
    - Successful overnight run of stage widget without any errors.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,3 +20,4 @@ aind-data-schema==0.36.0
 aind-data-schema-models==0.1.10
 pyyaml
 pyqtgraph
+stagewidget
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -93,6 +93,9 @@ class Window(QMainWindow):
         # Load Rig Json
         self._LoadRigJson()
 
+        # Stage Widget
+        self.stage_widget = None
+
         # Load User interface
         self._LoadUI()
 
@@ -237,7 +240,8 @@ class Window(QMainWindow):
             for i in reversed(range(layout.count())):
                 layout.itemAt(i).widget().setVisible(False)
             # Insert new stage_widget
-            layout.addWidget(get_stage_widget())
+            self.stage_widget = get_stage_widget()
+            layout.addWidget(self.stage_widget)
 
     def _LoadUI(self):
         '''
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -30,6 +30,8 @@ from PyQt5.QtCore import QThreadPool,Qt,QThread
 from pyOSC3.OSC3 import OSCStreamingClient
 import webbrowser
 
+from StageWidget.main import get_stage_widget
+
 import foraging_gui
 import foraging_gui.rigcontrol as rigcontrol
 from foraging_gui.Visualization import PlotV,PlotLickDistribution,PlotTimeDistribution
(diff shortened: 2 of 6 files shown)


alexpiet:
@jessy-liao by this: "If newstage configuration are set in ForagingSettings.json it will utilize the new stage widget" do you mean "newscale" ?

alexpiet:
@jessy-liao Can you merge into `develop` instead of `main` ?

alexpiet:
@jessy-liao Can you coordinate with @micahwoodard to test this in 446/447?
- Test the SIPE deployment sets environment variables
- Test that the dynamic-foraging-task is still pip installable

jessy-liao:
> @jessy-liao by this: "If newstage configuration are set in ForagingSettings.json it will utilize the new stage widget" do you mean "newscale" ?

Ah yes, I meant newscale. Updated the description.

requirements.txt
```diff
@@ -20,3 +20,4 @@ aind-data-schema==0.36.0
 aind-data-schema-models==0.1.10
 pyyaml
 pyqtgraph
+stagewidget
```
micahwoodard, line 23: Do you also need to add it to the pyproject.toml?

micahwoodard (commented):
Looks good to me! My only thought is stagewidget needs to be added to pyproject.toml unless we're omitting it from there for some reason

requirements.txt
jessy-liao: True, let me fix that real quick

micahwoodard:
@jessy-liao I'm getting errors trying to pip install dynamic-foraging-task in 447 on this branch because it can't find the stagewidget package. I think I remember someone mentioning it was a package on a SIPE specific server? Is that correct and if so, are there additional steps to be able to install packages from there?

src/foraging_gui/Foraging.py
```diff
@@ -208,6 +210,35 @@ def _show_disk_space(self):
...
+        # Get QWidget object
+        widget = getattr(self, widget_to_replace, None)
+
+        if widget is not None:
+            layout = widget.layout()
+            # Hide all current items within widget being replaced
+            for i in reversed(range(layout.count())):
+                layout.itemAt(i).widget().setVisible(False)
+            # Insert new stage_widget
+            layout.addWidget(get_stage_widget())
```
micahwoodard, line 240: Could we make the new stage widget an attribute so it's easy to access if we need it? I'll probably need to use it to move the stage for the autolick port feature. Looking through you widget, I think going through the widget would be the only way to move the stage since the StageModel and device is embedded within. Is that right and would there be any issues with doing that?

micahwoodard:
@jessy-liao I was wondering what the status is on this?

jessy-liao:
@micahwoodard Here are the remaining tasks left for this PR: 
- I need to make the stagewidget into an attribute (agree that this is the only way to interact with the stage)
- Need to bump version to release 

Ill have these done by the end of the week.

micahwoodard (approved):
Looks good to me if you've confirmed it works as expected!

PR #781, opened 2024-10-16
fixes for writing logs to file and console

micahwoodard:
  
### Describe changes:
- importing aind-slims-api messes up logging to file and console. Code now clears root log handlers to write both to file and console as before
- N/A
- None
- None
- [ ] 446

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -44,6 +44,9 @@ from foraging_gui.GenerateMetadata import generate_metadata
 from foraging_gui.RigJsonBuilder import build_rig_json
 from aind_data_schema.core.session import Session
 
+logger = logging.getLogger(__name__)
+logger.root.handlers.clear()    # clear handlers so aind_slims_api doesn't bulldoze log file
+
 class NumpyEncoder(json.JSONEncoder):
     def default(self, obj):
         if isinstance(obj, np.ndarray):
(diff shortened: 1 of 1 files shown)


micahwoodard:
I'm not an expert on logging but this works locally on my machine to restore logs to the txt file and display in console if in settings

alexpiet:
You should double check that this doesn't mess with the slims logging. I'm not an expert with logging either

micahwoodard:
@alexpiet Okay tested on rig and seems to be writing files and logging to console correctly. It doesn't seem to mess with slims logging because the slims logger is just writing to sys.stderr, which is what the gui sets up later on if the settings call for it.

src/foraging_gui/Foraging.py
```diff
@@ -4649,14 +4652,12 @@ def start_gui_log_file(box_number):
...
-    logging.basicConfig(
-        format=log_format,
-        level=logging.INFO,
-        datefmt=log_datefmt,
-        handlers=[
-            logging.FileHandler(logging_filename),
-        ]
-    )
+
+    log_formatter = logging.Formatter(fmt=log_format, datefmt=log_datefmt)
```
alexpiet, line 4656: You don't set the level as logging.INFO, is that the default? I would probably set it explicitly

alexpiet: maybe you are just inheriting the level from the existing logger? In that case I would absolutely set the level

micahwoodard: Oh yeah I forgot to set the level I'll add that in

alexpiet (approved):

Issue #770, opened 2024-10-10
Wrong project name when the mouse is not in the schedule


XX-Yin:
Related to update from https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/694

When the animal ID is not in the schedule, the project information will be forced to set as default. This is problematic. When the animal is training in the ephys rig without the schedule information, the project will also be set to default after clicking start (usually we'd like to use a different project information). Please cancel the force default setting.

XX-Yin:
Or make the `default` optional so that we can disable the default settings in the ephys rig

alexpiet:
- Adding a parameters in `ForagingSettings.json` that disables auto filling the project name seems like a good solution. 
- Or changing the logic so that the default doesn't get set if the project name has already been manually configured.

XX-Yin:
The second solution seems to be more universal.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #783, opened 2024-10-16
Fix default data description

XX-Yin:
  
### Describe changes:
Added the "add_default_project_name" to the settings csv to enable us to turn off the default setting for the project name when the mouse ID is not in the planning spreadsheet. 

resolved https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/770

Please add `"add_default_project_name":false` to the settings csv if you have your own project information when running mice not in the behavior rig. 

No

Tested on my own computer


src/foraging_gui/Foraging.py
```diff
@@ -3917,8 +3918,8 @@ def _Start(self):
                     self.Metadata_dialog.ProjectName.setCurrentIndex(index)
                     self.Metadata_dialog._show_project_info()
                     logging.info('Setting Project name: {}'.format(project_name))
-                    add_default = False
-            if add_default:
+                    self.add_default_project_name = False
+            if self.add_default_project_name:
```
alexpiet, line 3922: This logic seems wrong. We only want to add the default project name if `self.add_default_project_name` is True, AND the (project_name is not None).

So I would add back `add_default`, and then change the if block to `if self.add_default_project_name and add_default`

src/foraging_gui/Foraging.py
```diff
@@ -3918,27 +3941,11 @@ def _Start(self):
...
-                        'Grant Number': ['nan'],
-                        'Investigators': ['Jeremiah Cohen'],
-                        'Fundee': ['nan'],
-                    }
-                    self.Metadata_dialog.project_info = project_info
-                    self.Metadata_dialog.ProjectName.addItems([project_name])
-            self.project_name = project_name
 
+            if self.add_default_project_name and add_default:
+                self._set_default_project()
```
alexpiet, line 3946: It should be `project_name = self._set_default_project()`, right?
And then that function should return `project_name`?

XX-Yin: that's right! Updated

alexpiet:
Can you merge into "develop", instead of main?

src/setting_templates/ForagingSettings.json
```diff
@@ -18,5 +18,6 @@
     "go_cue_decibel_box1":73,
     "go_cue_decibel_box2":73,
     "go_cue_decibel_box3":73,
-    "go_cue_decibel_box4":73
+    "go_cue_decibel_box4":73,
+    "add_default_project_name":false
```
alexpiet, line 22: I would set the default in the settings_template to make the default in the code (true, not false)

XX-Yin: changed

alexpiet (approved):
Looks great, thanks for taking care of this

PR #787, opened 2024-10-18
 log sessions

### LINKED ISSUE #592: Save python log in behavior folder
Right now the GUI saves a log each time it runs. We should save this python log in the saved data for each session. 
- The log is for every time the GUI opens, which could contain multiple sessions
- One option is to simply copy the log to the saved data folder. This would include all activity in the GUI before the session starts
- Another option is to write the logger to an additional file in the saved data folder when the session starts (or when the saved data folder is created), and then stop that additional logging when the session ends (although we want to log until the post-session weight is entered and saved)
- The log should be saved in the "win.TrainingFolder", which is the same folder the .json file is saved in. 
- The log should have be a .txt and should be named `python_gui_log.txt`

micahwoodard:
  
### Describe changes:
- write log to session folder and close upon start of new session
- resolves #592 
- None
- None
- [x] 446

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2273,10 +2273,6 @@ class Window(QMainWindow):
                 event.ignore()
                 return
 
-        # if session log handler is not none, stop logging for previous session
-        # if self.session_log_handler is not None:
-        #     self.end_session_log()
-
         event.accept()
         self.Start.setChecked(False)
         if self.InitializeBonsaiSuccessfully==1:
@@ -3191,10 +3187,6 @@ class Window(QMainWindow):
         self.load_tag=1
         self.ID.returnPressed.emit() # Mimic the return press event to auto-engage AutoTrain
 
-        # if session log handler is not none, stop logging for previous session
-        # if self.session_log_handler is not None:
-        #     self.end_session_log()
-
     def _LoadVisualization(self):
         '''To visulize the training when loading a session'''
         self.ToInitializeVisual=1
@@ -3280,10 +3272,6 @@ class Window(QMainWindow):
         else:
             logging.info('Clearing declined')
             return
-
-        # if session log handler is not none, stop logging for previous session
-        # if self.session_log_handler is not None:
-        #     self.end_session_log()
 
     def _StartFIP(self):
         self.StartFIP.setChecked(False)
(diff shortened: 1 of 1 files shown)


micahwoodard:
Tested locally but need to test on rig. Log of previous session will close users presses start, clear, new session, close, or load button after session. Note that if user declines a pop up box after pressing buttons, log file will not close.

src/foraging_gui/Foraging.py
```diff
@@ -2263,13 +2266,17 @@ def closeEvent(self, event):
...
-                'Box {}, Foraging Close'.format(self.box_letter), 
+                'Box {}, Foraging Close'.format(self.box_letter),
                 'Close the GUI?',
-                QMessageBox.Yes | QMessageBox.No, QMessageBox.Yes)  
+                QMessageBox.Yes | QMessageBox.No, QMessageBox.Yes)
             if reply == QMessageBox.No:
                 event.ignore()
                 return
 
+        # if session log handler is not none, stop logging for previous session
```
alexpiet, line 2276: presumably this was part of testing and can be removed?

src/foraging_gui/Foraging.py
```diff
@@ -3184,6 +3191,10 @@ def _Open(self,open_last = False,input_file = ''):
         self.load_tag=1
         self.ID.returnPressed.emit() # Mimic the return press event to auto-engage AutoTrain
 
+        # if session log handler is not none, stop logging for previous session
+        # if self.session_log_handler is not None:
+        #     self.end_session_log()
```
alexpiet, line 3196: again, can this commented code be removed?

src/foraging_gui/Foraging.py
```diff
@@ -3248,28 +3259,31 @@ def _Clear(self):
...
         # If yes, clear parameters
         if reply == QMessageBox.Yes:
             for child in self.TrainingParameters.findChildren(QtWidgets.QLineEdit)+ self.centralwidget.findChildren(QtWidgets.QLineEdit):
                 if child.isEnabled():
                     child.clear()
+        else:
+            logging.info('Clearing declined')
+            return
+
+        # if session log handler is not none, stop logging for previous session
```
alexpiet, line 3284: again

alexpiet (approved):

PR #803, opened 2024-10-23
change modality in upload_manifest if fib is being run or not

### LINKED ISSUE #793: sessions have mislabeled metadata on modalities
The metadata on modalities (which is used to query for sessions on mongoDB) isn't always correct. 

Per @alexpiet , the modality `fib` is generated automatically if there is a `fiber_photometry_start_time` variable in the main GUI window. This makes sense since the sessions that have the wrong metadata are trained in 447 but finish in the EPHYS rig afterwards. 

The bug may also be in `src/foraging_gui/GenerateMetadata.py`

The fix should be done when the data is generated. we need to make sure that a fiber folder is created to have the `fib` modality. 

Examples of this include: 717263, 714116. 

Fib modality listed here: 
Image

But data doesn't have fiber photometry data

Image

On streamlit we see that it's finished trained on the 323 ephys rig 
Image

micahwoodard:
  
### Describe changes:
- Update the upload manifests to only list "fib" and "behavior-videos" if they are actually recorded in the session
- resolves #793 
- none
- none
- [x] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4605,7 +4605,7 @@ class Window(QMainWindow):
                          '&session_plot_selected_draw_types=1.+Choice+history'
         )
 
-    def _generate_upload_manifest(self,FIP=False):
+    def _generate_upload_manifest(self,FIP=False):  # FIXME: How does this ever get set to True?
         '''
             Generates a manifest.yml file for triggering data copy to VAST and upload to aws
         '''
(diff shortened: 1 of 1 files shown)


micahwoodard:
Still unclear how behavior videos are being set to record in gui but for now, they are included if fib is being run

src/foraging_gui/Foraging.py
```diff
@@ -4616,6 +4616,7 @@ def _generate_upload_manifest(self,FIP=False):
             if not hasattr(self, 'project_name'):
                 self.project_name = 'Behavior Platform'
 
+            # FIXME: How is this if statement working if it does the exact same thing for each condition?
```
alexpiet, line 4619: I think we had originally planned on triggering different code ocean capsules based on data modalities (FIP), but now we just trigger the same capsule. So we could remove the entire `FIP` variable in this function

src/foraging_gui/Foraging.py
```diff
@@ -4625,6 +4626,18 @@ def _generate_upload_manifest(self,FIP=False):
                 capsule_id = 'c089614a-347e-4696-b17e-86980bb782c1'
                 mount = 'FIP'
 
+            # only include fib and behavior videos if fib is being run
```
alexpiet, line 4629: We need to figure out if there is video separately from the FIB

alexpiet (approved):

alexpiet:
@micahwoodard I renamed the PR for future clarity

PR #805, opened 2024-10-25
Feat 682 end fib on timeout

micahwoodard:
  
### Describe changes:
- when _Start function is used when Start button is checked, FIB excitation is turned off if in use
- resolves [#682](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/682)
- if session is stopped, whether from time out or user presses a checked Start button, FIB excitation will end 
- None
- [x] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3995,6 +3995,17 @@ class Window(QMainWindow):
             b_bias_len = len(self.GeneratedTrials.B_Bias)
             self.GeneratedTrials.B_Bias += [last_bias]*((self.GeneratedTrials.B_CurrentTrialN+1)-b_bias_len)
 
+            # stop FIB if Teensy_COM is attribute and if found in com port list
+            if hasattr(self, 'Teensy_COM') and \
+                    self.Teensy_COM in [port.name for port in serial.tools.list_ports.comports()]:
+                ser = serial.Serial(self.Teensy_COM, 9600, timeout=1)
+                # Trigger Teensy with the above specified exp mode
+                ser.write(b's')
+                ser.close()
+                self.TeensyWarning.setText('Stopped FIP excitation')
+                self.TeensyWarning.setStyleSheet(self.default_warning_color)
+                self.fiber_photometry_end_time = str(datetime.now())
+
         if (self.StartANewSession == 1) and (self.ANewTrial == 0):
             # If we are starting a new session, we should wait for the last trial to finish
             self._StopCurrentSession()
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3995,16 +3995,9 @@ class Window(QMainWindow):
             b_bias_len = len(self.GeneratedTrials.B_Bias)
             self.GeneratedTrials.B_Bias += [last_bias]*((self.GeneratedTrials.B_CurrentTrialN+1)-b_bias_len)
 
-            # stop FIB if Teensy_COM is attribute and if found in com port list
-            if hasattr(self, 'Teensy_COM') and \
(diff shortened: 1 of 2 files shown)


micahwoodard:
I'm not sure if this is the exact functionality desired because this will also turn of FIB if a user manually stops a session by clicking the check start button. Let me know if we only want to stop if the session is automatically stopped

micahwoodard:
Also I realized that the _CheckStop function didn't actually call _Start, it only unchecked the button. From my testing, it didn't seem like there was any unintended consequences to calling _Start to perform the stopping logic for an automatically stopped session, but wanted a second opinion

hagikent:
Thanks Micah. It makes sense. To me it's not 100% clear what would happen to the pre-defined baseline, where we intend to record spontaneous activity for 5-10min after the behavior ends (so triggering should be still on).

micahwoodard:
Ahh okay so I should probably take turning of the FIB excitation out of the _Start function and put it in the _CheckStop function if in some cases we need FIB excitation to stay on after unchecking the Start button

micahwoodard:
I left calling _Start in the _CheckStop function because it makes sense to me to call the stop logic for a session if automatically stopped, but let me know if there is logic that we want to skip for automatically stopped sessions.

hagikent:
Hmm... I see, I already see a few edge cases. Let's briefly discuss this next Tue9.

micahwoodard:
@hagikent based on the update you provided, I believe this pull request should work how you want. If the task is auto stopped, excitation is turned off. If a user clicks the start button, excitation remains on. Could you clarify on where you want to 10-5 second buffer?

alexpiet:
@hagikent I believe this will stop excitation, but won't stop the FIP workflow. So the data will continue to be collected, but with no signal. To my knowledge we don't have a way to stop the FIP workflow programmatically. Is that ok?

hagikent (approved):

hagikent:
@micahwoodard 
thanks. re 10min, between `behavior autostop` and `excitation turn-off`

@alexpiet 
Yes, afaik there is no way to stop bonsai. I reconfirmed this point with @ellahiltonvano  so RAs are still supposed to stop FIPbonsai instance manually.

Note, when excitation stops, FPsensors(CMOSs) are no longer sending images to Bonsai. So, even tho bonsai is still running, it is recording only videos (and harp events) no FP anymore. 
The reason we want to stop excitation automatically is to avoid recording "fiber disconnection signal" that messs up some filtering-based preprocessing (e.g. NaN issue raised by Johannes).

Hope this clarifies. Thanks for double-checking.

hagikent:
@micahwoodard 
A couple of issues spotted while Integration Testing.

1. Automatic stop -> pop-up window -> "yes" -> excitation immediately stopped. 
   Can we have a 5s baseline here? 
2. Automatic stop -> pop-up window -> "No" -> behavior was restarting but excitation immediately stopped. 
   Excitation shouldn't stop when "No" is selected.

____
When behavior is restarted, ignored trial number is not reset but accumulated, and thus immediately giving you automoatic stop if a mouse ignores 1 trial right after break. This should be fixed. (Margaret told me there is a separate issue/report on this)

micahwoodard:
@hagikent  Right now, I think there are two pop ups when the session is automatically stopped. One that alerts user the session is stopping and then another that is triggered by calling the _Start function that says "'Stop current session?" Is the "'Stop current session?" pop up the one you are referring to in 2.?

hagikent:
@micahwoodard 
> Is the "'Stop current session?" pop up the one you are referring to in 2.?

Yes.

Yeah, saying "No" would be super rare cases.
Thanks, that would be great!

hagikent:
After pressing "no" on the pop-up.

Issue #308, opened 2024-03-19
Update, division by zero


alexpiet:
logfile at `/allen/programs/mindscope/workgroups/behavioral-dynamics/aind_logs/W10DT713883_gui_log/W10DT713883-A_gui_log_2024-03-18_14-28-19.txt`

```
02:46:17 PM:INFO:Foraging:Foraging.py:_StartTrialLoop:line 2993:Current trial: 9
02:46:17 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 54:['/TrialStartTime', 438541.254976, '127.0.0.1', 4002]
02:46:17 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 54:['/ITIStartTimeHarp', 438541.259488, '127.0.0.1', 4002]
02:46:17 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 54:['/LeftLickTime', 438541.316032, '127.0.0.1', 4003]
02:46:17 PM:WARNING:warnings:warnings.py:_showwarnmsg:line 109:C:\Users\svc_aind_ephys\Documents\GitHub\dynamic-foraging-task\src\foraging_gui\Visualization.py:269: RuntimeWarning: invalid value encountered in scalar divide
  choice_R_frac[idx]=LeftChoiceN/(LeftChoiceN+RightChoiceN)

02:46:17 PM:ERROR:Visualization:Visualization.py:_Update:line 92:division by zero
```

PR #807, opened 2024-10-25
Lick interval update

micahwoodard:
  
### Describe changes:
- updates user if percentage of same and cross side licks under 100 ms is larger than 10%
- resolves [#308](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/308)
- text boxes in the lick distribution dialog box will display warning if percentage of same and cross side licks under 100 ms is larger than 10%
- N/A
- [x] 446/447


micahwoodard:
I've tested what I could in the UI and then tested the calculate function outside, but I am unsure how to test without a real licks? Using the simulated mouse doesn't seem to use the GenerateTrials but I could be wrong and am not seeing it. Any ideas would be great!

micahwoodard:
- load old data and trigger function
- ask Ella if there is a mouse to test on

src/foraging_gui/MyFunctions.py
```diff
@@ -890,7 +897,56 @@ def _LickSta(self,Trials=None):
...
+    def calculate_inter_lick_intervals(self):
+        """
+        Calculate and categorize lick intervals
+        """
+
+        right = self.B_RightLickTime
+        left = self.B_LeftLickTime
+        threshold = .1
+
+        if right is not None and left is not None and len(right) > 0 and len(left) > 0:
```
alexpiet, line 910: have you tried the edge case of just one right, or just one left lick?

src/foraging_gui/MyFunctions.py
```diff
@@ -890,7 +897,56 @@ def _LickSta(self,Trials=None):
...
+            # calculate left interval and fraction
+            same_side_l = np.diff(left)
+            same_side_l_frac = len(same_side_l[~(same_side_l > threshold)]) / len(same_side_l)
```
alexpiet, line 913: On line 913 you can considering clarify this code:
```
same_side_l_frac = len(same_side_l[~(same_side_l > threshold)]) / len(same_side_l)
```
into this
```
same_side_l_frac = np.mean(same_side_l  <= threshold)
```

Notes:
- `~(same_side_l > threshold)` is the same as `same_side_l <= threshold`
- You can get the count of a boolean array with `np.sum(same_side_l <=threshold)`, and then divide by `len(same_side_l)`
- You can get the % true of a boolean array with `np.mean(same_side_l <= threshold)`

src/foraging_gui/MyFunctions.py
```diff
@@ -890,7 +897,56 @@ def _LickSta(self,Trials=None):
...
+            logging.info(f'Percentage of left lick intervals under 100 ms is {same_side_l_frac * 100}%.')
+            self.B_LeftLickIntervalPercent = same_side_l_frac * 100
+
+            # calculate right interval and fraction
+            same_side_r = np.diff(right)
+            same_side_r_frac = len(same_side_r[~(same_side_r > threshold)]) / len(same_side_r)
+            logging.info(f'Percentage of right lick intervals under 100 ms is {same_side_r_frac * 100}%.')
+            self.B_RightLickIntervalPercent = same_side_r_frac * 100
+
+            if same_side_l_frac + same_side_r_frac >= .1:
```
alexpiet, line 923: I'm not sure if adding these fractions is what we want. 
I think if either the left or right is above .1 (which you should set as a variable instead of hard coding), then we should alert the user

src/foraging_gui/MyFunctions.py
```diff
@@ -890,7 +897,56 @@ def _LickSta(self,Trials=None):
...
+                self.win.same_side_lick_interval.setText(f'Percentage of same side lick intervals under 100 ms is '
+                                                         f'over 10%: {(same_side_r_frac+same_side_l_frac) * 100}%.')
+            else:
+                self.win.same_side_lick_interval.setText('')
+
+            # calculate cross side interval and frac
+            dummy_array = np.ones(right.shape)  # use array to tell lick side after merging
+            stacked_right = np.column_stack((dummy_array, right))
+            stacked_left = np.column_stack((np.negative(dummy_array), left))
+            merged_sorted = sorted(np.concatenate((stacked_right, stacked_left)),
```
alexpiet, line 933: I'm not sure I understand what `merged_sorted` contains

src/foraging_gui/MyFunctions.py
```diff
@@ -890,7 +897,56 @@ def _LickSta(self,Trials=None):
...
+                                   key=lambda x: x[1])  # sort by second element
+            cross_sides = []
+            for i in range(1, len(right)):
+                if merged_sorted[i - 1][0] == -merged_sorted[i][0]:  # cross side lick
+                    cross_sides.append(merged_sorted[i][1] - merged_sorted[i - 1][1])
+            cross_sides = np.array(cross_sides)
+            cross_side_frac = len(cross_sides[~(cross_sides > threshold)]) / len(cross_sides)
```
alexpiet, line 940: This code block is very confusing, and I'm not totally sure whats going on. 

I think I would have implemented this with a while loop and a pointer in each of the left and right lick times

src/foraging_gui/MyFunctions.py
micahwoodard: Okay I was doing that based on what you said [here](https://github.com/orgs/AllenNeuralDynamics/projects/47/views/10?sortedBy%5Bdirection%5D=asc&sortedBy%5BcolumnId%5D=Assignees&sortedBy%5Bdirection%5D=desc&sortedBy%5BcolumnId%5D=Status&pane=issue&itemId=69427925&issue=AllenNeuralDynamics%7Caind-behavior-blog%7C308#:~:text=If%20the%20proportion%20of%20intervals%20is%20%3E10%25%20for%20same%20side%2C%20show%20a%20message.%20(That%20is%2C%2010%25%20of%20all%20same%2Dside%20intervals%2C%20combining%20across%20left%2Dleft%20and%20right%2Dright)). I will change to if either is over the threshold

alexpiet: I suppose its hard to implement things when I tell you inconsistent things. Sorry!

I'm not sure if we want to check if either left or right same side, or combine across sides. But if we do combine, then instead of adding the fractions you should instead concatenate the array of diffs and then compute the fraction:

```
same_side_l = np.diff(left)
same_side_r = np.diff(right)
same_side_combined = np.concat([same_side_l, same_side_r])
same_side_frac = np.mean(same_side_combined <= threshold)
```

If we add the fractions, then the combined fraction ranges from 0 to 2, so that doesn't make sense. 

I guess I would err on the side of checking the combined fraction, since it should be more stable when there are a small number of licks

src/foraging_gui/MyFunctions.py
alexpiet: Ok, I understand!

src/foraging_gui/MyFunctions.py
alexpiet: Ok, I understand whats happening now and it makes sense. I would add a few comments so we can understand how this works in the future. Just explain the structure of `merged_sorted` , and `cross_sides` (which contains only the ITIs for cross side intervals

micahwoodard:
Tested with mocked data. @ellahiltonvano would there be a good mouse to test this feature on? I currently can't test on a rig because the lick times will always be empty. I am fairly confident it will work based on mocked data so if we can't test with a real mouse, I think it will probably be okay

alexpiet (approved):
Feel free to merge after you finished any testing. A screenshot would be nice before merging.

ellahiltonvano:
@micahwoodard as discussed we could use one of madeline's training mice that start next week since they are not assigned to a particular scientist. If you want to merge before then we can just do normal testing on monday

alexpiet:
> * Will it be visible if the user doesn't click on the lick distribution dialog?
> 
> No. Is that something we want? I think I assumed that the message should be on the lick distribution dialog but should it be a separate window or pop up?

Up to @ellahiltonvano

ellahiltonvano:
@alexpiet I think if we want to use this as a flag that something is going on with the lick detection board then eventually a popup to alert the user would be helpful. This would requires scis defining a very specific value of what is acceptable though

Currently:
ILI > 0.05 is good, no tuning necessary
ILI < 0.05 indicates crosstalk (right to left licks, left to right)
An ILI < 0.05 indicates rebound (left and right licks)

Here is the [SOP](https://alleninstitute.sharepoint.com/:w:/s/NeuralDynamics/EUFfxQ-ah3pOrWvPh_Y5YnoBPr3rGa6UtkP09c5rVo4J0A?e=RPT9fB), for reference. Let me know if y'all think it would be helpful to just code in alarms for these specific abnormalities.

micahwoodard:
@alexpiet, Ella and I chatted about this a little bit. Instead of a pop up that would prompt RA's to alert someone/submit a ticket, would it make sense to log an error which will create an issue if either same side licks or cross side licks exceeds the threshold? This would ensure that these alerts won't get lost or ignored with the other windows. Might take some work making sure it only logs one error and not all instances, but would ensure this issue is being looked at

alexpiet:
> @alexpiet, Ella and I chatted about this a little bit. Instead of a pop up that would prompt RA's to alert someone/submit a ticket, would it make sense to log an error which will create an issue if either same side licks or cross side licks exceeds the threshold? This would ensure that these alerts won't get lost or ignored with the other windows. Might take some work making sure it only logs one error and not all instances, but would ensure this issue is being looked at

That works! I would still add a warning in the warning box (especially if you make a queue log)

micahwoodard:
> That works! I would still add a warning in the warning box (especially if you make a queue log)
The queue log should also display all error log messages so that should be taken care of if implemented

micahwoodard:
@ellahiltonvano is there a mouse that we can test with this week? If so what time/works best for you?

ellahiltonvano:
@micahwoodard we could run one of madeline's training mice on a branch with this update. Wednesday afternoon (12:30) would work!

micahwoodard:
Yep that is perfect thanks Ella!

micahwoodard:
logging error
- add box number 
- mouse
-

micahwoodard:
Tested with Ella and Madeline this afternoon so merging in

PR #870, opened 2024-10-31
Remove data description and update upload manifest

### LINKED ISSUE #804: Either update data_description.json, or let the transfer service generate it for us
Right now we are making a data_description.json and then uploading it. We can also not make one, and let the transfer service generate it for us. 

If we want to update the data_description.json
@dyf wrote on teams:
> if you want to continue making the data description on the rig, if you aren't already I suggest you fetch funding information from http://aind-metadata-service/ rather than hard-code it. e.g. http://aind-metadata-service/funding/Behavior%20Platform

If we want to let the transfer service generate it for us:
> The other thing we could do is pass modalities and project name to the DTS and it will take care of it for you.

This means making sure the upload manifest contains the accurate modalities in session.json rather, and project_name

micahwoodard:
  
### Describe changes:
- Removed creation of data description and use session to populate the upload manifest with correct modalities
- resolves #804 
- None
- None
- [ ] 446/7


micahwoodard:
Based on the [issue](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/804), I'm still a little unclear if I need to remove any hardcoded funding information. Let me know! Also a little unclear to me how to update the project name. It seems like this is the logic to [set the project name](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/a66a1f16c4ff6d754ffd5d050388e60f80775e4a/src/foraging_gui/Foraging.py#L3990C12-L4005C45. Any help there would be great!

XX-Yin:
I have two questions: 
Whether the data description has been automatically generated by the transfer service? 
Are there any advantages to using the transfer service to generate data descriptions?

alexpiet:
> I have two questions: Whether the data description has been automatically generated by the transfer service? Are there any advantages to using the transfer service to generate data descriptions?

The transfer service will automatically create the data description is one is not provided.

The advantage of letting the transfer service generate the data description is it will track project funding properly

XX-Yin:
"The transfer service will automatically create the data description is one is not provided." 
We may need to wait for the finish of automatically generating data description from the transfer service, otherwise no data description is provided after uploading the data.

alexpiet:
> "The transfer service will automatically create the data description is one is not provided." We may need to wait for the finish of automatically generating data description from the transfer service, otherwise no data description is provided after uploading the data.

I'm not sure what you mean. The transfer service does this already, it only takes a few seconds to generate

XX-Yin:
"The advantage of letting the transfer service generate the data description is it will track project funding properly" 

I got it, but it seems to be a minor improvement. The transfer service must fetch the funding information from a fixed source based on the mouse information. The GUI may do the same thing. Anyway, both ways are OK for me.

alexpiet (approved):

XX-Yin:
> > "The transfer service will automatically create the data description is one is not provided." We may need to wait for the finish of automatically generating data description from the transfer service, otherwise no data description is provided after uploading the data.
> 
> I'm not sure what you mean. The transfer service does this already, it only takes a few seconds to generate

I misunderstood it. Never mind!

XX-Yin:
@micahwoodard Have you tested on one of the rigs to see if the metadata generation is correct?

micahwoodard:
I'm testing on a rig now, but I'm having issues seeing where metadata is generated if you run a FIP session. Is it not within the _Save function during a session? Whenever I run a session with FIP, I'm not seeing metadata generation or an upload_manifest

XX-Yin:
Does it only happen for FIB sessions?

XX-Yin:
I tested a video session in the ephys rig. It looks good. The session and rig metadata and manifest file were generated. 
image

XX-Yin:
Do we have the same issue for FIB session in the current main branch?

micahwoodard:
It was actually an issue with how I was running the session so with Ella's help I just confirmed that the data_description file is correctly writing the modalities

PR #874, opened 2024-10-31
Fix  finish frac

micahwoodard:
  
### Describe changes:
- finish_frac will now be calculated correctly
- resolves [#721](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/721)
- finish_frac will go below .5 
- None
- [x] 446/7


micahwoodard:
I mistakenly set the TP_RunLength to 10 which is what was being used to calculate choice_frac. After looking at what TP_RunLength was doing, I realized it was being set as the kernel size which should be 2.

alexpiet (approved):

Issue #445, opened 2024-05-14
Remove "Plot settings" from the GUI


alexpiet:
The main GUI window has fields for plot settings. These should be hard coded, and removed from the GUI to reduce visual clutter.

PR #926, opened 2024-11-04
changing kernel size to 10

### LINKED ISSUE #877: visualization bug
@micahwoodard @alexpiet 
The RunLength was set to a fixed value, which caused the update of the figure is not smooth. This variable is set and changed in the GUI. 

Left is the current code, and the right is my previous code. Please make it set by GUI instead of hardcoded!

image
image
image

micahwoodard:
  
### Describe changes:
- change kernel size to 10
- resolves #877 
- graph better sized
- None
- [ ] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -30,7 +30,7 @@ from PyQt5.QtCore import QThreadPool,Qt,QThread
 from pyOSC3.OSC3 import OSCStreamingClient
 import webbrowser
 
-from StageWidget.main import get_stage_widget
+#from StageWidget.main import get_stage_widget
 
 import foraging_gui
 import foraging_gui.rigcontrol as rigcontrol
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -30,7 +30,7 @@ from PyQt5.QtCore import QThreadPool,Qt,QThread
 from pyOSC3.OSC3 import OSCStreamingClient
 import webbrowser
 
-#from StageWidget.main import get_stage_widget
+from StageWidget.main import get_stage_widget
 
 import foraging_gui
 import foraging_gui.rigcontrol as rigcontrol
--- a/src/foraging_gui/Visualization.py
+++ b/src/foraging_gui/Visualization.py
@@ -129,7 +129,7 @@ class PlotV(FigureCanvas):
         RightChoice_UnRewarded=np.where(np.logical_and(self.B_AnimalResponseHistory==1, self.B_RewardedHistory[1]==False))
 
         # running average of choice
-        self.kernel_size=2
+        self.kernel_size=10
         ResponseHistoryT=self.B_AnimalResponseHistory.copy()
         ResponseHistoryT[ResponseHistoryT==2]=np.nan
         ResponseHistoryF=ResponseHistoryT.copy()


XX-Yin:
Hi @micahwoodard 

Can you set the kernel_size by the RunLength from the GUI?

self.RunLength=win.RunLength.text



        # running average of choice
        if self.RunLength()!='':
            kernel_size = int(self.RunLength())
            if kernel_size==1:
                kernel_size=2
                self.RunLengthSetValue(2)
        else:
            kernel_size=2
            self.RunLengthSetValue(2)
        self.kernel_size=kernel_size

micahwoodard:
The issue with that is that those widgets were removed in a previous [pull ](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/445). We either need to add the plot settings back in or hard code this value to 10

XX-Yin:
This parameter is important. This widget is not removed from the ephys GUI. You can directly copy it to the behavior GUI. 
image

micahwoodard:
What does this parameter do? Plot settings were removed from the other gui because they weren't being changed or used by users. I am reluctant to add them back in because of their disuse

micahwoodard:
Also what values does self.RunLength() get set to in your use cases?

XX-Yin:
> What does this parameter do? Plot settings were removed from the other gui because they weren't being changed or used by users. I am reluctant to add them back in because of their disuse

This parameter controls the smoothness of the choice plot, which is important to check it online in some experiments (e.g. in optogenetics, we'd like to see the choice in higher resolution, so the RunLength is set to a smaller value than normal).

alexpiet:
> > What does this parameter do? Plot settings were removed from the other gui because they weren't being changed or used by users. I am reluctant to add them back in because of their disuse
> 
> This parameter controls the smoothness of the choice plot, which is important to check it online in some experiments (e.g. in optogenetics, we'd like to see the choice in higher resolution, so the RunLength is set to a smaller value than normal).

Why do you want the ability to check this in the GUI? For analysis, I understand why you want to look at different resolution, but do you use this for online QA?

XX-Yin:
> > > What does this parameter do? Plot settings were removed from the other gui because they weren't being changed or used by users. I am reluctant to add them back in because of their disuse
> > 
> > 
> > This parameter controls the smoothness of the choice plot, which is important to check it online in some experiments (e.g. in optogenetics, we'd like to see the choice in higher resolution, so the RunLength is set to a smaller value than normal).
> 
> Why do you want the ability to check this in the GUI? For analysis, I understand why you want to look at different resolution, but do you use this for online QA?

Yes, I use it online for visualizing the optogenetics perturbation.

alexpiet:
> > > > What does this parameter do? Plot settings were removed from the other gui because they weren't being changed or used by users. I am reluctant to add them back in because of their disuse
> > > 
> > > 
> > > This parameter controls the smoothness of the choice plot, which is important to check it online in some experiments (e.g. in optogenetics, we'd like to see the choice in higher resolution, so the RunLength is set to a smaller value than normal).
> > 
> > 
> > Why do you want the ability to check this in the GUI? For analysis, I understand why you want to look at different resolution, but do you use this for online QA?
> 
> Yes, I use it online for visualizing the optogenetics perturbation.

But is this just for viewing the data? Or do you modify the way the experiment is going as a result of this perturbation?

XX-Yin:
It's mainly for online monitoring.

alexpiet:
I think we need to move away from treating the GUI as a tool for visualizing the behavior, since its goal is to collect data. Trying to make the GUI too much is part of why its such a messy code base. 

But I know you might disagree. So lets merge this update in, to set the kernel size back to 10, and we can discuss later about what features we want the GUI to support.

alexpiet (approved):

XX-Yin:
> I think we need to move away from treating the GUI as a tool for visualizing the behavior, since its goal is to collect data. Trying to make the GUI too much is part of why its such a messy code base.
> 
> But I know you might disagree. So lets merge this update in, to set the kernel size back to 10, and we can discuss later about what features we want the GUI to support.

How does this affect the code after adding this functionality?

XX-Yin:
@alexpiet What's exactly your concern? It's just an extra parameter.

alexpiet:
> @alexpiet What's exactly your concern? It's just an extra parameter.

Its not just this one parameter. Its the multiplicative effect of extra parameters, code blocks, and functions. If we want to update something we have to check more cases and conditions.

XX-Yin:
> > @alexpiet What's exactly your concern? It's just an extra parameter.
> 
> Its not just this one parameter. Its the multiplicative effect of extra parameters, code blocks, and functions. If we want to update something we have to check more cases and conditions.

@alexpiet You removed this important functionality before and now it's just to restore it. Also, the streamlit doesn't support visualization in different resolution right now (it seems to be difficult to achieve this in the streamlit). We should also ensure that the GUI is functionally complete in terms of basic visualization and other basic functions. In the future, individuals outside of this institute who wish to replicate our behavioral setup won’t have access to the daily Streamlit updates. However, they should still be able to use the GUI to monitor the training data.

alexpiet:
Lets merge this in, because we both agree kernel size of 10 is better than 2. 

Then we can continue this discussion. My point is that the GUI shouldn't be a tool for basic visualization! The GUI should be for data acquisition

XX-Yin:
> Lets merge this in, because we both agree kernel size of 10 is better than 2.
> 
> Then we can continue this discussion. My point is that the GUI shouldn't be a tool for basic visualization! The GUI should be for data acquisition

This is important to make a complete update here!! We need basic visualization for the GUI for normal monitoring and visualization. The choice visualization we are currently updating here is basic enough. How do you define the basic visualization?

alexpiet:
@XX-Yin I moved this discussion to a separate issue: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/955

PR #681, opened 2024-09-10
Feat 517 upload manifest fix

micahwoodard:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
- generate upload manifest is called after acquisition_datetime has been populated
### What issues or discussions does this update address?
[#517](https://github.com/AllenNeuralDynamics/aind-behavior-blog)
### Was this update tested in 446/447?
- [x] 446

PR #991, opened 2024-11-11
manual water water and valve time are same as earned reward

micahwoodard:
  
### Describe changes:
- manual water water and valve time are same as earned reward
- resolves [#681](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/681)
- user cannot change manual water 
- None
- [x] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -361,12 +361,8 @@ class Window(QMainWindow):
         self.actionOpen_video_folder.triggered.connect(self._OpenVideoFolder)
         self.LeftValue.textChanged.connect(self._WaterVolumnManage1)
         self.RightValue.textChanged.connect(self._WaterVolumnManage1)
-        self.GiveWaterL.textChanged.connect(self._WaterVolumnManage1)
-        self.GiveWaterR.textChanged.connect(self._WaterVolumnManage1)
         self.LeftValue_volume.textChanged.connect(self._WaterVolumnManage2)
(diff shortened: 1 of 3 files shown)


micahwoodard:
https://github.com/user-attachments/assets/28400683-410a-414c-bb25-5a2711b70988

micahwoodard:
Tested on Ephys gui as well

XX-Yin:
@micahwoodard Thanks for making this update. 

"user cannot change manual water" Does this mean this field is forced to match the earned water and the user can't change it to a different value?

alexpiet:
@micahwoodard This looks great to me.

@XX-Yin Yes, the user will not be able to give a manual water reward that is a different size from the earned reward. Note the user can always give multiple rewards

alexpiet (approved):

XX-Yin:
> @micahwoodard This looks great to me.
> 
> @XX-Yin Yes, the user will not be able to give a manual water reward that is a different size from the earned reward. Note the user can always give multiple rewards

Alternatively, it's set to match the earned reward by default, but the user can still make changes. 

I'm not sure if other scientists would like to change the manual reward size in certain situations. 
@ZhixiaoSu @KanghoonJ @hagikent @galenlynch

micahwoodard:
Maybe we can bring it up tomorrow in the meeting. I'll add it to the agenda

XX-Yin:
> Maybe we can bring it up tomorrow in the meeting. I'll add it to the agenda

Sounds good!

micahwoodard:
I'm merging this in because no scientist had an issue with it in the meeting today

PR #1003, opened 2024-11-14
creating model and saving

### LINKED ISSUE #799: Implement AindBehaviorSessionModel
https://allenneuraldynamics.github.io/Aind.Behavior.Services/api.base/session.html
- [x] Implement session model. When a user loads a mouse, generate this schema and validate it.
- [x] Save the model as a json file. Presumably in the `TrainingFolder` directory?
- [x] Refactor the code such that this information is always pulled from the model, instead of loose variables in the Window class. 
- [ ] Add to NWB file (Not entirely clear where)
    - Maybe this session model gets added into the AIND session metadata file, under parameters. Then it will be included in the NWB file

micahwoodard:

### Describe changes:

AINDBehaviorSessionModel is created in the init of the UI with default values that will be updated as session information is created.  Attributes of the main window that are now nested in the data model are as follows:
       
            self.behavior_session_model.experiment = self.Task.currentText()
            self.behavior_session_model.experimenter = [self.Experimenter.text()]
            self.behavior_session_model.date = N/A
            self.behavior_session_model.root_path = self.TrainingFolder
            self.behavior_session_model.session_name = self.session_name
            self.behavior_session_model.subject = self.ID.text()
            self.behavior_session_model.experiment_version = self.version
            self.behavior_session_model.notes = self.ShowNotes.toPlainText(),
            self.behavior_session_model.commit_hash = self.commit_ID
            self.behavior_session_model.allow_dirty_repo = self.repo_dirty_flag
            
If the Task, Experimenter, ID, or ShowNotes widgets are edited by user, the behavior_session_model is updated with latest values. The model is validated and saved at the stopping of a session. If validation errors are raised, the errors are logged but the session model is still saved. The session model will be saved to the root_path (previously TrainingFolder) with the prefix `'behavior_session_model_{session_name}.json'`. Naming conventions of root_path and session names remain the same. The model's date is a datetime object, not a string and can be formatted to match currently used convention 

- resolves #799 
- None
- None
- [x] 446/7

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -2335,7 +2335,7 @@ class AutoTrainDialog(QDialog):
         self._setup_allbacks()
         
         # Sync selected subject_id
-        self.update_auto_train_fields(subject_id=self.MainWindow.ID.text())
+        self.update_auto_train_fields(subject_id=self.MainWindow.behavior_session_model.subject)
                 
     def _setup_allbacks(self):
         self.checkBox_show_this_mouse_only.stateChanged.connect(
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -49,8 +49,6 @@ from aind_data_schema.core.session import Session
 from aind_data_schema_models.modalities import Modality
 from aind_behavior_services.session import AindBehaviorSessionModel
 
-from pprint import pprint
-
 logger = logging.getLogger(__name__)
 logger.root.handlers.clear() # clear handlers so console output can be configured
 
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -29,6 +29,7 @@ from PyQt5 import QtWidgets,QtGui,QtCore, uic
 from PyQt5.QtCore import QThreadPool,Qt,QThread
 from pyOSC3.OSC3 import OSCStreamingClient
 import webbrowser
+from pydantic import ValidationError
 
 from StageWidget.main import get_stage_widget
 
@@ -3977,6 +3978,11 @@ class Window(QMainWindow):
             # stop lick interval calculation
             self.GeneratedTrials.lick_interval_time.stop()  # stop lick interval calculation
 
+            # validate behavior session model and document validation errors if any
+            try:
+                AindBehaviorSessionModel(**self.behavior_session_model.model_dump())
+            except ValidationError as e:
+                logging.error(str(e), extra={'tags': [self.warning_log_tag]})
             # save behavior session model
             with open(self.behavior_session_modelJson, "w") as outfile:
                 outfile.write(self.behavior_session_model.model_dump_json())
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -49,6 +49,8 @@ from aind_data_schema.core.session import Session
 from aind_data_schema_models.modalities import Modality
 from aind_behavior_services.session import AindBehaviorSessionModel
 
+from pprint import pprint
+
 logger = logging.getLogger(__name__)
 logger.root.handlers.clear() # clear handlers so console output can be configured
 
@@ -402,7 +404,9 @@ class Window(QMainWindow):
         self.Task.currentTextChanged.connect(lambda task: setattr(self.behavior_session_model, 'experiment', task))
         self.Experimenter.textChanged.connect(lambda text: setattr(self.behavior_session_model, 'experimenter', [text]))
         self.ID.textChanged.connect(lambda subject: setattr(self.behavior_session_model, 'subject', subject))
-        self.ShowNotes.textChanged.connect(lambda notes: setattr(self.behavior_session_model, 'notes', notes))
+        self.ShowNotes.textChanged.connect(lambda: setattr(self.behavior_session_model, 'notes',
+                                                           self.ShowNotes.toPlainText()))
+
 
         # Set manual water volume to earned reward and trigger update if changed
         for side in ['Left', 'Right']:
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -112,15 +112,14 @@ class Window(QMainWindow):
 
         # create AINDBehaviorSession model to be used and referenced for session info
         self.behavior_session_model = AindBehaviorSessionModel(
-            experiment=lambda: self.Task.currentText(),
-            experimenter=[lambda: self.Experimenter.text()],
+            experiment=self.Task.currentText(),
+            experimenter=[self.Experimenter.text()],
             date=datetime.now(),   # update when folders are created
             root_path='',         # update when created
-            session_name=lambda: f'{self.behavior_session_model.subject}_'
-                                 f'{ self.behavior_session_model.date.strftime("%Y-%m-%d_%H-%M-%S")}',
-            subject=lambda: self.ID.text(),
+            session_name= '',   # update when date and subject are filled in 
+            subject=self.ID.text(),
             experiment_version=foraging_gui.__version__,
-            notes=lambda:self.ShowNotes.toPlainText(),
+            notes=self.ShowNotes.toPlainText(),
             commit_hash= subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip(),
             allow_dirty_repo=subprocess.check_output(['git','diff-index','--name-only', 'HEAD']).decode('ascii').strip() != '',
             skip_hardware_validation=True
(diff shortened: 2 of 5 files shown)


micahwoodard:
First pass at adding and saving AindBehaviorSessionModel. Still a rough draft and a couple of things to be done like : 
- [ ] Refactor the code such that this information is always pulled from the model, instead of loose variables in the Window class. 
- [ ] Add to NWB file (Not entirely clear where)
    - Maybe this session model gets added into the AIND session metadata file, under parameters. Then it will be included in the NWB file

but wanted to get eyes on it. No Need to review immediately!

src/foraging_gui/Foraging.py
```diff
@@ -109,6 +110,21 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         # Load User interface
         self._LoadUI()
 
+        # create AINDBehaviorSession model to be used and referenced for session info
+        self.behavior_session_model = AindBehaviorSessionModel(
+            experiment=self.Task.currentText(),
+            experimenter=[self.Experimenter.text()],
+            date=datetime.now(),   # update when folders are created
+            root_path='',         # update when created
+            session_name= '',   # update when date and subject are filled in
```
bruno-f-cruz, line 119: I would pass a null for the session name since this field is generated automatically at model instantiation.

bruno-f-cruz (commented):
I only reviewed the stuff i am familiar with.

src/foraging_gui/Foraging.py
```diff
@@ -109,6 +110,21 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
...
+            subject=self.ID.text(),
+            experiment_version=foraging_gui.__version__,
+            notes=self.ShowNotes.toPlainText(),
+            commit_hash= subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip(),
+            allow_dirty_repo=subprocess.check_output(['git','diff-index','--name-only', 'HEAD']).decode('ascii').strip() != '',
```
alexpiet, line 124: I think we want "allow_dirty_repo" to be "True". Since we are allowing people to run with dirty code. 

The subprocess call you make will report if the repo is dirty, which is different from what the schema is asking https://allenneuraldynamics.github.io/Aind.Behavior.Services/api.base/session.html
@bruno-f-cruz please clarify if I'm wrong.

src/foraging_gui/Foraging.py
alexpiet: The "AindBehaviorSessionModel" generates the name? Even if you pass in a name?

The docs say "Name of the session. This will be used to create a folder in the root path." (https://allenneuraldynamics.github.io/Aind.Behavior.Services/api.base/session.html)
Does creating the schema make the folder?

Right now our code sets the session name and makes a folder when the session starts. Note that the user can open the GUI and do things before starting the session. I think Micah was planning on setting the session name to the empty string until the session actually starts, and then pass in the final session name

micahwoodard: I was playing around with that but I think the time format is different that the one currently used. It seems like the  [format_datetime ](https://github.com/AllenNeuralDynamics/Aind.Behavior.Services/blob/0654486951b766ea6ef2f767c75949adf25efa1b/src/aind_behavior_services/utils.py#L411C1-L420C1) uses "%Y-%m-%dT%H%M%SZ" and we use "%Y-%m-%d_%H-%M-%S".  Is there a way to format the date differently? Or @alexpiet could we change the date format on our side?

src/foraging_gui/Foraging.py
```diff
@@ -338,7 +354,7 @@ def connectSignalsSlots(self):
         )
         self.ID.returnPressed.connect(
             lambda: self.AutoTrain_dialog.update_auto_train_fields(
-                subject_id=self.ID.text(),
```
alexpiet, line 341: @micahwoodard So you are replacing all calls to `ID.text()` with the schema. I would make sure to check other files. It looks like Dialogs.py calls ID.text() in at least one place.

src/foraging_gui/Foraging.py
bruno-f-cruz: Yep. `allow_dirty` is True, means you intend people to be able to run the code with uncommitted changes and potentially untracked files.

micahwoodard: That line should be a boolean since I'm evaluating if its not an empty string. I thought if the repo was dirty, then set it to True and if not, False. I grabbed the logic for that line from [here ](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/60f8beb5cf259b096e71331217b2a631d1eaeb43/src/foraging_gui/Foraging.py#L4715C5-L4715C40)

src/foraging_gui/Foraging.py
micahwoodard: Ah yes really good point for probably all of the replaced variables

src/foraging_gui/Foraging.py
alexpiet: When did that become the time standard ? I know last year we had a LOT of discussion about time format and settled on "%Y-%m-%d_%H-%M-%S"

Maybe the use case is to override the validator until the session starts and we have a final session name, and then allow the validator to work.

alexpiet: It looks like the other pipelines are all using "%Y-%m-%d_%H-%M-%S" based on code ocean data assets.

src/foraging_gui/Foraging.py
alexpiet: Oh sorry @micahwoodard I missed that you were checking for equality with the empty string. 

So your code will set `allow_dirty` to False if the repo is clean. And `allow_dirty` to True if the repo is dirty. Thats fine, given that we are allowing people to run dirty.

src/foraging_gui/Foraging.py
bruno-f-cruz: I dont think it has become a standard, hence the "proposed". If the current standard is this one: https://aind-data-schema.readthedocs.io/en/latest/data_organization.html. When I checked this I found it odd that the same document uses at least 4 different standards (some of which are in direct conflict with what is written in the document itself, like not being timezone aware).

Additionally, the string does not seem to follow the ISO8061 (or at least not the native python implementation). Such that the following example fails. I would strongly encourage people to standardize formats that allow us to use already existing tools.

```python
import datetime

tz = "2022-04-26T114809"
tz_alex = "2022-04-26_11-48-09"

print(datetime.datetime.fromisoformat(tz))
# 2022-04-26 11:48:09
print(datetime.datetime.fromisoformat(tz_alex))
# Throws error ValueError: Invalid isoformat string: '2022-04-26_11-48-09'
```

alexpiet: I see, well, I think we should keep using the existing standard for consistency with existing data, other pipelines, and AIND metadata. 

(personal aside, I think 11-48-09 is much easier to read than T114809)

src/foraging_gui/Foraging.py
micahwoodard: Checked widgets being referenced outside of Foraging.py and changed ones that were referencing the information and left the ones interacting with actual widget

alexpiet:
@micahwoodard Can you do two things before we merge:
- Test this with autotrain (ask an RA if you dont know how). I'm paranoid that autotrain is going to be effected. Maybe @hanhou can weigh in
- Add more documentation to the PR about when the data model is created, if/when the model is validated and differences from the main schema (time format, for example), what variables that were attributes of the main window are now nested in the data model. I need to make sure code that runs in the processing pipeline is compatible.

micahwoodard:
Yep will do!

src/foraging_gui/Foraging.py
```diff
@@ -3934,6 +3978,16 @@ def _Start(self):
             # stop lick interval calculation
             self.GeneratedTrials.lick_interval_time.stop()  # stop lick interval calculation
 
+            # validate behavior session model and document validation errors if any
+            try:
+                AindBehaviorSessionModel(**self.behavior_session_model.model_dump())
+            except ValidationError as e:
+                logging.error(str(e), extra={'tags': [self.warning_log_tag]})
```
micahwoodard, line 3985: @bruno-f-cruz is this way a good way to validate the model? I thought the model wouldn't allow attributes to be set to variables with incorrect types after initialization, but after some testing, it doesn't seem to be the case?

micahwoodard:
@alexpiet tested with auto train and didn't find any bugs/errors. Also elaborated in the pull request description. Let me know if anything is unclear or unexplained there

alexpiet:
@hanhou Any details to add?

src/foraging_gui/Dialogs.py
```diff
@@ -2335,7 +2335,7 @@ def __init__(self, MainWindow, parent=None):
         self._setup_allbacks()
         
         # Sync selected subject_id
-        self.update_auto_train_fields(subject_id=self.MainWindow.ID.text())
+        self.update_auto_train_fields(subject_id=self.MainWindow.behavior_session_model.subject)
```
hanhou, line 2338: I think this is the only line related to autotrain, and it has been taken care of.

XX-Yin:
I believe this PR is important, but I missed some key background information about how it relates to the downstream pipeline. Could someone provide a brief introduction?

alexpiet:
This PR adopted the AindBehaviorSessionModel from @bruno-f-cruz's Aind.Behavior.Services ecosystem. Adopting this model means that information about the session has been organized into a single point rather than spread out as many variables. This will help us reduce bugs

XX-Yin:
> organized into a single point

How is "organized into a single point" related to the later pipeline and reduce bugs?

alexpiet:
- It ensures that information is encapsulated in one place, and we don't have contradicting information in multiple variables. 
- It makes it easier to update code, since variables are in one place. 
- This was a necessary step to adopt the AIND Behavior Curriculum

PR #1028, opened 2024-11-18
Track and record aind stage positions

### LINKED ISSUE #948: Get AIND stage controller and record
We do this for the newscale stage, but not the AIND stage controller.

micahwoodard:
  
### Describe changes:
- Track and record aind stage positions
- resolves #948 
- Users should be able to set reference points for aind stages
- None
- [x] 446/7


micahwoodard:
- check if B_newscale or B_Stage exists and load either 
- coordinate before merging in for downstream processing

src/foraging_gui/TransferToNWB.py
```diff
@@ -409,9 +409,11 @@ class obj:
...
-                        lickspout_position_x=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[0],
-                        lickspout_position_y=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[1],
-                        lickspout_position_z=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[2],
+                        lickspout_position_x=getattr(obj, 'B_StagePositions')[i]['x'] if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan, 
+                        lickspout_position_y=getattr(obj, 'B_StagePositions')[i].get('y', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
+                        lickspout_position_y1=getattr(obj, 'B_StagePositions')[i].get('y1', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
+                        lickspout_position_y2=getattr(obj, 'B_StagePositions')[i].get('y12', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
+                        lickspout_position_z=getattr(obj, 'B_StagePositions')[i].get('z', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
```
micahwoodard, line 417: @hanhou alex mentioned you would be a good person to ask how these lines work? I'm slightly confused by the _get_field function. Since the B_StagePositions are no longer a list of list coordinates but rather a list of dictionaries, does the approach I wrote work similarly?

micahwoodard:
I've just been testing at 447, and @ellahiltonvano showed me that the aind stage can already upload the last stage position of a mouse ? @alexpiet or @jessy-liao does anyone know where this is already happening? I can't find where it is

micahwoodard:
That is so weird because @jessy-liao also doesn't know. I will try and get video and add to show what is happening

ellahiltonvano:
@micahwoodard to note, it displays the coordinates from the last session, but the stage does not actually move to these coordinates. Also, when you hit enter on the saved X coordinate, for example, all other coordinates reset to 0

alexpiet:
> @micahwoodard to note, it displays the coordinates from the last session, but the stage does not actually move to these coordinates. Also, when you hit enter on the saved X coordinate, for example, all other coordinates reset to 0

Is it the last session ran in the GUI, or the last session from the mouse?

ellahiltonvano:
@alexpiet for that mouse, it usually matches what RAs have manually noted in the notes section of the GUI

alexpiet:
> @alexpiet for that mouse, it usually matches what RAs have manually noted in the notes section of the GUI

@ellahiltonvano I assume this started happening once the stage widget was integrated into the GUI?
@micahwoodard I assume the save/load function iterates through all widgets, and grabs the text from the stage widget.

ellahiltonvano:
@alexpiet correct, this did not save before the widget was integrated

micahwoodard:
> @micahwoodard I assume the save/load function iterates through all widgets, and grabs the text from the stage widget.

Oh okay that makes sense why I couldn't find a direct reference

micahwoodard:
Tested: 
- loading last position in aind-stage and move stage there
- loading last position in newscale-stage and move stage there- old and new code version
- logging both aind and newscale stage looks correct
- reference metadata box still works (although not sure what it is used for)

Still waiting on: 
-@hanhou to help with TransferToNWB.py question

src/foraging_gui/TransferToNWB.py
```diff
@@ -409,9 +409,11 @@ class obj:
                         auto_train_stage_overridden=_get_field(obj, 'TP_auto_train_stage_overridden', index=i, default=np.nan),
 
                         # lickspout position
-                        lickspout_position_x=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[0],
-                        lickspout_position_y=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[1],
-                        lickspout_position_z=_get_field(obj, 'B_NewscalePositions', index=i, default=[np.nan] * 3)[2],
+                        lickspout_position_x=getattr(obj, 'B_StagePositions')[i]['x'] if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan, 
+                        lickspout_position_y=getattr(obj, 'B_StagePositions')[i].get('y', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
+                        lickspout_position_y1=getattr(obj, 'B_StagePositions')[i].get('y1', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
+                        lickspout_position_y2=getattr(obj, 'B_StagePositions')[i].get('y12', np.nan) if len(getattr(obj, 'B_StagePositions', [])) > i else np.nan,
```
alexpiet, line 415: you have a typo here "y12" should be "y2"

src/foraging_gui/TransferToNWB.py
alexpiet: I just took care of this for the NWB packaging, which has to load either format of data.

https://github.com/AllenNeuralDynamics/NWB-Packaging-FiberPhotometry-Base-Capsule/pull/54/files

alexpiet: I don't think we want to save both "y" and "y1" and "y2".

I'm not sure I follow what this code is trying to do.

src/foraging_gui/TransferToNWB.py
```diff
@@ -340,83 +340,100 @@ class obj:
...
+            if stage_positions[i].keys() == ['x', 'y', 'z']:    # newscale stage
+                trial_kwargs['lickspout_position_y'] = stage_positions[i]['y']
+            else:       # aind stage
+                trial_kwargs['lickspout_position_y1'] = stage_positions[i]['y2']
+                trial_kwargs['lickspout_position_y2'] = stage_positions[i]['y2']
+        else:   # if i not valid index, populate values with nan
+            trial_kwargs['lickspout_position_x'] = np.nan
+            trial_kwargs['lickspout_position_y'] = np.nan
+            trial_kwargs['lickspout_position_z'] = np.nan
+
```
micahwoodard: @alexpiet changed so a dictionary of relevant info is created and passed into add_trial. This way, only correct stage axis are added to nwb. This logic block first checks if i is a valid index in B_StagePositions. If the obj does not have a B_StagePositions attribute, the default is empty list so logic block skips to else because i is not a valid index. Otherwise, the lickspout_position keys are created based on the keys in B_StagePositions. The default will always be
 `trial_kwargs['lickspout_position_x'] = np.nan
  trial_kwargs['lickspout_position_y'] = np.nan
  trial_kwargs['lickspout_position_z'] = np.nan`
if B_StagePositions isn't an attribute in obj or the dictionary keys are incorrect. Let me know if that is not a good default or if this is not the right approach

alexpiet: makes sense

alexpiet:
@micahwoodard 
- this branch as conflicts you need to resolves
- I want to be super careful about this PR, because it moves the lickspout stage. Can you find time to test with Ella before we merge?

src/foraging_gui/TransferToNWB.py
micahwoodard: After testing though I'm getting value errors for lickspout_position_y1, lickspout_position_y2 and missing lickspout_position_y. Is this something intrensic to the nwb.add_trial function or is it a schema on our side? I'm not super familiar with nwb files

alexpiet: You just need to modify this line: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/cb95f267f15042ee4efe4ca2bbcae0e09ecffaef/src/foraging_gui/TransferToNWB.py#L275

Which is defining what columns should be added to the trials table, so then when you call "add_trial" its complaining because there is a mis-match.

micahwoodard: ahh great thank you

micahwoodard: Okay now tested and working for json with B_StagePositions with 'x', 'y', 'z' keys and 'x', 'y1', 'y2', 'z' keys. Also working in case that B_StagePositions doesn't exist or has incorrect keys

micahwoodard:
testing with ella at 12:30

micahwoodard:
Video of aind stage moving to coordinates IF load file contains correct stage position configuration. NOTE the stage will NOT move if the gui simply sets the stage widget text boxes to the correct coords. 


https://github.com/user-attachments/assets/da9eddd7-0c29-48d9-abd0-eece72198713

micahwoodard:
Okay test with ella on both types of stages

PR #1049, opened 2024-11-20
LickVisualizer update

hagikent:
Addressing the following issue:
https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/561

As Bonsai doesn't allow you to change the title of windows after compiling/starting, now Lick Visualizer uses `Label` function from `Bonsai.GUI` package and dynamically indicating Box info by fetching it from setting.json file.

Screenshot 2024-11-19 at 4 42 47 PM

No. It just makes easier to tell correspondence between LickVis windows and Box#.

No.

428-9-A / 447-3-C / 447-3-D


hagikent:
@micahwoodard 
sorry I sent a minor commit before signing you out from GH-desktop at the rig, and it seems as if you co-authored...

XX-Yin (approved):

PR #1050, opened 2024-11-20
added in error bars, updated window, and graph by trial number

### LINKED ISSUE #1027: Update Bias indicator
In order to more accurately calculate bias, we need to: 
- change argument n_trial_back to 5 
- change cross validation argument cv to 5
- only input last 200 choice/reward history
- add bootstrap [confidence intervals](https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models/blob/80361bf594fc99ba6d6558c35308550b0264e1a7/src/aind_dynamic_foraging_models/logistic_regression/model.py#L228) to graph

micahwoodard:
  
### Describe changes:
- calculating bias is updated to be for last 200 trials and bias graph updated to show upper and lower CI
- resolves #1027 
- bias should be more accurately calculated 
- None
- [ ] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -164,7 +164,7 @@ class Window(QMainWindow):
         self.threadpool_workertimer=QThreadPool() # for timing
 
         # create bias indicator
-        self.bias_n_size = 500
+        self.bias_n_size = 200
         self.bias_indicator = BiasIndicator(x_range=self.bias_n_size)  # TODO: Where to store bias_threshold parameter? self.Settings?
         self.bias_indicator.biasValue.connect(self.bias_calculated)  # update dashboard value
         self.bias_indicator.setSizePolicy(QSizePolicy.Maximum, QSizePolicy.Maximum)
@@ -4242,25 +4242,23 @@ class Window(QMainWindow):
                 # calculate bias every 10 trials
                 if (GeneratedTrials.B_CurrentTrialN+1) % 10 == 0 and GeneratedTrials.B_CurrentTrialN+1 > 20:
                     # correctly format data for bias indicator
-                    choice_history = [1 if x == 2 else int(x) for x in self.GeneratedTrials.B_AnimalResponseHistory]
-                    any_reward = [any(x) for x in np.column_stack(self.GeneratedTrials.B_RewardedHistory)]
 
-                    # use self.bias_n_size of trials to compute bias over or take n the first 0 to N trials
-                    l = len(choice_history)
-                    # FIXME: .6  of choice history is a little arbitrary. If the n_trial_back is close to equaling the
-                    #  trial count, the logistic regression can't be calculated because of an error saying
-                    #  'Cannot have number of splits n_splits=10 greater than the number of samples: 2'
-                    n_trial_back = self.bias_n_size if l > self.bias_n_size else \
-                        round(len(np.array(choice_history)[~np.isnan(choice_history)])*.6)
+                    formatted_history = [np.nan if x == 2 else int(x) for x in self.GeneratedTrials.B_AnimalResponseHistory]
+                    formatted_reward = [any(x) for x in np.column_stack(self.GeneratedTrials.B_RewardedHistory)]
+
+                    # only take last 200 trials if enough trials have happened
+                    choice_history = formatted_history[-200:] if len(formatted_history) > 200 else formatted_history
+                    any_reward = formatted_reward[-200:] if len(formatted_reward) > 200 else formatted_reward
+
                     # add data to bias_indicator
                     if not self.bias_thread.is_alive():
                         logger.debug('Starting bias thread.')
                         self.bias_thread = threading.Thread(target=self.bias_indicator.calculate_bias,
-                                                       kwargs={'time_point': self.GeneratedTrials.B_TrialStartTime[-1],
+                                                       kwargs={'trial_num': len(formatted_history),
                                                                'choice_history': choice_history,
                                                                'reward_history': np.array(any_reward).astype(int),
-                                                               'n_trial_back': n_trial_back,
-                                                               'cv': 2})
+                                                               'n_trial_back': 5,
+                                                               'cv': 5})
                         self.bias_thread.start()
                     else:
                         logger.debug('Skipping bias calculation as previous is still in progress. ')
--- a/src/foraging_gui/bias_indicator.py
+++ b/src/foraging_gui/bias_indicator.py
@@ -1,4 +1,4 @@
-from pyqtgraph import PlotWidget, GraphItem, setConfigOption, colormap, PlotDataItem, TextItem
+from pyqtgraph import PlotWidget, GraphItem, setConfigOption, colormap, PlotDataItem, TextItem, ErrorBarItem
 from PyQt5.QtCore import pyqtSignal
 from PyQt5.QtGui import QColor, QPen
 from PyQt5.QtWidgets import QMainWindow
@@ -98,7 +98,7 @@ class BiasIndicator(QMainWindow):
         self._x_range = value
 
     def calculate_bias(self,
-                       time_point: float,
+                       trial_num: int,
                        choice_history: Union[List, np.ndarray],
                        reward_history: Union[List, np.ndarray],
                        n_trial_back: int = 15,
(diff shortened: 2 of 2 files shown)


micahwoodard:
Video of bias graph changes

https://github.com/user-attachments/assets/3a52ca4b-8001-4206-acfa-8169556dd207

src/foraging_gui/Foraging.py
```diff
@@ -4242,25 +4242,23 @@ def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
...
                         logger.debug('Starting bias thread.')
                         self.bias_thread = threading.Thread(target=self.bias_indicator.calculate_bias,
-                                                       kwargs={'time_point': self.GeneratedTrials.B_TrialStartTime[-1],
+                                                       kwargs={'trial_num': len(formatted_history),
                                                                'choice_history': choice_history,
                                                                'reward_history': np.array(any_reward).astype(int),
-                                                               'n_trial_back': n_trial_back,
-                                                               'cv': 2})
+                                                               'n_trial_back': 5,
+                                                               'cv': 5})
```
alexpiet, line 4261: we should use cv = 1

src/foraging_gui/bias_indicator.py
```diff
@@ -98,7 +98,7 @@ def x_range(self, value: int) -> None:
         self._x_range = value
 
     def calculate_bias(self,
-                       time_point: float,
+                       trial_num: int,
                        choice_history: Union[List, np.ndarray],
                        reward_history: Union[List, np.ndarray],
                        n_trial_back: int = 15,
```
alexpiet, line 104: set n_trial_back to 5

alexpiet:
- I think we can set the y limits to (-1,1), I'm not sure what they are set at now
- Are the red and blue horizontal lines at +/- .7? I think I remember that being the case. I think thats fine for now. 
- Can you add a faint gray horizontal line at bias=0? That will help users know if the mouse is slightly biased towards left or right
- I think we have been going in circles on the x-axis of this plot. Sorry if I've been communicating poorly. In the video above, the ticks are trial number, but the label says "Trial start time(s)"
   - This is confusing because the ticks are trial number, but the label says "seconds"
   - Its also confusing because the other plot has ticks of seconds from session start. 
   - So I think we should EITHER
       - Make the label say "Trial #", and just live with the fact that the other plot is in time
       - Make the x-axis time relative to session start. Before this PR, we had time, but it wasn't relative to session start, so the units were hard to interpret and a user still couldn't compare to the other plot.

micahwoodard:
> * Are the red and blue horizontal lines at +/- .7? I think I remember that being the case. I think thats fine for now.

Yep exactly. 

> I think we have been going in circles on the x-axis of this plot. Sorry if I've been communicating poorly. In the video above, the ticks are trial number, but the label says "Trial start time(s)"
> * This is confusing because the ticks are trial number, but the label says "seconds"

Yeah I forgot to change the axis title, my bad. Personally, I like plotting bias vs trial number but I am not a user. maybe @ellahiltonvano could weight in if RA's would prefer x axis in time relative to session start or trial number. Alternatively we can use trial number for now and refactor if users would rather it be in time

src/foraging_gui/bias_indicator.py
micahwoodard: The default should be 5? I was using 15 based on the default of the fit_logistic_regression function. Easy and happy to change though

alexpiet: Yeah, based on some investigating I did this week, we should use 5

src/foraging_gui/Foraging.py
micahwoodard: When I use cv=1 I get this error: 
`sklearn.utils._param_validation.InvalidParameterError: The 'cv' parameter of LogisticRegressionCV must be an int in the range [2, inf), an object implementing 'split' and 'get_n_splits', an iterable or None. Got 1 instead.`

I thought on Friday we talked about changing it to 5, but maybe there was a conversation that I missed about something different

micahwoodard:
New bias indicator just dropped! 

https://github.com/user-attachments/assets/a573e7c9-7371-461e-948b-2412d10f6ab2

also @alexpiet I can't figure out how to pin >= version for gthub repos. Let me know if you know how. I was able to pin to v0.11.2

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
     "newscale@git+https://github.com/AllenNeuralDynamics/python-newscale@axes-on-target",
     "aind-auto-train@git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main",
     "aind-slims-api@git+https://github.com/AllenNeuralDynamics/aind-slims-api@main",
-    "aind-dynamic-foraging-models@git+https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models@main",
+    "aind-dynamic-foraging-models@git+https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models@v0.11.2",
     "aind-behavior-services<0.9",
```
alexpiet, line 31: the package is available on pypi, so I think its better to do `aind-dynamic-foraging-models>=0.11.2`, so we aren't pinned to that exact release

alexpiet:
> New bias indicator just dropped!
> 
>  2024-11-21.08-15-08.mp4 
> also @alexpiet I can't figure out how to pin >= version for gthub repos. Let me know if you know how. I was able to pin to v0.11.2
> 
> Also couldn't get cv=1 to work so still using cv=5. Can change to cv=2 maybe?

No, we should absolutely use cv=1. Which depends on v 0.11.2. I mentioned this in a comment, but repeating here. The package is on pypi, so we can use `aind-dynamic-foraging-models>=0.11.2`

You should also set fit_exponential=False when you call logistic regression (https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models/blob/e094eb36fab34eefc861ceacdfee29b026cb7862/src/aind_dynamic_foraging_models/logistic_regression/model.py#L124)

src/foraging_gui/Foraging.py
alexpiet: Are you sure you are on the v0.11.2?

micahwoodard:
> Huh I was also having issues using aind-dynamic-foraging-models>=0.11.2. Will investigate further

Oh actually ever since installing the stage-widget my pip install has been a little wonky that's probably what it is. Will test on a rig instead

alexpiet:
This probably got lost on teams. But you should also implement logic like I did here (https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-basic-analysis/blob/44eefc5099309207ffaa4c9efac126a211fc490b/src/aind_dynamic_foraging_basic_analysis/metrics/trial_metrics.py#L146)

The idea is that we need to pro-actively check to see if we have all left, or all right, or all ignore choices, and handle those edge cases.

alexpiet:
Making a note based on in person conversations that the confidence intervals look like they are plotted wrong on the last demo

micahwoodard:
Slightly worse video but now CI is shaded grey area

https://github.com/user-attachments/assets/78e03398-b6f5-4ccf-b8a8-6c02fbc7de15

src/foraging_gui/bias_indicator.py
```diff
@@ -129,68 +144,99 @@ def calculate_bias(self,
...
+
+            else:   # skip bias calculation if no conditions are met
+                return
+
+            # update
+            self._upper_scatter_item.setData(x=np.append(self._upper_scatter_item.xData, trial_num),
+                                             y=np.append(self._upper_scatter_item.yData, bias+upper))
+            self._lower_scatter_item.setData(x=np.append(self._lower_scatter_item.xData, trial_num),
+                                             y=np.append(self._lower_scatter_item.yData, bias+lower))
+
```
micahwoodard, line 190: @alexpiet I'm calculating the upper and lower ci coords with logic: 
upper_coord = bias + upper 
lower_coord = bias + lower (since lower is always negative)

I'm pretty sure that's correct but let me know if I'm missing an edge case or something

alexpiet: No, the upper_coord is just 'upper'. and the lower_coord is just 'lower'. You don't need to add the 'bias' in

micahwoodard: Ohh okay great that is why sometimes they were off. Thanks for clarifying!

alexpiet (approved):

Issue #510, opened 2024-06-12
Metadata errors to fix


alexpiet:
## Rig
- [x] Rig ID uses YYYYMMDD, not YYYY_MM_DD
- [x] Tube manufacturer should be CUSTOM or OTHER, not null
- [x] Motorized stage needs a model name
- [x] Reward spouts needs a manufacturer. CUSTOM, OTHER, JANELIA
- [x] Solenoid - manufacturer, model, serial number?
- [x] How is there a lick_sensor_type without a lick_sensor?
- [x] Isn’t the lick_sensor the janelia lick detector rather than the spout?
- [x] Speaker Manufacturer, model, serial number? 
   - [x] Need info
- [x] Camera lens: 
   - [x] How is there a model number without a manufacturer? 
   - [x] Do we need to change the focal length field if it doesn’t work for the lens?
- [ ] Add position of cameras
- [ ] Add enclosure
- [ ] Harp daq: 
   - [ ] core_version for each harp device, should get dynamically from the devices. I hard-coded the value used in 447. 
- [x] Calibrations: typo in lack spout right description
- [x] Beamsplitter:
    - [x] Confirm height of beamsplitter (24 or 25)
    - [x] This is more a choice thing, but if the notes contain “493/574 nm BrightLine dual-edge standard epi-fluorescence dichroic beamsplitter" then I would use a name like “beamsplitter” rather than duplicating half that information in the name. If these are details we want to track, let’s get it into the schema.

### Session
- [x] Daq device names don’t match the names in the rig
- [x] Dynamic foraging task software should be in the stimulus_epoch section
    - there are different stimulus_epoch (auditory or optogenetics). Which one to store the software? 
- [x] Speaker name doesn’t match the name in rig
- [ ] How was there 343 uL water consumed in the session and none in the stimulus epoch?
    - I think it makes more sense to store the water in the session instead of a specific stimulus (we have different stimulus auditory tone occurs every trial, optogenetics occurs in some trials...). 
- [ ] You list several patches, only one is listed in the Rig (and not with this name)
- [ ] These are surgically implanted fibers I assume?
- [ ] Add Behavior parameters in the stimulus software's parameters fields in the stimulus epoch.
- [ ] https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/465

alexpiet:
@XX-Yin Can you take care of the Session metadata updates?

XX-Yin:
[...]

PR #1051, opened 2024-11-21
Update metadata

XX-Yin:
  
### Describe changes:
- https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1051
- [x] "Lick sensor left" and "Lick sensor right" are not daq_names for your rig and thus can't be listed as under `daq_names`. 
      (Removed the behavior from the data stream and added the behavior to the stimulus epoch instead. And no daq names required.)
- [x] Camera names don't match the names in the rig. They must match. There is a utility function you can use to make sure that the names align between rig and session.
- [x] Reward delivery cannot be null. How are you delivery rewards without it?
 (We already have the code to generate the reward delivery, and we are waiting for the SOP to collect the reference position and reference area. )
- [x] Your `session_end_time` is before the end of one of your data streams. The end time should be the last time of all data streams and stimulus epochs.
- [x] Missing `stimulus_device_name`
- [x] These output parameters and task parameters duplicate almost the entire metadata schema. Can we please please please not do this? Very seriously, this is going to make our DocDB painful and it ultimately undermines having the schema in the first place.
- [x] eg. calibrations should be in the calibration field
- [x] frame rate should be in the device (also, check the version of the schema you're using. The cameras in your device class have a `max_frame_rate` which was changed a while ago)
 (The max_frame_rate has been removed and replaced by frame_rate)
- [x] Your `frame_num` item specifically is blocking upload to DocDB. You can't have fields with ".csv" in the name. This information should live in the quality_control class (which was only recently added to the schema, but that's where it should live) (These fields should go to the qc and they are removed here)
- [x] adding software name to the stimulus_epochs
- [x] mismatch harp names
resolved https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/510
No
No
On my own computer


src/foraging_gui/GenerateMetadata.py
```diff
@@ -152,17 +152,19 @@ def _mapper(self):
...
             },
-            'ephys_rig_behavior_daq_names':[
+            'rig_daq_names_janelia_detector':[
                 "Harp Behavior",
                 "Harp Sound",
                 "Harp clock synchronization board",
                 "Harp sound amplifier",
                 ],
-            'behavior_rig_behavior_daq_names':[
+            'rig_daq_names_new_detector':[
```
alexpiet, line 161: Can we call it "aind_detector" and not "new" since that will get confusing as time goes one

alexpiet:
Other than the name issue, this seems good to me. Can you share a new rig.json and session.json for approval from Saskia?

Are there any other issues left to implement, or is this ready to go?

src/foraging_gui/GenerateMetadata.py
XX-Yin: updated!

XX-Yin:
> Other than the name issue, this seems good to me. Can you share a new rig.json and session.json for approval from Saskia?
> 
> Are there any other issues left to implement, or is this ready to go?

It's ready for review. I will generate a new session and rig metadata tomorrow for review

src/foraging_gui/GenerateMetadata.py
```diff
@@ -1017,7 +1123,12 @@ def _get_behavior_software(self):
             version=f'behavior branch:{self.Obj["current_branch"]}   commit ID:{self.Obj["commit_ID"]}    version:{self.Obj["version"]}; metadata branch: {current_branch}   commit ID:{commit_ID}   version:{version}',
             url=self.Obj["repo_url"],
         ))
```
alexpiet, line 1125: Whats going on here? It seems like we are adding Software info to "behavior_software" and "behavior_software_origin"

alexpiet:
@XX-Yin Can you add details to the PR about what issues were fixed in the metadata? Probably just copying the itemized list from the issue is sufficient.

src/foraging_gui/GenerateMetadata.py
XX-Yin: The behavior_software contains both the software version for running the behavior and the software version for generating the session metadata. The behavior_software_origin only contains the version for running the behavior. I have removed the behavior_software_origin for simplicity.

alexpiet (approved):
This seems like an improvement over the current metadata, so even if its not perfect, lets merge it.

PR #1088, opened 2024-11-25
Only logging error for fip if mouse training stage is at certain point

### LINKED ISSUE #1045: Update FIP schedule check to use "First FP Stage"
If the schedule shows:
FP		Axon	FINAL

Then don't send an alert for starting the session without FIP unless the mouse is on "FINAL"

micahwoodard:
  
### Describe changes:
- check schedule for what stage mouse will start fip and only log errors for fip if mouse training stage is at certain point
- resolves #1045 
- None
- None
- [x] 446

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3797,20 +3797,22 @@ class Window(QMainWindow):
             logging.info('Start button pressed: starting trial loop')
             self.keyPressEvent()
 
-            # check if FIP setting match schedule
+            # check if FIP setting match schedule. skip if test mouse or mouse isn't in schedule or
             mouse_id = self.behavior_session_model.subject
-            if hasattr(self, 'schedule') and mouse_id in self.schedule['Mouse ID'].values and mouse_id not in ['0','1','2','3','4','5','6','7','8','9','10'] : # skip if test mouse or mouse isn't in schedule or
-                FIP_Mode = self._GetInfoFromSchedule(mouse_id, 'FIP Mode')
-                FIP_is_nan = (isinstance(FIP_Mode, float) and math.isnan(FIP_Mode)) or FIP_Mode is None
+            if hasattr(self, 'schedule') and mouse_id in self.schedule['Mouse ID'].values and \
+                    mouse_id not in ['0','1','2','3','4','5','6','7','8','9','10'] :
+                fip_mode = self._GetInfoFromSchedule(mouse_id, 'FIP Mode')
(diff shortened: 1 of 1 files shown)


micahwoodard:
Functionality should be if First FIP Stage is greater than current stage, no warning will appear. If First FIP Stage is undefined and FIP mode is specified, warning will appear. If autotrain has no specified stage and FIP mode is specified, warning will appear regardless of First FIP Stage.

alexpiet (approved):
This works. Its a little bit of a brittle solution since it:
(1) will break if the number of stages changes, or the names of the stages change
(2) depends on a convention of how information is stored in the schedule.

@hanhou @ellahiltonvano any input?

ellahiltonvano:
@alexpiet (theoretically) we agreed to have the schedule naming convention match autotrain. I think that is now happening in all cases. 

I do think this will be helpful, there have been mice run under the incorrect curriculum. But I agree that it will likely need to be updated as we iterate through curriculum changes.

micahwoodard:
> @micahwoodard if first FP stage is unspecified this means RAs start FIP at stage 1, fyi

in the streamlit app or just in the schedule? 

> @micahwoodard if first FP stage is unspecified this means RAs start FIP at stage 1, fyi. I also think it would make sense to pull training stage information from Streamlit, rather than relying on RAs to update the schedule (this gets missed a fair bit)
 
The training stage info is being pulled from the AutoTrain_dialog which I believe is relying on the Streamlit app, not on the schedule.

ellahiltonvano:
@micahwoodard on the schedule

micahwoodard:
> @micahwoodard on the schedule

Oh sorry I read your comment wrong that makes total sense and the logic should follow that

micahwoodard:
> This works. Its a little bit of a brittle solution since it: (1) will break if the number of stages changes, or the names of the stages change (2) depends on a convention of how information is stored in the schedule.
> 
> @hanhou @ellahiltonvano any input?

@alexpiet any thoughts on how to make it less hardcoded? It would probably be a good idea to put in some checks to make sure nothing breaks if the schedule has an unexpected value. I could also populate the stages list programmatically if there is anywhere that stores that info.

micahwoodard:
@alexpiet do we want to merge this in for this week or wait for a more flexible solution

alexpiet:
@micahwoodard Sorry I didn't respond promptly. I would merge it in. As we adopt data models and the new training curriculum manager we can make this more flexible

PR #1098, opened 2024-11-26
Update ignore trials stopping logic

micahwoodard:
  
### Describe changes:
- add two training parameters auto_stop_ignore_win (default = 30) and auto_stop_ignore_ratio_threshold (default = 0.8) 
- after each trial, if session_time > 30 min and if  percentage of ignored on-autowater trials exceeds auto_stop_ignore_ratio_threshold, automatically stop the session
- resolves [#680](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/680)
- None
- None
- [x] 446

### DIFF
--- a/src/foraging_gui/MyFunctions.py
+++ b/src/foraging_gui/MyFunctions.py
@@ -1170,7 +1170,8 @@ class GenerateTrials():
         # Check for reasons to stop early 
         non_auto_reward = self.B_AnimalResponseHistory[np.where(~self.TP_AutoReward)]   # isolate non-auto-reward trials
         win_sz = self.TP_auto_stop_ignore_win
-        if non_auto_reward.shape[0] > win_sz and non_auto_reward[-win_sz:][np.where(2)].shape[0] >= StopIgnore:
+        if self.BS_CurrentRunningTime >= 30 and \
+                non_auto_reward.shape[0] > win_sz and non_auto_reward[-win_sz:][np.where(2)].shape[0] >= StopIgnore:
             stop=True
             msg = f'Stopping the session because the mouse has ignored at least ' \
                   f'{self.TP_auto_stop_ignore_ratio_threshold*100}% of {self.TP_auto_stop_ignore_win} ' \
--- a/src/foraging_gui/MyFunctions.py
+++ b/src/foraging_gui/MyFunctions.py
@@ -1154,7 +1154,7 @@ class GenerateTrials():
      
     def _CheckStop(self):
         '''Stop if there are many ingoral trials or if the maximam trial is exceeded MaxTrial'''
-        StopIgnore=int(self.TP_StopIgnores)-1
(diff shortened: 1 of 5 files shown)


micahwoodard:
Probably buggy since I don't have a rig to test on but get's the general idea across.

src/foraging_gui/MyFunctions.py
```diff
@@ -1166,26 +1166,26 @@ def _CheckStop(self):
         stop = False
         msg =''
         warning_label_text = ''
-        warning_label_color = 'color: gray;'        
 
         # Check for reasons to stop early 
-        if (np.shape(self.B_AnimalResponseHistory)[0]>=StopIgnore) and (np.all(self.B_AnimalResponseHistory[-StopIgnore:]==2)):
+        non_auto_reward = self.B_AnimalResponseHistory[np.where(~self.TP_AutoReward)]   # isolate non-auto-reward trials
+        win_sz = self.TP_auto_stop_ignore_win
+        if self.BS_CurrentRunningTime >= 30 and \
```
micahwoodard, line 1173: @hanhou is it okay that the session time check is hard coded to 30? Would there be any reason to make it a training parameter?

hanhou: Making it as a parameter would be great

micahwoodard:
Also just as a note to self, need to check formating on ephys ui since I can't see if full screen on my laptop

micahwoodard:
Here is how the ephys gui will look like with the new parameters: 
image

micahwoodard:
This is what the other gui looks like 
image

XX-Yin:
@micahwoodard The logic looks good! Have you tested it in a session?

hanhou:
Looks good to me too! Just a minor point: in the GUI, it might be better to put the "auto stop ignore window" field next to "auto stop ignore ratio threshold".

hanhou (approved):

micahwoodard:
> @micahwoodard The logic looks good! Have you tested it in a session?

Yep just tested this morning

micahwoodard:
> Looks good to me too! Just a minor point: in the GUI, it might be better to put the "auto stop ignore window" field next to "auto stop ignore ratio threshold".

And will do easy fix

XX-Yin (approved):

micahwoodard:
@alexpiet @hanhou is there any concern that this will break post processing anywhere? Is B_StopIgnores being used anywhere?

alexpiet:
I don't think so. 

FYI, this is the repo that does the downstream processing: https://github.com/AllenNeuralDynamics/aind-fip-nwb-base-capsule/tree/main

We should probably develop a better system such that changes to parameters in the GUI are automatically reflected in this repo, instead of having to manually make sure they are compatible. Probably something to adopt with the AIND Session/Rig/Task model

PR #1115, opened 2024-12-04
Add version compatibility ranges for dependencies

galenlynch:
### Describe changes:
Updated the pyrpoject.toml to specify dependency compatibility ranges, instead of not specifying any version range. If the compatibility range is not specified, incompatible versions can be installed and used, causing run-time errors like the one I ran into with python-harp.

Incompatible versions of dependencies being used when installing this project.

Installing the project allows you to run the project without errors.

None

No, reinstall would be needed


alexpiet:
What does `~=` do?

galenlynch:
`~=` is already in the pyproject.toml and is specified by [PEP440](https://peps.python.org/pep-0440/#compatible-release) and [PyPA](https://packaging.python.org/en/latest/specifications/version-specifiers/#id5). However, it's not working for me locally so I wrote out the range explicitly instead. 

I just successfully installed the python project using the new pyproject.toml on ephys rig 2, and it has the "correct" dependency versions (i.e. what was installed previously when I did `python -m pip install -U -e '.'` and then successfully ran a behavior session without errors or crashes)

pyproject.toml
```diff
@@ -17,31 +17,30 @@ readme = "README.md"
...
-    "openpyxl",
-    "deepdiff",
+    "pynwb >=2, <3",
+    "requests >=2, <3",
+    "harp-python >=0.3, <2",
+    "openpyxl >=3, <4",
+    "deepdiff >=8, <9",
     "aind-data-schema==1.1.0",
     "aind-data-schema-models==0.5.6",
-    "pydantic==2.9.2",
```
alexpiet, line 38: Can you keep the versions that were already set for pydantic, python-logging-loki, and pykeepass?

requirements.txt
```diff
@@ -1,27 +0,0 @@
-
-numpy
```
alexpiet, line 2: Why remove the requirements.txt file?

alexpiet:
Can you test this in 446/447 by doing the following:
```
# switch to your branch
conda activate Foraging
pip install --upgrade .
```
The output should say successful installation, and then run a test session. Obviously please click the `switch to main` bat file on the desktop when you are done

galenlynch:
I did this on ephys 2.

On Wed, Dec 4, 2024, at 08:37, Alex Piet wrote:
> 
> 
> Can you test this in 446/447 by doing the following:
> 
> `# switch to your branch
> conda activate Foraging
> pip install --upgrade .
`

galenlynch:
I can test in 446 though

On Wed, Dec 4, 2024, at 08:44, Galen Lynch wrote:
> I did this on ephys 2.
> 
> On Wed, Dec 4, 2024, at 08:37, Alex Piet wrote:
>> 
>> 
>> Can you test this in 446/447 by doing the following:
>> 
>> `# switch to your branch
>> conda activate Foraging
>> pip install --upgrade .
`

pyproject.toml
galenlynch: Why do you want to pin at a particular version? Pedantic likely follows semantic versioning, presumably any version 2.9.2 or higher in the same compatible range (i.e. `[2.9.2, 3)`) should work.

requirements.txt
galenlynch: Why have two sources of truth that are copies of each other? The dependencies of dynamic-foraging should be described in the pyproject.toml. pip can (and does) use the toml, not requirements.txt to build the environment. Is it for conda? Conda just uses pip (and therefore pyproject.toml) if you do `pip install --upgrade`.

pyproject.toml
galenlynch: Basically anything in that range should still have the features you want (presumably something happened in pydantic exactly at 2.9.2 that this project depends on) while still maintaining a backwards-compatible API, and you will get bug fixes etc. for free. Unless they are not adhering to semantic versioning, and you know that 2.9.3 breaks your code (!) then I think it's bad practice to pin a specific version. It just makes the project unnecessarily difficult to install. I understand that many aind packages (e.g. data schema) do not adhere to semantic versioning and break backwards compatibility in compatible version ranges, so I have left those pinned at a single version.

XX-Yin (approved):

galenlynch:
Blocked by #1118, will address in a separate PR.

pyproject.toml
alexpiet: The versions of python-logging-loki and pykeepass were added by Ross recently, and I assume he knows what he is doing. (I guess I dont really understand the difference between ~= and <, > pairs)

you are right that we just need greater than 2.9.2 for pydantic.

requirements.txt
alexpiet: I guess I'm not sure what the requirements.txt file was for. I thought (incorrectly) that the https://github.com/AllenNeuralDynamics/aind-library-template had both a requirements.txt and pyproject.toml. I didn't make the requirements.txt file, so I'm unsure about any unintended consequences of deleting it. 

That being said, I think we can go ahead and delete it

pyproject.toml
galenlynch: `~=2.9.2` is equivalent to `>=2.9.2, <3`, but since I was not able to use `~=` locally I changed everything to use the more verbose but equivalent format. Semantically, I did not change the compat range for python-logging-loki or pykeepass, I only changed the syntax.

requirements.txt
galenlynch: I think it may have been used to set up a conda environment. I think it's redundant at the moment, and is a source of potential confusion.

galenlynch:
Confirmed that this works in 446-6A, with #1119 additionally applied

alexpiet (approved):

Issue #1114, opened 2024-12-03
Upload manifest issues


alexpiet:
It appears there are multiple issues related to data upload. This appears to only being happening on the production_testing branch
- Saved data is stored at `behavior_behavior_<mouse_id>_<date time>` instead of `behavior_<mouse_id>_<date time>`
- The behavior modality is not accurately being added to the upload manifest

alexpiet:
I suspect the `behavior_behavior_` issue is something to do with https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1003. @micahwoodard Can you take a look?

I suspect the behavior modality not being added is due to https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1051 causing a breaking change. @XX-Yin Can you take a look?

alexpiet:
To be explicit about the behavior modality: the upload manifest code checks for `Modality.BEHAVIOR in stream.stream_modalities`, yet in sessions generated by the testing branch, there is no `Modality.BEHAVIOR` in stream.stream_modalities.

@XX-Yin Is this intentional? If so, we need to modify the upload manifest code. If not, we need to fix how we make the session.json

XX-Yin:
It's intentional to remove the behavior from the stream per Saskia's suggestion. This is one error I found.

XX-Yin:
@alexpiet I will make PR to update the manifest code

PR #1116, opened 2024-12-04
Update manifest

XX-Yin:
  
### Describe changes:
1. Update the detection of stream modality.
2. Save the manifest file every trial. 

1. resolved: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1114#issuecomment-2515779489

2. The manifest file should be updated every trial as new data stream may be added after start of the session. 

No

No

tested on 323_Ephys3


src/foraging_gui/Foraging.py
```diff
@@ -223,9 +222,6 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         # Initializes session log handler as None
         self.session_log_handler = None
 
-        # generate an upload manifest when a session has been produced
-        self.upload_manifest_slot = self.sessionGenerated.connect(self._generate_upload_manifest)
```
alexpiet, line 227: Why remove the signal/slot architecture Micah added?

XX-Yin: I guess that structure is used for running the "_generate_upload_manifest" only once, which could be problematic (e.g. when we start the video recording after the first trial, the video modality will not be recorded.)

src/foraging_gui/Foraging.py
```diff
@@ -4628,7 +4625,7 @@ def _open_mouse_on_streamlit(self):
                          '&session_plot_selected_draw_types=1.+Choice+history'
         )
 
-    def _generate_upload_manifest(self, session: Session):
+    def _generate_upload_manifest(self, generated_medadata):
```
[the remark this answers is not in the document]
XX-Yin: It's just safer to check the modality and generate the manifest file each trial, which is not computation intensive.

alexpiet:
Can you make a simplified PR that just adds the correct modality to the manifest? I don't think we should change the way the signal/slot architecture is used, or update the manifest on every trial. Especially without consulting @micahwoodard and the incorrect manifest is going to effect all the sessions on monday

XX-Yin:
> Can you make a simplified PR that just adds the correct modality to the manifest? I don't think we should change the way the signal/slot architecture is used, or update the manifest on every trial. Especially without consulting @micahwoodard and the incorrect manifest is going to effect all the sessions on monday

We can wait for the feedback from @micahwoodard. This PR doesn't change the manifest file. It just updates the manifest file every trial.

alexpiet:
Well, no. The `develop` branch has a bug that will cause all data to fail to upload: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1114

If we don't fix it this week then we can't merge the updates (3rd week in a row...)

XX-Yin:
So, we should review this PR ASAP. If you still have concerns after I tested it, you can also test it on your own computer for double check.

alexpiet:
I'm asking you to remove everything that isn't essential to fixing the bug

XX-Yin:
> I'm asking you to remove everything that isn't essential to fixing the bug

Please feel free to make commits to this PR to remove the code you think that is not essential. They all look essential to me.

XX-Yin:
Then let's merge this PR at the end of this week!

alexpiet:
Ok, I just made a new PR with only the essential fixes(https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1122). We can close this PR

alexpiet:
Ok, you are right that if you run a session with the camera, and then start a new session without the GUI, then it incorrectly lists behavior-video as a modality in the manifest and in the session metadata. I'll make a separate issue, since I don't have time to fix that right now. Lets merge this fix in, since it will effect every session

PR #1117, opened 2024-12-04
Add running wheel

XX-Yin:
  
### Describe changes:
1: Added running wheel optional to the bonsai workflow. 
image

A running-wheel subfolder will be created under the session folder if there is a running wheel detected. 
Add/Set `RunningWheel` to `0` in rigs without the running wheel

Tested in 323_Ephys rig 1 and 3


alexpiet:
Is the three camera view always there, even if there isn't a third camera? Thats going to be really annoying in the training room where screen space is at a premium

alakunina:
Can we have the 2 vs 3 cameras be a setting? The third camera view is kind of important for the experiments on the running wheel and it's annoying to pull up each time.

XX-Yin:
@alakunina It's more difficult to customize the bonsai layout based on settings in each rig. 

The visualization is different in different ephys rigs. @alexpiet I'm wondering if we could specifically exclude the bonsai layout file from each day update? However, issue here is we may need to ask people to manually update the layout file when we made changes to it, which does not seem to be a good idea.  

Or we can pull up the code once a week?

alexpiet:
@bruno-f-cruz We want to be able to add a setting that will toggle whether three cameras are visualized, or just two. Can you show @XX-Yin how to set that up?

alakunina:
The metadata script I made for rig 1 includes it, I can share with you guys
if need be.

XX-Yin:
The session metadata will be updated automatically if the rig metadata is provided.

XX-Yin:
We should merge this PR first, and then find a flexible way to set the visualization panel based on settings.

ellahiltonvano:
Just to weigh in for 446/7 perspective, adding a third camera is not ideal. We already need to visualize the LickVis, Camera recording, FIP, and GUI on half of a monitor - expanding the camera window in these rooms will be difficult to deal with organizationally. 

If we can have a toggle to always show 2 panels in 446/7 that's fine, but I don't think pushing this update to 24 rigs without that option is ideal.

XX-Yin:
> Just to weigh in for 446/7 perspective, adding a third camera is not ideal. We already need to visualize the LickVis, Camera recording, FIP, and GUI on half of a monitor - expanding the camera window in these rooms will be difficult to deal with organizationally.
> 
> If we can have a toggle to always show 2 panels in 446/7 that's fine, but I don't think pushing this update to 24 rigs without that option is ideal.

Yes, it will influence a lot of behavior rigs. Let's remove the body camera visualization here. Then, adding it after we have a flexible solution.

alakunina:
Sounds good! I can cope with bringing up the body can manually in the
meantime.

bruno-f-cruz:
> @bruno-f-cruz We want to be able to add a setting that will toggle whether three cameras are visualized, or just two. Can you show @XX-Yin how to set that up?

If the cameras are hardcoded, the best you can do without a substantial amount of work is to have 2 layout files that are passed via the command line interface (https://bonsai-rx.org/docs/articles/cli.html#specifying-layout-files-in-application-mode) and probably ensure that the application does not try to connect with the non-existing camera (see https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/190#issuecomment-2012901709).

If a deeper code refactoring is on the table, I would probably suggest adopting a procedural generation pattern that works with any number of connected cameras. See the the [Aind.Behavior.VideoEncodingBenchmarks](https://github.com/AllenNeuralDynamics/Aind.Behavior.VideoEncodingBenchmarks/blob/main/src/main.bonsai) repository for an example on how to implement this.

XX-Yin:
> > @bruno-f-cruz We want to be able to add a setting that will toggle whether three cameras are visualized, or just two. Can you show @XX-Yin how to set that up?
> 
> If the cameras are hardcoded, I think the best you can do without a substantial amount of work is to have 2 layout files that are passed via the command line interface (https://bonsai-rx.org/docs/articles/cli.html#specifying-layout-files-in-application-mode) and probably ensure that the application does not try to connect with the non-existing camera (see [#190 (comment)](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/190#issuecomment-2012901709)).
> 
> If a deeper code refactoring is on the table, I would probably suggest adopting a procedural generation pattern that works with any number of connected cameras. See the the [Aind.Behavior.VideoEncodingBenchmarks](https://github.com/AllenNeuralDynamics/Aind.Behavior.VideoEncodingBenchmarks/blob/main/src/main.bonsai) repository for an example on how to implement this.

@bruno-f-cruz Thanks! The second approach seems to be a good solution for us.

XX-Yin:
Can you approve this?

alexpiet:
@XX-Yin Did you update this to not automatically open the body camera? Can you update the screenshots in the main PR?

XX-Yin:
@alexpiet The does not open the body camera automatically. Let's address that issue separately. I updated the screenshot

alexpiet:
@XX-Yin 

I'm confused because the PR still says this "The camera window will have three panels."

XX-Yin:
@alexpiet 
Oh, sorry. I didn't update the visualization window in this PR. I have deleted that sentence.

XX-Yin:
I made a separate issue about the [flexible camera visualization layout](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1171), which requires more efforts and testing.

alexpiet:
@XX-Yin Can you add this to all the rigs in 446/447?

> Add/Set RunningWheel to 0 in rigs without the running wheel

alexpiet (approved):

PR #1122, opened 2024-12-05
Fix behavior modality in upload manifest

### LINKED ISSUE #1114: Upload manifest issues

### LINKED ISSUE #1123: Upload manifest/session metadata does not register behavior video when video is started through FIP workflow

alexpiet:
  
### Describe changes:
Always adds the behavior modality, instead of checking stream_modalities which had behavior removed as a modality

- resolves #1114 
- resolves #1123

- none

- none

- [x] tested in 447

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4663,10 +4663,9 @@ class Window(QMainWindow):
             mount = 'FIP'
 
             modalities = {}
+            modalities['behavior'] = [self.behavior_session_model.root_path.replace('\\', '/')]
             for stream in session.data_streams:
-                if Modality.BEHAVIOR in stream.stream_modalities:
-                    modalities['behavior'] = [self.behavior_session_model.root_path.replace('\\', '/')]
-                elif Modality.FIB in stream.stream_modalities:
+                if Modality.FIB in stream.stream_modalities:
                     modalities['fib'] = [self.PhotometryFolder.replace('\\', '/')]
                 elif Modality.BEHAVIOR_VIDEOS in stream.stream_modalities:
                     modalities['behavior-videos'] = [self.VideoFolder.replace('\\', '/')]
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2827,7 +2827,7 @@ class Window(QMainWindow):
         session_name = self.behavior_session_model.session_name=f'behavior_{self.behavior_session_model.subject}_' \
                                                      f'{self.behavior_session_model.date.strftime("%Y-%m-%d_%H-%M-%S")}'
         self.SessionFolder=os.path.join(self.default_saveFolder,
-            self.current_box,self.behavior_session_model.subject, f'behavior_{session_name}')
+            self.current_box,self.behavior_session_model.subject, session_name)
         self.behavior_session_model.root_path=os.path.join(self.SessionFolder,'behavior')
         self.SaveFileMat=os.path.join(self.behavior_session_model.root_path,f'{session_name}.mat')
         self.SaveFileJson=os.path.join(self.behavior_session_model.root_path,f'{session_name}.json')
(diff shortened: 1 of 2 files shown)


alexpiet:
@XX-Yin This is ready for review, and replaces #1116

XX-Yin:
@alexpiet 
The way to check the modality has two issues:
1) The video modality will not be added to the manifest file as the video streams will only be added to the session streams when there is an end time detected. 
2) The current code would only generate manifest file in the first trial and could miss modality started after first trial. 

I know that you thought the point 2 is not an issue. Please make changes to the point 1 to make the modality check more robust. I suggest we directly check the folder empty for each modality instead of using the session streams.

alexpiet:
Why is the video only added to the stream modalities when there is an end time, but the FIP is added before there is an end time?

XX-Yin:
It's a history issue. You can also change the video modality to be like the FIP

            if hasattr(self, 'fiber_photometry_start_time'):
                Obj['fiber_photometry_start_time'] = self.fiber_photometry_start_time
                if hasattr(self, 'fiber_photometry_end_time'):
                    end_time = self.fiber_photometry_end_time
                else:
                    end_time = str(datetime.now())
                Obj['fiber_photometry_end_time'] = end_time

alexpiet:
@XX-Yin what about this update

XX-Yin:
> @XX-Yin what about this update

This works. 

1) Can you also remove the `camera_start_time` and `fiber_photometry_start_time` when starting a session(like I did in https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1116/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR4052)?  Otherwise, it may interact with the next session. 
2) And don't delete the `fiber_photometry_start_time` at here https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1116/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aL3628. Otherwise, it may not generate the fiber photometry data stream after the new session button is clicked (e.g. after saving). 
3) Can you also make changes in the fiber photometry data stream in the GeneratedMetadata.py to keep consistent with the behavior video data stream?

alexpiet:
Those seem like separate issues from the upload manifest modality.

XX-Yin:
At least the first one is directly associated the manifest file as the code is using the Session Metadata to get the modalities.

alexpiet:
But I'm not setting the start time for the camera or photometry in either the metadata or the window object.

XX-Yin:
If the previous session has the `camera_start_time` and `camera_end_time` and the current session doesn't have the video stream and we don't restart the GUI, the video stream will be added to the current session. I think the root issue is from the session metadata, but it will also influence the modality detection in the manifest file.

alexpiet:
I just tried it, and that does not happen

alexpiet:
Although that may be a result of this bug I just found: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1123

XX-Yin:
> I just tried it, and that does not happen

Did you click the start recording (not the preview)?

alexpiet:
No, I recorded video through the FIP workflow. I just fixed #1123. I will check in 446 and just record video

alexpiet:
Commented on the wrong PR:

Ok, you are right that if you run a session with the camera, and then start a new session without the GUI, then it incorrectly lists behavior-video as a modality in the manifest and in the session metadata. I'll make a separate issue, since I don't have time to fix that right now. Lets merge this fix in, since it will effect every session
here is the issue: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1124

alexpiet:
@XX-Yin do you have any remaining concerns?

XX-Yin:
@alexpiet We can first merge this and fix other issues

XX-Yin (approved):

PR #1156, opened 2024-12-11
Incorrect Waterlog info fix

### LINKED ISSUE #1155: Incorrect Waterlog info
Something appears to be wonky with SLIMS and the GUI. it looks like for both of alex's mice, what should be the post session weight being entered is actually the baseline weight: 
image
WT entry is 29.2 (baseline weight) rather than ~23g (fluctuating depending on day).
Currently the gui is logging waterlogs with this logic
image
weight_g should be the post session weight

micahwoodard:
  
### Describe changes:
- waterlog now record weight_g as animal_weight_post and records operator of session
- resolves #1155 
- None
- None
- [ ] 446/7

### DIFF
--- a/tash
+++ b/tash
@@ -1,0 +1,31 @@
+[1mdiff --git a/src/foraging_gui/Foraging.py b/src/foraging_gui/Foraging.py[m
+[1mindex 122706fe..b9a37510 100644[m
+[1m--- a/src/foraging_gui/Foraging.py[m
+[1m+++ b/src/foraging_gui/Foraging.py[m
+[36m@@ -9,7 +9,7 @@[m [mimport math[m
+ import logging[m
+ from hashlib import md5[m
+ [m
+[31m-import logging_loki[m
+[32m+[m[32m#import logging_loki[m
+ import socket[m
+ import harp[m
+ import threading[m
+[36m@@ -25,7 +25,7 @@[m [mfrom aind_slims_api import models[m
+ import serial[m
+ import numpy as np[m
+ import pandas as pd[m
+[31m-from pykeepass import PyKeePass[m
+[32m+[m[32m#from pykeepass import PyKeePass[m
+ from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar[m
+ from scipy.io import savemat, loadmat[m
+ from PyQt5.QtWidgets import QApplication, QMainWindow, QMessageBox, QSizePolicy[m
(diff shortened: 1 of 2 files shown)


micahwoodard:
@jessy-liao just added you to let you know that operator is being added

alexpiet (approved):
Thanks for the quick fix!

micahwoodard:
@XX-Yin this patch should resolve [#799](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/799). Pulling software for stimulus epoch instead of data_stream

XX-Yin (approved):

XX-Yin:
We probably need to merge the develop to the main after fixing https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/799#issuecomment-2537395989 later of today.

micahwoodard:
Yeah @alexpiet any opinions on how we handle this? Since production_testing merged with main yesterday, only things merged into develop will be #1156 and whatever fixes @XX-Yin makes. Merging develop into production_testing and then production_testing to main once tested seems okay to me.

XX-Yin:
> Yeah @alexpiet any opinions on how we handle this? Since production_testing merged with main yesterday, only things merged into develop will be #1156 and whatever fixes @XX-Yin makes. Merging develop into production_testing and then production_testing to main once tested seems okay to me.

That sounds good to me.

micahwoodard:
@XX-Yin actually it looks like quite a few merges were made since the last production_testing merge. I think to limit the amount of new issues, we should just merge this branch and your fix into production testing and then production_testing into main. I can make those pull requests if you can approve and merge?

micahwoodard:
> @XX-Yin actually it looks like quite a few merges were made since the last production_testing merge. I think to limit the amount of new issues, we should just merge this branch and your fix into production testing and then production_testing into main. I can make those pull requests if you can approve and merge? I think that will be fine in terms of merge errors? What do you think?

Ack I don't think there is a way around merging in the other branches since the other two branches were branched off of development. I guess the only option to fix is to merge develop into production_testing and main.

tash
```diff
@@ -0,0 +1,31 @@
+[1mdiff --git a/src/foraging_gui/Foraging.py b/src/foraging_gui/Foraging.py[m
```
XX-Yin, line 1: What is the tash file?

XX-Yin: After talking with Micah, this file seems to be added by mistake. @micahwoodard Could you make a PR to delete it?

Issue #818, opened 2024-10-28
"No record found" error while trying to fetch mouse 746729. Attempting to add mouse model to Slims.


alexpiet:
This issue was automatically generated from GUI logs.<br><br>file: Foraging.py<br>function: _AddWaterLogResult<br>line 1255<br>logfile: W10DT714028/W10DT714028-B_gui_log_2024-10-23_13-32-08.txt<br>boxes: 3B<br>count: 1<br>text: "No record found" error while trying to fetch mouse 746729. Attempting to add mouse model to Slims.<br><br><br>Log context<br>

```
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976290.278304, '127.0.0.1', 4013]
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976290.411136, '127.0.0.1', 4013]
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976290.538496, '127.0.0.1', 4013]
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976290.670368, '127.0.0.1', 4013]
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/LeftLickTime', 24976290.670464, '127.0.0.1', 4013]
03:13:51 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976290.841024, '127.0.0.1', 4013]
03:14:15 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976314.927872, '127.0.0.1', 4013]
03:14:16 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976315.06272, '127.0.0.1', 4013]
03:14:16 PM:INFO:rigcontrol:rigcontrol.py:msg_handler:line 56:['/RightLickTime', 24976315.200544, '127.0.0.1', 4013]
03:18:54 PM:INFO:Foraging:Foraging.py:_Save:line 2451:Saving current session, ForceSave=True
03:18:54 PM:INFO:Foraging:Foraging.py:_GetWaterCalibration:line 1008:Loaded Water Calibration
[...]

PR #1211, opened 2024-12-16
Fix 818 id name

micahwoodard:
  
### Describe changes:
- fix mismatch between mouse id and experimenter name
- resolves [#818](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/818)
- name and id should match in load drop down 
- None
- [x] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2975,7 +2975,7 @@ class Window(QMainWindow):
                         experimenters.append(name)
                         break
         dates = [datetime.fromtimestamp(os.path.getmtime(os.path.join(filepath, path))) for path in mouse_dirs]
-        two_week = [mouse_dir for mouse_dir, mod_date in zip(mouse_dirs, dates) if (now - mod_date).days <= 14]
+        two_week = [mouse_dir for mouse_dir, mod_date in zip(mice, dates) if (now - mod_date).days <= 14]
         return mice, experimenters, two_week
 
     def _Open(self,open_last = False,input_file = ''):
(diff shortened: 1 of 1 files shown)


micahwoodard:
@alexpiet @XX-Yin could we also merge this into production_testing so we can role out updates this week? I feel like merging production_testing with this issue unresolved still lingering could be confusing and cause issues.

src/foraging_gui/Foraging.py
```diff
@@ -2987,7 +2987,9 @@ def _Open(self,open_last = False,input_file = ''):
                 # list of mice, experimenters, and two week in chronological order form date modified
                 mice, experimenters, two_week = self._Open_getListOfMice()
                 # only add mice from two weeks in drop down.
-                W = MouseSelectorDialog(self, [m + ' ' + n for m, n in zip(two_week, experimenters)])
+                print(mice)
```
XX-Yin, line 2990: Why print(mice) and print(two_week) here?

XX-Yin:
> @alexpiet @XX-Yin could we also merge this into production_testing so we can role out updates this week? I feel like merging production_testing with this issue unresolved still lingering could be confusing and cause issues.

I'm fine with making a parallel merging to the production_testing

src/foraging_gui/Foraging.py
micahwoodard: I need to do one more test in 446 but integration testing is going on. I will take out before merging!

alexpiet (approved):
Yes, sorry I missed this

XX-Yin (approved):

PR #1220, opened 2024-12-17
Update trial logic

XX-Yin:
  
### Describe changes:
Updated the trial logic for optogenetics session. 

Previously, the optogenetics session is using a trial logic without the delay period (due to precise control of duration during the ITI period) for both opto and control trials. 
image

After updating, the opto trials and control trials will use the opto and normal trial logic, respectively. 
image

See description on the "Describe changes"

No

No

Tested on 323_Ephys3


XX-Yin:
@alexpiet @micahwoodard This is ready for review.

XX-Yin:
@micahwoodard @alexpiet Can you review this PR this week so that we can merge it EOD Friday?

alexpiet:
Does this change anything about non-opto sessions?

@hanhou any input?

micahwoodard:
I do not know enough bonsai to provide any meaningful review on this, but I'll look through and see if anything makes sense to me!

XX-Yin:
> Does this change anything about non-opto sessions?
> 
> @hanhou any input?

No, nothing was changed for the non-opto session.

hanhou:
I don't have enough bonsai knowledge to review the workflow change either. But in general, could you explain why maintaining two trial structures is necessary? What is the delay period you mentioned? Could you discuss with Bruno to see whether the change is necessary and optimized?

XX-Yin:
> I don't have enough bonsai knowledge to review the workflow change either. But in general, could you explain why maintaining two trial structures is necessary? What is the delay period you mentioned? Could you discuss with Bruno to see whether there are better solutions?
> 
> And maybe I misunderstood, but the first diagram you posted seems to be the updated one, as there are two separate workflows?

Yes, there are two trial logic. For the opto, we don't have the delay period in order to give precise duration of laser. For normal trials, the logic is as before (no change in the code), and we have a delay period. 

I have implemented the same logic in the previous version. However, they are separated and can't be used in the same session. Here, the update is to permit using two logic in the same session. 

I have already talked with Bruno long time ago.

XX-Yin:
I don't change the trial logic here, instead I combine two trial logic to a single chain. Previously, only one trial logic can be used (either normal or the opto) in a single session (due to termination of the other chain after first trial). After combining two trial logic to a single chain, two trial logic can be interleaved in a single session.

XX-Yin:
> I don't change the trial logic here, instead I combine two trial logic to a single chain. Previously, only one trial logic can be used (either normal or the opto) in a single session (due to termination of the other chain after first trial). After combining two trial logic to a single chain, two trial logic can be interleaved in a single session.

alexpiet:
I'm nervous because I also don't understand the bonsai workflow to give a meaningful review. Its problematic that we don't have anyone else to review this

XX-Yin:
> I'm nervous because I also don't understand the bonsai workflow to give a meaningful review. Its problematic that we don't have anyone else to review this

I got your concern. This is definitely something we should improve. In terms of this PR, it doesn't change the trial logic. We may consider merging it.

alexpiet (approved):

PR #1223, opened 2024-12-19
AddingAindBehaviorTaskLogicModel

### LINKED ISSUE #800: Implement AindBehaviorTaskLogicModel
- Waiting for #799 

micahwoodard:
  
### Describe changes:
- Adding AindBehaviorTaskLogicModel into framework 
- resolves #800 
- None
- None 
- [x] 446/7


micahwoodard:
I'm going to change the imports to pull from the [Aind.Behavior.DynamicForaging](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging) repo. Will request review then but should really only be the import that changes as well as the pyproject.toml

src/foraging_gui/Foraging.py
```diff
@@ -243,6 +254,233 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
             self._ReconnectBonsai()
         logging.info('Start up complete')
 
+    def initialize_task_parameters(self) -> DynamicForagingParas:
```
XX-Yin, line 257: Can we move this part to a separate python file? Foraging.py is already quite large. And this part is relatively independent. The idea is to make AindBehaviorTaskLogicModel relatively independent and modular.

XX-Yin:
It seems that the current modification logic is like this: 

1. Copy parameters to behavior_task_logic_model -> update these parameters if parameters are changed 
2. Copy parameters to behavior_task_logic_model -> replacing existing parameters using behavior_task_logic_model (like **replacing self.warmup.currentText()** using **self.behavior_task_logic_model.task_parameters.warmup**) 

What is the entire logic and purpose here in terms of step 1 and step 2? How is this related to the curriculum?

micahwoodard:
I don't think this is ready to merge. I prematurely thought this is what it would look like but I think this will change as I figure out how the curriculum works with this

micahwoodard:
> What is the entire logic and purpose here in terms of step 1 and step 2? How is this related to the curriculum?

 For number 1, the intent was so changes in gui would be reflected in the task logic. Number 2 of makes it so we are not using the gui state as a reference of how the task was run but rather a defined schema.  

For how this is related to curriculum, the stages in the curriculum require a task logic model. We can adopt some of Bruno's architecture hare and leverage it in the curriculum.

XX-Yin:
Can we create a task logic model at the end of the session? Here the step 2 is circular. For curriculum, we only need a to set the parameters? Just like what we are doing now.

micahwoodard:
This would be a great discussion to loop @alexpiet and @hanhou into. The task parameters (same ones we are putting in the task logic model) are part of the Stage class in the aind-behavior-curriculum. In my mind, when a mouse is loaded, the gui will query slims for the trainer state of this mouse. The trainer state will include the stage of this mouse, and therefore the task parameters for the session. We can use the stage task parameters to update the task logic model and this is what the gui will use to set up and run the session. If a mouse is loaded in, that will disabled the task parameter widgets so mice are limited to the session outlined in slims. Integrating the task logic model and using it to run a session rather than the widgets will make it easier to use the trainer state from slims and also provides better documentation on how the session was run. That is the workflow and benefits as I see them. If we're not all on the same page, let me know and I can update my approach

alexpiet:
Yes, thats right @micahwoodard. The task parameters will already be determined at the start of the session by the curriculum, so we should make the task model at the start of the session.

hanhou:
> This would be a great discussion to loop @alexpiet and @hanhou into. The task parameters (same ones we are putting in the task logic model) are part of the Stage class in the aind-behavior-curriculum. In my mind, when a mouse is loaded, the gui will query slims for the trainer state of this mouse. The trainer state will include the stage of this mouse, and therefore the task parameters for the session. We can use the stage task parameters to update the task logic model and this is what the gui will use to set up and run the session. If a mouse is loaded in, that will disable the task parameter widgets so mice are limited to the session outlined in slims. Integrating the task logic model and using it to run a session rather than the widgets will make it easier to use the trainer state from slims and also provides better documentation on how the session was run. That is the workflow and benefits as I see them. If we're not all on the same page, let me know and I can update my approach

Agreed.

XX-Yin:
" We can use the stage task parameters to update the task logic model and this is what the gui will use to set up and run the session." 

Why do we need to use the stage task parameters to update the task logic model and then update the GUI parameters instead of updating the GUI parameters directly?

micahwoodard:
@hanhou @alexpiet Would the idea be to use only the task logic parameters instead of the TP_ parameters to run a session? So we're not updating the gui parameters, but we're only using the task logic parameters to run the session, increasing readability and reproducibility of code.

micahwoodard:
Okay this is ready for rereview using external task logic repo, using pythonic naming conventions, and replacing widgets and TP parameters with task logic. I also changed the LineEdits to Spinboxes when applicable to simplify logic

micahwoodard:
Okay this has been tested and is ready for review/merging.

pyproject.toml
```diff
@@ -28,7 +28,8 @@ dependencies = [
     "aind-auto-train@git+https://github.com/AllenNeuralDynamics/aind-foraging-behavior-bonsai-automatic-training.git@main",
     "aind-slims-api@git+https://github.com/AllenNeuralDynamics/aind-slims-api@main",
     "aind-dynamic-foraging-models@git+https://github.com/AllenNeuralDynamics/aind-dynamic-foraging-models@main",
-    "aind-behavior-services >=0.8, <0.9",
+    "aind-behavior-dynamic-foraging@git+https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging@main",
+    "aind-behavior-services",
```
alexpiet, line 32: Is there a reason you removed the version dependency of aind-behavior-services?

micahwoodard: It was conflicting with the latest version of the aind-behavior-curriculum that I'm using in aind-behavior-dynamic-foraging

alexpiet: I think pinning a version would be a good idea, since I expect there will be breaking changes

micahwoodard: I just pinned. I want to check one more time in the rig room that all the dependencies are fine and then I'll merge

micahwoodard:
Also as a note, I need to run one last check that if a mouse is loaded in, the task_logic model will be updated with the correct values. I know that the auto-train-dialog will update the TP parameters and widgets and I'm pretty sure that will trigger an update to the task_logic, but I need to double check. Will probably do when the 446 is empty around 3

micahwoodard:
> Also as a note, I need to run one last check that if a mouse is loaded in, the task_logic model will be updated with the correct values. I know that the auto-train-dialog will update the TP parameters and widgets and I'm pretty sure that will trigger an update to the task_logic, but I need to double check. Will probably do when the 446 is empty around 3

Okay just check it updates

XX-Yin:
@micahwoodard This is a big change. Thank you for making this. However, I'm still not fully convinced the change logic.

micahwoodard:
@XX-Yin @alexpiet @hanhou seems like there is a discrepancy in views about this change. I am unsure of the best way forward

alexpiet:
@XX-Yin I'm not sure what your concern is, could you elaborate? We have been planning to adopt the schemas from Aind.Behavior.Services for several months (almost a year actually), and its an essential step in adopting the AIND curriculum system

XX-Yin:
The current logic for making changes is showing below. I don't think they are directly related to the curriculum. Generally, for the curriculum, you only need to update parameters as we are doing now, then analyze the behavior data in the cloud and adjust the curriculum accordingly. 

- Copy parameters to behavior_task_logic_model -> update these parameters if parameters are changed
- Copy parameters to behavior_task_logic_model -> replacing existing parameters using behavior_task_logic_model (like replacing self.warmup.currentText() using self.behavior_task_logic_model.task_parameters.warmup)

micahwoodard:
I don't understand your argument. Integrating the task logic model and using it to run a session rather than the widgets will make it easier to use the trainer state from slims and also provides better documentation on how the session was run, increasing repeatability and visibility. This pr is an interim step in moving away from the widgets to define a task. We can't move away from the widgets if we keep using the widgets.

XX-Yin:
"Integrating the task logic model and using it to run a session rather than the widgets will make it easier to use the trainer state from slims and also provides better documentation on how the session was run, increasing repeatability and visibility. " 

This is definitely not the case. The logic of the new curriculum should be similar to the current one, meaning the two changes I mentioned earlier are not necessary.

micahwoodard:
I agree, the logic of the new curriculum is similar to the current version. The stage model in the aind-behavior-curriculum has a task_logic attribute that I've defined to be the AindDynamicForagingTaskLogic type that I've used for the task_logic model in this pr. This means that when a curriculum is pulled from slims, running the curriculum only requires setting the task_logic model equal to the curriculum stage task_logic model. That to me is easier and clearer than updating widgets. Also, since the widgets run a check every keypress where values are manipulated, there is less of a chance that these parameters are incorrectly set and used in a session. Because of the validation required of values in the task_logic, this is unlikely to happen.

alexpiet:
@XX-Yin We have been planning this move for almost a year. As we adopt the new curriculum manager we can streamline the way parameters are set. Right now parameters get modified from AutoTrain and from the widgets, meaning there are multiple opportunities for mistakes and modifications. In the new system, all parameter modifications will happen through the curriculum.

This PR is one step of many in transitioning to the new curriculum. So what seems like an unnecessary step of moving parameters from the task model to the widgets will go away after we fully adopt the new system.

micahwoodard:
The functionality of updating the task_logic model when a widget changes would be removed when the new curriculum logic is added. It's only used here because auto train updates the widgets to apply the curriculum [here](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/c1aee4728c806d407e031eabfb257d14ecd0dd06/src/foraging_gui/Dialogs.py#L2868C4-L2991C44). This a confusing and unclear way to set task logic values. When the new curriculum is integrated in, the widgets won't be used to run the session so there won't be a circular dependency like there is now.

XX-Yin:
"The functionality of updating the task_logic model when a widget changes would be removed when the new curriculum logic is added."

This approach makes more sense; however, relying solely on behavior_task_logic_model to update training parameters introduces some issues. It would mean that training parameters can only be read from the online curriculum, preventing users from updating them through the GUI. 

We should put the merge on hold and wait until we have a thorough plan and PR in place.

micahwoodard:
I think restricting users from deviating from the curriculum when loading a mouse is part of the point of using slims and the new curriculum. I also think we do have a thorough plan. See [this issue](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1329) for outline on how to integrate new curriculum, this [document ](https://alleninstitute-my.sharepoint.com/:w:/g/personal/alex_piet_alleninstitute_org/EfUouTXRNIxOhGAqXsoXBoEBwQuG6sK5ru7MOSjtUaj1Mw?e=arRzN7) for outline of total plan, and [this project board](https://github.com/orgs/AllenNeuralDynamics/projects/111) for implementation with milestones and issues.

alexpiet:
> It would mean that training parameters can only be read from the online curriculum, preventing users from updating them through the GUI.

Yes, this is exactly the intent.

JeremiahYCohen:
This looks like a straightforward implementation of the plan for the AIND curriculum (as @micahwoodard wrote). Let's proceed with this PR and discuss in person if there are concerns.

XX-Yin:
> This looks like a straightforward implementation of the plan for the AIND curriculum (as @micahwoodard wrote). Let's proceed with this PR and discuss in person if there are concerns.

@JeremiahYCohen Merging this PR doesn’t make sense at this stage. At the very least, we should resolve the circular issue first. Making changes will be much more difficult after the merge.

PR #1227, opened 2024-12-22
skip the manifest generation for ephys session

XX-Yin:
  
### Describe changes:
Exclude the generation of manifest file for the ephys session

Exclude the generation of manifest file for the ephys session

No

No

Tested in 323_Ephys3

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3766,20 +3766,6 @@ class Window(QMainWindow):
             self.Metadata_dialog.project_info = project_info
             self.Metadata_dialog.ProjectName.addItems([project_name])
         return project_name
-    def _empty_initialize_fields(self):
-        '''empty fields from the previous session'''
-        self.open_ephys=[]
-        self.ManualWaterVolume=[0,0]
-        # clear camera start and end time
-        if not self.Camera_dialog.StartRecording.isChecked():
-            self.Camera_dialog.camera_start_time=''
-            self.Camera_dialog.camera_stop_time=''
-        # clear fiber start and end time
-        if hasattr(self, 'fiber_photometry_end_time'):
-            del self.fiber_photometry_end_time
-        if not self.StartExcitation.isChecked():
-            if hasattr(self, 'fiber_photometry_start_time'):
-                del self.fiber_photometry_start_time
                 
     def _Start(self):
         '''start trial loop'''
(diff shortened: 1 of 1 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -4661,13 +4675,18 @@ def _generate_upload_manifest(self, session: Session):
             :param session: session to use to create upload manifest
         '''
 
+        # skip manifest generation for test mouse
         if self.behavior_session_model.subject in ['0','1','2','3','4','5','6','7','8','9','10']:
             logging.info('Skipping upload manifest, because this is the test mouse')
             return
-
+        # skip manifest generation if automatic upload is disabled
         if not self.Settings['AutomaticUpload']:
```
alexpiet, line 4683: Why not use the "AutomaticUpload" setting to turn off manifest generation instead?

src/foraging_gui/Foraging.py
```diff
@@ -3766,7 +3766,21 @@ def _set_default_project(self):
             self.Metadata_dialog.project_info = project_info
             self.Metadata_dialog.ProjectName.addItems([project_name])
         return project_name
-
+    def _empty_old_data(self):
```
alexpiet, line 3769: What is this doing?

alexpiet (changes requested):

src/foraging_gui/Foraging.py
XX-Yin: For the ephys rig, we are also using the manifest generation for behavior sessions without the ephys recording.

src/foraging_gui/Foraging.py
XX-Yin: To clear or initialize certain fields from previous sessions when the new session is started without restarting the GUI. These fields may interfere with the generation of the manifest file if not reset.

src/foraging_gui/Foraging.py
alexpiet: Why are you not uploading ephys sessions? We need a clear plan for how we are going to upload and process ephys data, and how we will generate ephys specific metadata. We had previously discussed with @KanghoonJ that ephys specific metadata should be handled by a ephys specific software base.

src/foraging_gui/Foraging.py
alexpiet: That doesn't seem related to skipping the manifest generation for ephys sessions. Can you make a dedicated PR

XX-Yin:
@alexpiet This is ready for review again

alexpiet:
@XX-Yin How is the ephys data getting uploaded? 

Are all ephys scientists using the same approach? @KanghoonJ @yonibrowning @galenlynch @alakunina

XX-Yin:
I believe we are all using the CSV uploading strategy. Generating csv file and sending jobs on the web.

alexpiet (approved):

PR #1265, opened 2024-12-30
Optical tagging

XX-Yin:

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -3067,7 +3067,7 @@ class OpticalTaggingDialog(QDialog):
         if save_folder=='':
             return
         # create the file name AnimalID_Date(day+hour+minute)_OpticalTaggingResults.csv
-        save_file = os.path.join(save_folder, f"{self.optical_tagging_par['AnimalID']}_{datetime.now().strftime('%Y-%m-%d-%H-%M')}_OpticalTaggingResults.json")
+        save_file = os.path.join(save_folder, f"{self.MainWindow.ID}_{datetime.now().strftime('%Y-%m-%d-%H-%M')}_OpticalTaggingResults.json")
         # save the data 
         with open(save_file, 'w') as f:
             json.dump(self.optical_tagging_par, f, indent=4)
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -3043,10 +3043,12 @@ class OpticalTaggingDialog(QDialog):
         self._connectSignalsSlots()
         self.MainWindow = MainWindow
         self.optical_tagging_par={}
-
+        self.finish_tag = 1
+        self.threadpool = QThreadPool()
     def _connectSignalsSlots(self):
         self.Start.clicked.connect(self._Start)
         self.WhichLaser.currentIndexChanged.connect(self._WhichLaser)
+        self.EmergencyStop.clicked.connect(self._emegency_stop)
 
     def _Start(self):
         '''Start the optical tagging'''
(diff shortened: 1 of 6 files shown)


XX-Yin:
@hanhou @alexpiet @micahwoodard This PR is ready for review.

XX-Yin:
There are quite a few changes. I'm happy to explain them in person if needed.

alexpiet:
This is a good example of something that belongs in a dedicated ephys tool, not the behavior gui. Adding it to the behavior gui complicates the code, but worse, means that other behavior tasks don't have the tool. We need a plan for how we are going to support ephys techniques across many behavioral tasks. Let's discuss with a broader group

XX-Yin:
"the optical tagging belongs to the ephys tool"? Can you explain more?

XX-Yin:
@alexpiet 
People are eager to use this tool for experiments that combine optical tagging and behavior, with compatibility with the downstream pipeline being a critical requirement. If you believe another solution might be needed, please arrange a discussion with a broader group as soon as possible.

src/foraging_gui/Dialogs.py
```diff
@@ -1247,7 +1251,7 @@ def _StartCamera(self):
                 # sleep for 1 second to make sure the trigger is off
                 time.sleep(1)
             # Start logging if the formal logging is not started
-            if self.MainWindow.logging_type!=0 or self.MainWindow.logging_type==-1:
+            if self.MainWindow.logging_type!=0:
```
alexpiet, line 1254: Why does this change?

src/foraging_gui/Dialogs.py
```diff
@@ -3029,3 +3033,535 @@ def headerData(self, col, orientation, role):
...
+            "laser_2_power": self.Laser_2_power.text(),
+            "frequency": self.Frequency.text(),
+            "pulse_duration": self.Pulse_duration.text(),
+            "duration_each_cycle": self.Duration_each_cycle.text(),
+            "interval_between_cycles": self.Interval_between_cycles.text(),
+            "cycles_each_condition": self.Cycles_each_condition.text(),
+        }
+        # save the data 
+        with open(save_file, 'w') as f:
+            json.dump(self.optical_tagging_par, f, indent=4)
```
alexpiet, line 3099: Does this save the data in the main JSON file? or a different file?

src/foraging_gui/Foraging.py
```diff
@@ -2338,6 +2341,16 @@ def _Camera(self):
         else:
             self.Camera_dialog.hide()
 
+    def _OpticalTagging(self):
```
alexpiet, line 2344: What is this function doing?

src/foraging_gui/Foraging.py
```diff
@@ -2681,6 +2694,8 @@ def _Save(self,ForceSave=0,SaveAs=0,SaveContinue=0,BackupSave=0):
         Obj['MetadataFolder']=self.MetadataFolder
         Obj['SaveFile']=self.SaveFile
 
+        # save optical tagging parameters
+        Obj['optical_tagging_par']=self.OpticalTagging_dialog.optical_tagging_par
```
alexpiet, line 2698: Should this parameters only get saved if optical tagging was actually performed?

src/foraging_gui/GenerateMetadata.py
```diff
@@ -516,7 +525,7 @@ def _session(self):
         self._get_ophys_stream()
         self._get_high_speed_camera_stream()
         self._get_session_time()
-        if self.session_start_time == '' or self.session_end_time == '':
+        if (self.session_start_time == '' or self.session_end_time == '') and (self.Obj['optical_tagging_par']['optical_tagging_start_time']=='' or self.Obj['optical_tagging_par']['optical_tagging_start_time']==''):
```
alexpiet, line 528: I'm not sure why we need this check here

alexpiet:
@XX-Yin This PR is extremely difficult to review. Part of the problem is because the optical tagging functions interact with a lot of parts of the existing GUI in ways that I don't understand. For example, the session start time. Is it the case that optical tagging sessions always have behavior? Is optical tagging always before, or always after the behavioral session? Does it matter?

I'm not really sure how to proceed given the complexity. I don't want to just merge it in and seeing what breaks

XX-Yin:
> @XX-Yin This PR is extremely difficult to review. Part of the problem is because the optical tagging functions interact with a lot of parts of the existing GUI in ways that I don't understand. For example, the session start time. Is it the case that optical tagging sessions always have behavior? Is optical tagging always before, or always after the behavioral session? Does it matter?
> 
> I'm not really sure how to proceed given the complexity. I don't want to just merge it in and seeing what breaks

@alexpiet Sorry for this complexity from this PR. I can explain every details and address your concerns.

src/foraging_gui/GenerateMetadata.py
XX-Yin: The previous code will not generate the session metadata when there is no behavior data(defined by empty session_start_time and session_end_time). In this update, the session metadata will still be generated when there is optical tagging stimulus without the behavior data.

src/foraging_gui/Foraging.py
alexpiet: Yes, I think its better to not have anything saved than save something that is empty

src/foraging_gui/Dialogs.py
XX-Yin: It simplified the code. **self.MainWindow.logging_type==-1** can be covered by **self.MainWindow.logging_type!=0**

src/foraging_gui/Foraging.py
XX-Yin: I will remove it.

XX-Yin: updated

XX-Yin:
> @XX-Yin This PR is extremely difficult to review. Part of the problem is because the optical tagging functions interact with a lot of parts of the existing GUI in ways that I don't understand. For example, the session start time. Is it the case that optical tagging sessions always have behavior? Is optical tagging always before, or always after the behavioral session? Does it matter?
> 
> I'm not really sure how to proceed given the complexity. I don't want to just merge it in and seeing what breaks

The behavior is not necessary for the optical tagging. The optical tagging can happen before or after the behavioral session, which doesn't matter.

XX-Yin:
@alexpiet I understand your concern. The main interaction is happening on the GenerateMetadata.py, which is easy to test.

XX-Yin:
@hanhou I have tested this branch multiple times. I can test more. In terms of the code structure, the optical tagging will not influence other parts. Just like the water calibration, which is relatively isolated from other parts. Here the main interaction is happening on metadata. I think other people will be interested in this new function, too.

XX-Yin:
@alexpiet @hanhou @micahwoodard We are ready to review this PR again. I have tested this branch in the ephys rig for several weeks in multiple mice and sessions.

alexpiet:
Does the optical tagging interact with the behavior in any way? If not, then why integrate those features into this task specific GUI?

src/foraging_gui/Foraging.py
```diff
@@ -201,6 +201,7 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         self.OpenLaserCalibration=0
         self.OpenCamera=0
         self.OpenMetadata=0
+        self.OpenOpticalTagging=0
```
alexpiet, line 204: What is this variable `OpenOpticalTagging` doing?

XX-Yin: This indicates whether the Optical Tagging window is open, to avoid opening it multiple times.

XX-Yin:
> Does the optical tagging interact with the behavior in any way? If not, then why integrate those features into this task specific GUI?

I think this is something we’ve discussed a lot before. There are several reasons both for and against doing it this way, but this function is the only one that allows us to perform timestamp alignment and metadata generation in a unified manner that’s compatible with our current pipeline.

We typically run the optical tagging before or after the behavior session. In this update, the optical tagging can be used independently or combined with other GUI functions.

alexpiet:
@XX-Yin I understand this is important functionality, and I want to make sure you can do the experiments you want. However we need to be intentional about adding additional functions to the GUI. I have several concerns about this PR

- @micahwoodard is close to finishing a large update of the GUI. This update will include encapsulating physiology modalities (like optical tagging) in data models. Adding new features right now will complicate those updates. 
- Adding optical tagging tools here prevents their re-use on other behavioral tasks. The better place for this sort of tool is an ephys specific tool. There has been a lot of discussion about creating flexible tools to work with multiple modalities, but nothing concrete yet. @jsiegle Do you have a timeline or plan to create ephys tools that are used by all ephys scientists for things like metadata collection, optical tagging, or other ephys specific tasks?
- If we must add optical tagging to this GUI, I think we need to wait for Micah's large updates to get merged, and we need to design a flexible system for adding optional features that will only be used on the physiology rigs.

XX-Yin:
@alexpiet You asked me merge this to the main last week. I'm ok to use this function in a separate branch.

XX-Yin:
@alexpiet If we plan to merge this PR, it would be best to do it sooner rather than later. @micahwoodard can then test this update within the context of the new curriculum updates as a whole. Otherwise, we may have to wait several months for the curriculum update to be completed and then need to test again.

hanhou:
Balancing flexibility and standardization is indeed a challenge. I suggest we look for a temporary middle ground: is it possible to isolate the optical tagging control as a "plugin" to the core GUI, where it could be turned off completely by the configure.json file on the training rigs? 

@micahwoodard do you have any idea how to properly implement this? This "core + plugin" structure can eventually be used in the general purpose task launcher as well.

XX-Yin:
> Balancing flexibility and standardization is indeed a challenge. I suggest we look for a temporary middle ground: is it possible to isolate the optical tagging control as a "plugin" to the core GUI, where it could be turned off completely by the configure.json file on the training rigs?
> 
> @micahwoodard do you have any idea how to properly implement this? This "core + plugin" structure can eventually be used in the general purpose task launcher as well.

Completely isolating the plugin from the core would require major code changes, which may not be feasible under our current setup. However, this approach might be more suitable for the task launcher.

Here, some major tools (e.g. optical tagging, meta data generation, water calibration, optogenetics) are relatively isolated from the main code, but not completely.

alexpiet:
Based on discussion on 10/07/2025, this PR can be closed because this functionality was merged in with the random reward PR

PR #1268, opened 2025-01-02
Feature/monument

jessy-liao:
  
### Describe changes:

Added feature for users to set monument position (saved per rig). 
- Positions will be displayed relative to a monument position that is saved in zookeeper (ex. monument position of [1,2,2,1] will show [-1,-2,-2,-1] when stage is at origin)
- Users can save new monument positions through the widget. 

[SIPE Ticket](http://mpe-redirects/MPETracking/ItemID=3734)

Users now have a reference to where monument is.

Normal deployment cycle. Just need to update stage widget package.

This was tested on FRG2A

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3208,7 +3208,7 @@ class Window(QMainWindow):
                             2: float(last_positions['y2']),
                             3: float(last_positions['z'])
                         }
-                        self.move_aind_stage(positions)
+                        self.stage_widget.stage_model.update_position(positions)
                         step_size = self.stage_widget.movement_page_view.lineEdit_step_size.returnPressed.emit()
                 elif 'B_NewscalePositions' in Obj.keys() and len(Obj['B_NewscalePositions']) != 0:  # cross compatibility for mice run on older version of code.
                     last_positions = Obj['B_NewscalePositions'][-1]
@@ -3249,14 +3249,6 @@ class Window(QMainWindow):
         self.keyPressEvent() # Accept all updates
         self.load_tag=1
         self.ID.returnPressed.emit() # Mimic the return press event to auto-engage AutoTrain
-
-    def move_aind_stage(self, positions: dict):
-        """
-        Move all axis of stage in stage widget
-
-        :param positions: Dict where key = motor index, and value = the desired position to move the motor
-        """
-        self.stage_widget.stage_model.update_position(positions)
 
     def _LoadVisualization(self):
         '''To visulize the training when loading a session'''
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -809,10 +809,12 @@ class Window(QMainWindow):
             self._UpdatePosition(current_position,(0,0,0))
             return {axis: float(pos) for axis, pos in zip(['x', 'y', 'z'], current_position) }
         elif self.stage_widget is not None:     # aind stage
-            return {'x': float(self.stage_widget.movement_page_view.lineEdit_x.text()),
-                    'y1': float(self.stage_widget.movement_page_view.lineEdit_y1.text()),
-                    'y2': float(self.stage_widget.movement_page_view.lineEdit_y2.text()),
-                    'z': float(self.stage_widget.movement_page_view.lineEdit_z.text())}
+            # Get absolute position of motors in AIND stage
+            positions = self.stage_widget.stage_model.get_current_positions_mm()
+            return {'x': positions[0],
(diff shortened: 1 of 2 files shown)


micahwoodard (commented):
lol looks great! I'm curious how will this interact with the current workflow of moving the stage to the previous session position? Will the previous session position be wrong the first run with the monument in place and then afterwards will be fine?

jessy-liao:
@micahwoodard 
Ah true, I forgot about that. This will break the previous session position since that code handles movement by updating the lineEdit boxes. The lineEdit values now get calculated relative to the current monument position.

Possible solution: I can create a utility function like "move_to_positions_mm" that you can call and it'll bypass the relative monument position and move to exact position coordinates. 

With the suggested solution above, you would have the following functions to handle position movements from dynamic foraging GUI:

- get_current_positions_mm (gets positions in mm, ignoring monument)
- move_to_positions_mm (move motors to positions in mm, ignoring monument)

Let me know what you think, or if you have other suggestions. 

If you're cool with this solution I can work on a new stage_widget dev version and make the respective changes to dynamic foraging GUI as well.

micahwoodard:
@jessy-liao I think having those utility functions would be great. @alexpiet wondering if it matters whether gui is logging monument relative positions or absolute positions.

jessy-liao:
@micahwoodard @alexpiet 
It shouldn't matter what positions they are logging as long as it is consistent. Currently all the logs will display the absolute positions.

jessy-liao:
Ok, tested it with the new version on FRG.2-A and it looks like everything works.

alexpiet:
We just need to coordinate with RAs when the coordinate system switches (@ellahiltonvano )

src/foraging_gui/Foraging.py
```diff
@@ -3199,11 +3201,14 @@ def _Open(self,open_last = False,input_file = ''):
...
-                        self.stage_widget.movement_page_view.lineEdit_y2.setText(str(last_positions['y2']))
-                        self.stage_widget.movement_page_view.lineEdit_z.setText(str(last_positions['z']))
-                        self.move_aind_stage()
+                        # Move AIND stage to the last session positions
+                        positions = {
+                            0: float(last_positions['x']),
+                            1: float(last_positions['y1']),
+                            2: float(last_positions['y2']),
+                            3: float(last_positions['z'])
+                        }
```
micahwoodard, line 3210: Just curious why you chose to use 0,1,2,3 as keys instead of axes within stage widget. Does it make indexing or communicating with hardware easier? I've always defaulted to axes so it's cool to see it another way

src/foraging_gui/Foraging.py
```diff
@@ -3245,12 +3250,13 @@ def _Open(self,open_last = False,input_file = ''):
...
-    def move_aind_stage(self):
+    def move_aind_stage(self, positions: dict):
         """
         Move all axis of stage in stage widget
+
+        :param positions: Dict where key = motor index, and value = the desired position to move the motor
         """
-        positions = self.stage_widget.movement_page_view.get_positions_from_line_edit()
-        self.stage_widget.movement_page_view.signal_position_change.emit(positions)
+        self.stage_widget.stage_model.update_position(positions)
```
micahwoodard, line 3259: We could probably get rid of this function since it's only used once and it's a one liner? I think I made it it's own function when moving all axes of the stage was a bit more complicated. Not hugely important, but could reduce extra clutter

src/foraging_gui/Foraging.py
```diff
@@ -3199,11 +3201,14 @@ def _Open(self,open_last = False,input_file = ''):
...
-                        self.stage_widget.movement_page_view.lineEdit_z.setText(str(last_positions['z']))
-                        self.move_aind_stage()
+                        # Move AIND stage to the last session positions
+                        positions = {
+                            0: float(last_positions['x']),
+                            1: float(last_positions['y1']),
+                            2: float(last_positions['y2']),
+                            3: float(last_positions['z'])
+                        }
+                        self.move_aind_stage(positions)
```
micahwoodard, line 3211: Maybe call self.stage_widget.stage_model.update_position(positions) instead

micahwoodard (approved):
Looks good to me! Maybe just remove function

src/foraging_gui/Foraging.py
jessy-liao: I had written them as indexes when I first started working on the stage widget. 

I agree that labeling by axes makes more sense but the rest of the stage widget code uses numerical index so I am continuing it for uniformity. 

I will probably refactored this one day because it is confusing.

PR #1272, opened 2025-01-03
Fix orientation

XX-Yin:
  
### Describe changes:
fixed the direction mismatch between the recording visualization and preview visualization. 

image
resolved https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/829
No
No
tested in 323_Ephys3

### DIFF
--- a/src/workflows/foraging.bonsai
+++ b/src/workflows/foraging.bonsai
@@ -7852,7 +7852,7 @@
                                 </PropertyMappings>
                               </Expression>
                               <Expression xsi:type="io:CsvWriter">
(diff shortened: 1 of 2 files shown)


XX-Yin:
@hagikent Can you check if this orientation is what you want?

hagikent:
Duplicates: @XX-Yin
don't use flip... the whole point of using Image Rotate Visualizer node is that the rotation can be managed in the layout file (and also you can flip it after running bonsai). as per the first comment, please fix the layout file not the bonsai per se.

XX-Yin:
close this PR as we don't want to use the flip

PR #1304, opened 2025-01-10
Fix warmup

XX-Yin:
  
### Describe changes:
Explicitly set the **UpdateParameter** to 1 after the warmup is turned off. 

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/840#issuecomment-2584676842

No

No

I can't repeat this issue on my side, but this PR could potentially solve it.


alexpiet (commented):
I don't know what the update parameters variable does. @hanhou you might be better prepared to review this

XX-Yin:
@hanhou @alexpiet Can we please review this PR?

alexpiet:
> @hanhou @alexpiet Can we please review this PR?

- I don't understand what "self.win.UpdateParameters" does, or why this fixes the issue with the infinite warmup. 
- @XX-Yin Can you explain the logic?
- @hanhou @micahwoodard can you take a look?

XX-Yin:
I need more tests to answer your question.

micahwoodard:
Is it maybe an issue on the Bonsai side? Seems like the parameters are all updating on the gui side based on the screenshots Han sent in the issue. Looking at the [updates leading up to this merge](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1287), there were several relating to the bonsai workflow in regards to running wheel and updating trial logic. Just a guess based on the merges that happened that week

micahwoodard:
I also had several merges that week but I don't think any of them would have affected the warm up progression. I'll double check though

micahwoodard:
Actually looking at the logs it seems like the error `RuntimeWarning: Mean of empty slice
  RewardedHistoryF[i+1-self.kernel_size]=np.nanmean(RewardedHistoryT[i+1-self.kernel_size:i+1])
` pops up after every trial. Could this be why it's not out of warmup because something is being disrupted before the trial logic can be set? I'll investigate the other sessions and see if this is a common trend.

micahwoodard:
> Actually looking at the logs it seems like the error `RuntimeWarning: Mean of empty slice RewardedHistoryF[i+1-self.kernel_size]=np.nanmean(RewardedHistoryT[i+1-self.kernel_size:i+1]) ` pops up after every trial. Could this be why it's not out of warmup because something is being disrupted before the trial logic can be set?

It's a pretty common error in all of the sessions, but looking through it's probably not disrupting anything since it's nested in a try finally

XX-Yin:
> Is it maybe an issue on the Bonsai side? Seems like the parameters are all updating on the gui side based on the screenshots Han sent in the issue. Looking at the [updates leading up to this merge](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1287), there were several relating to the bonsai workflow in regards to running wheel and updating trial logic. Just a guess based on the merges that happened that week

I can reproduce this issue but haven’t yet got a change to identify the root cause. 

Parameters are not updated if you check the behavior JSON file. There is an extra safety parameter, **UpdateParameter**, that enables parameter updates. When parameters change, **UpdateParameter** becomes false to disable parameter updates until users press Enter to confirm the change. Here, I can confirm that UpdateParameter becomes false when the AIND stage moves, and it is not excluded from the text check as it is for the newscale stage. However, I don't know why the parameters are not reverted when the warmup is turned off, where Enter should be pressed by default.

micahwoodard:
@XX-Yin it seems like the parameters eventually change in [python_gui_log_776293_2025-01-08_11-17-01.txt](https://github.com/user-attachments/files/18355148/python_gui_log_776293_2025-01-08_11-17-01.txt). I just check BlockMin and Multiplier parameters though so maybe others didn't. Is this after the user pressed enter then?

XX-Yin:
> @XX-Yin it seems like the parameters eventually change in [python_gui_log_776293_2025-01-08_11-17-01.txt](https://github.com/user-attachments/files/18355148/python_gui_log_776293_2025-01-08_11-17-01.txt). I just check BlockMin and Multiplier parameters though so maybe others didn't. Is this after the user pressed enter then?

Did you check the behavior json file?

micahwoodard:
> > @XX-Yin it seems like the parameters eventually change in [python_gui_log_776293_2025-01-08_11-17-01.txt](https://github.com/user-attachments/files/18355148/python_gui_log_776293_2025-01-08_11-17-01.txt). I just check BlockMin and Multiplier parameters though so maybe others didn't. Is this after the user pressed enter then?
> 
> Did you check the behavior json file?

Oh sorry I linked the wrong file. This is the one I was looking at that Han sent [776293_2025-01-08_11-17-01.json](https://github.com/user-attachments/files/18366009/776293_2025-01-08_11-17-01.json) . I was just pointing out that eventually the values change. Like this is the TP_Multiplier:
image

and here is TP_BlockMin:
image

It may not be after 50 trials though so I was wondering if you were saying that this was after the user pressed enter again.

XX-Yin:
For this session, it seems that the user hit the Enter to enable parameters change
image

micahwoodard (commented):
Seems like a minimal change that is worth merging in if you think it will fix the warmup issue @XX-Yin

src/foraging_gui/MyFunctions.py
```diff
@@ -289,6 +289,7 @@ def _CheckWarmUp(self):
             self.win.keyPressEvent()
             self.win.NextBlock.setChecked(True)
             self.win._NextBlock()
+            self.win.UpdateParameters=1
```
micahwoodard, line 292: Would it be worth calling _GetTrainingParameters directly or is there a reason to wait?

XX-Yin: Ideally, when UpdateParameters is set to 1, the TrainingParameters will be updated in next trial.  I believe calling _GetTrainingParameters directly will fix this issue, it just doesn't keep the logic consistent. @micahwoodard If you'd like to submit a PR for this, I'm fine with that as well.

@micahwoodard @alexpiet Could we merge this PR first? I don't have time to go into details of this issue this week, but I can do it next week.

alexpiet: sounds good

Issue #1174, opened 2024-12-13
Spontaneous crashes during session caused by backup saving


micahwoodard:
RA's have been having issues where gui's will crash mid session. It seems like this may be related to a bug in the save thread. Some examples of these instances are in issues: [#800](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/800), ⁠[#797](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/797), ⁠[#791](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/791), ⁠[#788](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/788), ⁠[#779](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/779), [#752](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/752)

XX-Yin:
I also speculate that it’s related to the backup save. The interference between different threads may cause crash, but the exact reason is unknown. I will turn off the backup save today through changing the settings in all behavior rig.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Micah Woodard ***@***.***>
Sent: Friday, December 13, 2024 8:31:13 AM
To: AllenNeuralDynamics/dynamic-foraging-task ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [AllenNeuralDynamics/dynamic-foraging-task] Spontaneous crashes during session (Issue #1174)


RA's have been having issues where gui's will crash mid session. It seems like this may be related to a bug in the save thread. Some examples of these instances are in issues: #800<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/800>, ⁠#797<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/797>, ⁠#791<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/791>, ⁠#788<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/788>, ⁠#779<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/779>, #752<https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/752>

micahwoodard:
@XX-Yin just to clarify, turning off backup save requires changing the setting.json file?

XX-Yin:
Yes. I can do that for all behavior rig this afternoon

[...]

PR #1331, opened 2025-01-27
Adding thread safety with data_lock to the backup save

XX-Yin:
  
### Describe changes:
Thread safety with data_lock:
   Ensure all access to shared data (GeneratedTrials, etc.) is guarded with a thread lock.

Resolved: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1174#issuecomment-2613946119

No

Change *save_each_trial* in the foraging settings json to *True*, or just remove this field. 

Not yet

No


alexpiet:
@micahwoodard Can you review this, since you have more knowledge about threads

src/foraging_gui/Foraging.py
```diff
@@ -4396,6 +4382,24 @@ def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
...
+            except Exception as e:
+                logging.error(f"Backup task failed: {e}")
+            finally:
+                self.backup_queue.task_done()
+
+    def perform_backup(self):
+        # Backup save logic
+        with self.data_lock:
+            self._Save(BackupSave=1)
+
```
micahwoodard, line 4402: Why use a queue and a lock this way? Seems like it would be easier/cleaner to start a backup_worker thread after every new trial and then have the back_up worker lock and save like this: 
```
 with self.data_lock:
            self._Save(BackupSave=1)
```
Then you wouldn't have this thread running full time with no way to break and clean up afterward.

micahwoodard: Also to make the lock mechanism more robust, it would be nice to use it everywhere save gets called. Or maybe at the beginning of the save function? That would ensure whenever you're saving the data, no other thread is also trying to do this

src/foraging_gui/Foraging.py
```diff
@@ -4228,7 +4227,7 @@ def end_session_log(self) -> None:
         else:
             logging.info(f'No active session logger')
 
-    def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
+    def _StartTrialLoop(self,GeneratedTrials,worker1):
```
micahwoodard, line 4230: I still think making this function a qobject and then putting it in a qthread would be really beneficial.  The use of QApplication.processEvents() can lead to crashes and is generally frowned upon.

src/foraging_gui/Foraging.py
XX-Yin: The lock prevents simultaneous access to the same data across different threads, which I believe is the primary cause of the crash. However, we don’t need to use the lock for the normal saving function, as no threads other than the main one are running in that case.

src/foraging_gui/Foraging.py
XX-Yin: We are still using "self.backup_queue.put(self.perform_backup)" to run the backup on a different thread and not dependent on QApplication.processEvents(). Can you expand more?

micahwoodard: Yeah I guess my comment here is unrelated to the backup saving so perhaps not the best pr to discuss it in. Using QApplication.processEvents() at all usually indicates the function should be it's own thread as it can cause race conditions causing unexplained crashes. I just think it would be good to restructure this specific function and wanted to voice that opinion again. But again, out of the scope of this pr so we don't have to dwell on it here.

XX-Yin: Makes sense.

src/foraging_gui/Foraging.py
micahwoodard: Why do you have to combine this with a queue though? I'm still unclear on that.

XX-Yin: In the backup_worker method, we can encapsulate try/except blocks around each task execution, logging errors and even deciding how to proceed if a task fails. This level of granularity makes it easier to isolate failures to individual tasks.

There are no other specific reasons to use the Queue over signals.finished.connect. Do you have any concerns?

micahwoodard: Could you not just do 
```
# Save data in a separate thread
if GeneratedTrials.B_CurrentTrialN>0 and self.save_each_trial and GeneratedTrials.CurrentSimulation==False:
   save_thread= threading.Thread(target=self.backup_worker)
   save_thread.start()

...
def perform_backup(self):
   # Backup save logic
   with self.data_lock:
      try:
          self._Save(BackupSave=1)
      except Exception as e:
          logging.error(f"Backup task failed: {e}")
```
? 

Not trying to nitpick, I just think I'm not understanding something. Only concern is readability and simplifying logic.

XX-Yin: I have updated this.

alexpiet:
@XX-Yin @micahwoodard I wanted to stress test this, so I added a line in the backup saving code that would raise an exception, and ran it on 3D. The error propagated to the main thread and was caught by the system hook "uncontrolled error". 

So the good news is it doesn't crash the GUI, and the error is caught. The bad news is I expected the try/except block to catch the error

src/foraging_gui/Foraging.py
```diff
@@ -4334,23 +4335,12 @@ def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
...
-                '''
+
                 # Save data in a separate thread
-                if GeneratedTrials.B_CurrentTrialN>0 and self.previous_backup_completed==1 and self.save_each_trial and GeneratedTrials.CurrentSimulation==False:
-                    self.previous_backup_completed=0
-                    self.GeneratedTrials_backup=copy.copy(self.GeneratedTrials)
-                    self.threadpool6.start(worker_save)
+                with self.data_lock:
+                    if GeneratedTrials.B_CurrentTrialN>0 and self.previous_backup_completed==1 and self.save_each_trial and GeneratedTrials.CurrentSimulation==False:
+                        self.previous_backup_completed=0
```
alexpiet, line 4342: Where does self.previous_backup_completed get set to 1?

alexpiet: oh nevermind. In the _thread_complete6 function

XX-Yin: It's here https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1331/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR3721. It will become to 1 when the thread finishes.

alexpiet:
> @XX-Yin @micahwoodard I wanted to stress test this, so I added a line in the backup saving code that would raise an exception, and ran it on 3D. The error propagated to the main thread and was caught by the system hook "uncontrolled error".
> 
> So the good news is it doesn't crash the GUI, and the error is caught. The bad news is I expected the try/except block to catch the error
> 
> To be clear, this is an improvement, so it shouldn't block merging it in. We should just understand what is happening

Oh! You just need to put the call in a try/except block like @micahwoodard suggested and it works!


```
def perform_backup(self):
   # Backup save logic
   with self.data_lock:
      try:
          self._Save(BackupSave=1)
      except Exception as e:
          logging.error(f"Backup task failed: {e}")
```

src/foraging_gui/Foraging.py
alexpiet: @micahwoodard Can you make a separate issue for this concern, so we can approve this PR?

XX-Yin:
> @XX-Yin @micahwoodard I wanted to stress test this, so I added a line in the backup saving code that would raise an exception, and ran it on 3D. The error propagated to the main thread and was caught by the system hook "uncontrolled error".
> 
> So the good news is it doesn't crash the GUI, and the error is caught. The bad news is I expected the try/except block to catch the error
> 
> To be clear, this is an improvement, so it shouldn't block merging it in. We should just understand what is happening

Thanks for testing. I speculate it's related to the logging here if the error information is not saved.  


 
    class Worker(QtCore.QRunnable):
      '''
       Worker thread

       Inherits from QRunnable to handler worker thread setup, signals and wrap-up.

    :param callback: The function callback to run on this worker thread. Supplied args and
                     kwargs will be passed through to the runner.
    :type callback: function
    :param args: Arguments to pass to the callback function
    :param kwargs: Keywords to pass to the callback function

    '''
    def __init__(self, fn, *args, **kwargs):
        super(Worker, self).__init__()

        # Store constructor arguments (re-used for processing)
        self.fn = fn
        self.args = args
        self.kwargs = kwargs
        self.signals = WorkerSignals()
        self.setAutoDelete(False) 
        # Add the callback to our kwargs

    @QtCore.pyqtSlot()
    def run(self):
        '''
        Initialise the runner function with passed args, kwargs.
        '''

        # Retrieve args/kwargs here; and fire processing using them
        try:
            result = self.fn(*self.args, **self.kwargs)
        except ValueError as e:
            exctype, value = sys.exc_info()[:2]
            self.signals.error.emit((exctype, value, traceback.format_exc()))
            logging.error(str(e))
        else:
            self.signals.result.emit(result)  # Return the result of the processing
        finally:
            self.signals.finished.emit()  # Done

alexpiet:
I think @micahwoodard is traveling the rest of the week. Based on my testing, this seems ready to merge

alexpiet (approved):

XX-Yin:
I also added the "try except" to the main thread to catch the error. @alexpiet tested this and showed that captured the error.

PR #1339, opened 2025-01-29
Fix B_Bias appending

### LINKED ISSUE #1337: Save Bias on trial by trial basis
Right now the side bias is being saved when its computed, but since its not computed on every trial, its not aligned to the other trial parameters. I propose saving the bias on each bias, and using the last-computed value (default to 0)

```TP_Bias = B_Bias[-1]```

For example, in this session, there were 725 trials, but B_Bias only has 200 values: behavior_761401_2025-01-28_12-54-00

@micahwoodard Is this the expected outcome?

micahwoodard:
  
### Describe changes:
- B_Bias correctly appends bias values across trials 
- resolves #1337 
- None
- None
- [ ] 446
- This is a non-breaking change, but the downstream processing needs to add the bias to the NWB file. Issue [here](https://github.com/AllenNeuralDynamics/aind-fip-nwb-base-capsule/issues/69)

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -173,6 +173,7 @@ class Window(QMainWindow):
         # create bias indicator
         self.bias_n_size = 200
         self.bias_indicator = BiasIndicator(x_range=self.bias_n_size)  # TODO: Where to store bias_threshold parameter? self.Settings?
+        self.bias_indicator.biasValue.connect(print)
         self.bias_indicator.biasValue.connect(self.bias_calculated)  # update dashboard value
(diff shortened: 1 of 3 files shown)


alexpiet:
@micahwoodard I'm testing this in 446 and I get an uncontrolled error "ValueError: operands could not be broadcast together with shapes (100,) (6,) (100,) (line 4037)

micahwoodard:
@alexpiet thanks for catching, should be updated now.

alexpiet (approved):

alexpiet:
@micahwoodard Looks good. I tested in 447 and verified the saved file has bias records that match the length of the completed trials

PR #1341, opened 2025-01-30
Making upload manifest failures more robust

alexpiet:

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2694,6 +2694,8 @@ class Window(QMainWindow):
             session = generated_metadata._session()
             if session is not None:     # skip if metadata generation failed
                 self.sessionGenerated.emit(session)   # emit sessionGenerated
+            else:
+                logging.error('Could not generated upload manifest, missing metadata')
 
             if BackupSave==0:
                 text="Session metadata generated successfully: " + str(generated_metadata.session_metadata_success)+"\n"+\
@@ -4675,6 +4677,7 @@ class Window(QMainWindow):
             logging.info('Skipping upload manifest, because this is an ephys session')
             return
 
+        logging.info('Generating upload manifest')
         try:
             if not hasattr(self, 'project_name'):
                 self.project_name = 'Behavior Platform'
@@ -4730,7 +4733,7 @@ class Window(QMainWindow):
             # Write the manifest file
             with open(filename,'w') as yaml_file:
                 yaml.dump(contents, yaml_file, default_flow_style=False)
-
+            logging.info('Finished generating manifest')
         except Exception as e:
             logging.error('Could not generate upload manifest: {}'.format(str(e)))
             QMessageBox.critical(self, 'Upload manifest',


alexpiet:
Maybe we should emit SessionGenerated in the "start" function after this  ```self.session_run = True   # session has been started```

src/foraging_gui/Foraging.py
```diff
@@ -4235,6 +4232,7 @@ def end_session_log(self) -> None:
     def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
         if self.Start.isChecked():
             logging.info('starting trial loop')
+            self.sessionGenerated.emit()   # Generate upload manifest
```
micahwoodard, line 4235: Should we wait until one trail has been generated? I remember at one time that was an issue that manifest were being created when no trials ran.

alexpiet: I can't remember why that was causing problems. Whats the easiest way to wait until one trial has been generated, other than relying on the backup save

src/foraging_gui/Foraging.py
```diff
@@ -4685,8 +4683,6 @@ def _generate_upload_manifest(self, session: Session):
             date_format = "%Y-%m-%d_%H-%M-%S"
             schedule = self.behavior_session_model.date.strftime(date_format).split('_')[0]+'_20-30-00'
             schedule_time = datetime.strptime(schedule,date_format) + timedelta(seconds=np.random.randint(30*60))
-            # This ID is outdated as of 11/21/2024. We will remove this comment once we confirm everything works
-            #capsule_id = 'c089614a-347e-4696-b17e-86980bb782c1'
             capsule_id = '0ae9703f-9012-4d0b-ad8d-b6a00858b80d'
             mount = 'FIP'
```
micahwoodard, line 4688: How are the modalities being generated now? It used to rely on the passed in session.data_stream

micahwoodard (changes requested):
I have questions about how the modality dictionary is being generated now that the session isn't being past in. Could be missing something though!

src/foraging_gui/Foraging.py
micahwoodard: We could put in a check in the while self.Start.isChecked(): loop and if GeneratedTrials.B_CurrentTrialN == 2, call self.sessionGenerated.emit()

src/foraging_gui/Foraging.py
alexpiet: Oof. I just missed that it was being used for that. I will need to rethink this

micahwoodard: You could maybe parse the modalities without it. Looks like FIB streams are present if fiber_photometry_start_time != ', and behavior_videos are present if Camera_dialog.camera_start_time != ''.

src/foraging_gui/Foraging.py
alexpiet: done. I use GeneratedTrials.B_CurrentTrialN == 1, since this counter starts at 0

src/foraging_gui/Foraging.py
alexpiet: done

src/foraging_gui/Foraging.py
```diff
@@ -4303,6 +4300,11 @@ def _StartTrialLoop(self,GeneratedTrials,worker1,worker_save):
                 if self.actionLicks_sta.isChecked():
                     self.PlotLick._Update(GeneratedTrials=GeneratedTrials)
 
+                # Generate upload manifest when we generate the second trial
+                # counter starts at 0
+                if GeneratedTrials.B_CurrentTrialN == 1:
+                    self.sessionGenerated.emit()
```
micahwoodard, line 4306: Since you changed the workflow of this, it doesn't really make sense for it to be a signal. It should probably just be a function call here. I think I made it a signal when it was in the _Save function and save_every_trial was True so I needed a way to ensure it only created the manifest once after the first trial.

src/foraging_gui/Foraging.py
```diff
@@ -72,7 +72,7 @@ def default(self, obj):
 
 class Window(QMainWindow):
     Time = QtCore.pyqtSignal(int) # Photometry timer signal
-    sessionGenerated = QtCore.pyqtSignal(Session)  # signal to indicate Session has been generated
+    sessionGenerated = QtCore.pyqtSignal()  # signal to indicate Session has been generated
```
micahwoodard, line 75: probably should just get rid of this

src/foraging_gui/Foraging.py
```diff
@@ -4730,7 +4729,7 @@ def _generate_upload_manifest(self, session: Session):
             # Write the manifest file
             with open(filename,'w') as yaml_file:
                 yaml.dump(contents, yaml_file, default_flow_style=False)
-
+            logging.info('Finished generating manifest')
         except Exception as e:
             logging.error('Could not generate upload manifest: {}'.format(str(e)))
             QMessageBox.critical(self, 'Upload manifest',
```
micahwoodard, line 4735: Also get rid of line 4741 disconnecting the slot

micahwoodard (approved):
Would work with the way it is now, but seems a little weird to keep using signals here

src/foraging_gui/Foraging.py
alexpiet: done

src/foraging_gui/Foraging.py
alexpiet: done

src/foraging_gui/Foraging.py
alexpiet: done

alexpiet:
- tested in 447
- removed the slot/signal system 
@micahwoodard You already approved, so I'm going to merge this. Speak up if you have any concerns

micahwoodard:
Looks good to me!

PR #1350, opened 2025-01-31
adding pydantic model for ForagingSettings.json and Settings_box.csv

### LINKED ISSUE #1343: Create pydantic model to validate ForagingSettings.json
We should validate the entries and values of ForagingSettings.json as we load them, to catch issues before anything else happens.

alexpiet:
  
### Describe changes:
- Adds a pydantic model for the ForagingSetting.json and Settings_box.csv file
- [x] Updates required fields of Settings_box.csv to match current needs
- [x] test that its ok to have extra entries in the settings file

- resolves #1343 
- Should prevent issues with missing files, or poorly inputed files

- will get a validation error if settings are missing or incorrect type

- none

- [x] tested

- no


alexpiet:
@micahwoodard This is a work in progress, I need to test on a rig, and add more fields on the BonsaiSettingsModel, but I'd like your feedback now

src/foraging_gui/settings_model.py
```diff
@@ -0,0 +1,61 @@
...
+    '''
+    # TODO, need to add more required fields here
+    Behavior: str = Field(pattern=r"^COM[0-9]+$")
+    Soundcard: str = Field(pattern=r"^COM[0-9]+$")
+    BonsaiOsc1: int
+    BonsaiOsc2: int
+    BonsaiOsc3: int
+    BonsaiOsc4: int
+    AttenuationLeft: int
+    AttenuationRight: int
```
micahwoodard, line 16: Why didn't you make these fields? I'm not a pydantic expert so I was just wondering

tash
```diff
@@ -1,31 +0,0 @@
...
- from PyQt5.QtWidgets import QApplication, QMainWindow, QMessageBox, QSizePolicy[m
-[36m@@ -252,6 +252,8 @@[m [mclass Window(QMainWindow):[m
-     def _show_disk_space(self):[m
-         '''Show the disk space of the current computer'''[m
-         total, used, free = shutil.disk_usage(self.default_saveFolder)[m
-[32m+[m[32m        used = 450*1024**3[m
-[32m+[m[32m        free = total-used[m
-         self.diskspace.setText(f"Used space: {used/1024**3:.2f}GB    Free space: {free/1024**3:.2f}GB")[m
-         self.DiskSpaceProgreeBar.setValue(int(used/total*100))[m
-         if free/1024**3 < 100 or used/total > 0.9:[m
```
micahwoodard, line 31: Is this file supposed to be here?

src/foraging_gui/settings_model.py
micahwoodard: Oh sorry still work in progress my bad

micahwoodard (approved):
Looks good and makes sense to me. I'm assuming the schemas will be fleshed out later

tash
alexpiet: I have no idea what this file is, I assume i got accidentally committed

alexpiet:
@micahwoodard Can you take a second look at this now that its finished?

src/foraging_gui/Foraging.py
```diff
@@ -1255,10 +1252,15 @@ def _GetSettings(self):
                     logging.error('Missing setting ({}), is required'.format(key))
                     raise Exception('Missing setting ({}), is required'.format(key))
 
+        # Check that settings are valid
+        DFTSettingsModel(**self.Settings)
```
micahwoodard, line 1256: Would there be any benefit to making the model an attribute, and then using the model to reference settings? The ones I can thing of would be better type hinting, documentation, and validation but if the code isn't really interacting with the Settings dictionary too much, it may not be worth it.

src/foraging_gui/settings_model.py
```diff
@@ -0,0 +1,139 @@
...
+    Defines a model for the Settings_box.csv file
+    """
+
+    Behavior: str = Field(
+        pattern=r"^COM[0-9]+$", description="COM port for Harp Behavior Board"
+    )
+    Soundcard: str = Field(
+        pattern=r"^COM[0-9]+$", description="COM port for Harp Behavior Board"
+    )
+    BonsaiOsc1: str = Field(
```
micahwoodard, line 17: Are these Osc channels used for specific communication? Like is BonsaiOsc1 used for communicating [normal behavior events](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/3f3950d9b2ea76bac8fd2e8fb1be3343c7fe96af/src/foraging_gui/Foraging.py#L1489C8-L1492C57) and channel 2 right and left reward time? Might be nice to be more specific if so. Or are these names locked in on the Bonsai side? That's the only place I see them being used.

src/foraging_gui/settings_model.py
```diff
@@ -0,0 +1,139 @@
+from pydantic import BaseModel
+from pydantic import Field
+from typing import Literal, Optional
+
+
+class BonsaiSettingsModel(BaseModel):
```
micahwoodard, line 6: The naming convention is a little weird in these schemas since it's a mix of camel and snake case. Feel like it should be all snake case based on convention. Might be a little annoying since all of the boxes would have to change though.

src/foraging_gui/Foraging.py
alexpiet: I think that makes sense to do, but its outside the scope of this PR. Making a separate issues: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1354

src/foraging_gui/settings_model.py
alexpiet: They are locked in on the Bonsai side. They are always the same for Box A, always the same for Box B, etc. Once 447 upgrades are complete we can make them the same for all boxes

src/foraging_gui/settings_model.py
alexpiet: Right, its a weird mix. I think its outside the scope of this PR to fix

PR #1352, opened 2025-02-03
Fix disabled UncoupledReward associated with behavior_session_model

XX-Yin:
  
### Describe changes:
Changing the self.behavior_session_model.experiment to self.Task.currentText().

When activating the curriculum, the uncoupled reward was disabled. Upon investigation, this behavior is linked to the **behavior_session_model**. Specifically, **self.behavior_session_model.experiment** is copied from **self.currentText()**.

When the **Task** is switched to a **coupled task**, it triggers a change in **self.currentText()**, which subsequently leads to **self._Task** disabling the unused **uncoupled reward field**. However, **self.behavior_session_model.experiment** does not update immediately; instead, it lags behind. As a result, **self.behavior_session_model.experiment** continues using the previous value.

This delay in synchronization may introduce additional issues, potentially affecting other functionalities that rely on **self.behavior_session_model.experiment** being up to date.

Solved https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/897#issuecomment-2632276338

No

No

Can repeat and fix this issue on my own computer

No


XX-Yin:
@alexpiet @hanhou @micahwoodard This is ready for review

XX-Yin:
There are potentially other issues associated with the circular logic. 

Copy gui parameters to behavior_session_model, and then use behavior_session_model to update parameters.

XX-Yin:
I can reproduce the issue where the uncoupled reward field was incorrectly set, which was also addressed here. When the initial uncoupled reward field was disabled due to the Task being set to a coupled task, the curriculum was unable to update this field. This is because a disabled field is not modified by the curriculum during updating parameters.

code here 

            # If the parameter is disabled by the GUI in the first place, skip it
            # For example, the field "uncoupled reward" in a coupled task.
            if not widget.isEnabled():
                logger.info(f''' Widget "{key}" has been disabled by the GUI. skipped...''')
                continue

hanhou (approved):
Thank Xinxin for this fix. I'll approve this so that the urgent training parameter problems can be resolved. Meanwhile, @micahwoodard when you get a chance, could you look at why the `self.behavior_session_model.experiment` is not updating correctly and if there are any other potential issues?

alexpiet:
This seems like a quick fix, I'm fine with merging it in for the moment. But lets work on cleaning up the underlying logic. We shouldn't be keeping track of state variables in GUI widgets.

micahwoodard:
@XX-Yin @alexpeit if we wanted to solve this issue while still using the session model, we could connect the signal of updating the session before the self._Task function is connected. That would trigger the update to the session model before the widgets are updated.

micahwoodard:
> Thank Xinxin for this fix. I'll approve this so that the urgent training parameter problems can be resolved. Meanwhile, @micahwoodard when you get a chance, could you look at why the `self.behavior_session_model.experiment` is not updating correctly and if there are any other potential issues?

I think it's updating correctly but the order in which the signals are connected effects how the functions are called. So the _Task function is called before the session model is updated which I believe is why we're seeing this behavior

PR #1349, opened 2025-01-31
removing Go Cue Decibel from metadata dialog

alexpiet:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
Right now, the go cue amplitude is saved in Settings_box.csv, and also can be manually entered on the Metadata dialog. This means there are two sources of truth, and the metadata dialog can accidently be modified. I removed the metadata dialog version, since this has to be calibrated and set in the settings file. 

### What issues or discussions does this update address?
- resolves #1345 
- I think this is also why we are getting pydantic validation errors from the empty string

### Describe the expected change in behavior from the perspective of the experimenter
- You cannot modify the go cue amplitude in the metadata dialog

### Describe any manual update steps for task computers
- none

### Was this update tested in 446/447?
- [x] tested

### Does this update impact downstream processing by adding new saved files, or changing their format? If so, have you documented changes?
- no

PR #1357, opened 2025-02-06
Handle empty go cue decibel

XX-Yin:
  
### Describe changes:
When the go cue decibel is not provided, setting it to 74dB in the session metadata and logging error information. 

Session metadata error when the go cue decibel is not set in the foraging settings file. 

No

No

Tested on my own computer

No


alexpiet:
This error should be handled by other updates that are merging in next week
- Remove the decibel from the metadata dialog, which I think is why we are getting empty strings (https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1349)
- Add a validation step that prevents the session from running if the decibel is not in the settings file (https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1350)

But your PR handles a separate case of loading an old session that doesn't have the decibel saved, and then you try to regenerate the metadata. So I think its a good addition.

alexpiet (approved):

PR #1369, opened 2025-02-11
high-speed camera recording window direction fix

hagikent:
A patch to flip camera recording window to match the preview window. Originally those windows were designed to be consistent, but at one point, an inadvertently flipped local change (likely in a ephys rig) was merged. This is to revert it back.
  
### Describe changes:
Horizontal direction of the camera recording is flipped so it would match preview and the "intuitive" direction.

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/889

NA
NA
in 447 2D
No. Pure change in visualization.

### DIFF
--- a/src/workflows/foraging.bonsai.layout
+++ b/src/workflows/foraging.bonsai.layout
@@ -68,8 +68,8 @@
         <Y>3</Y>
       </Location>
       <Size>
-        <Width>1506</Width>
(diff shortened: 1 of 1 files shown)


hagikent:
@XX-Yin 
Pure visualization change in the .layout file.

XX-Yin (approved):

PR #1374, opened 2025-02-11
Use most recent experimenter name when loading mice

alexpiet:
  
### Describe changes:
When loading a mouse, the pop up currently displays the FIRST experimenter to run a mouse, rather than the most recent. This is confusing to users because inconsistent information is displayed in the loading selection box, the verification box, the GUI window, and the confirmation before starting a session. 

This PR will make this more consistent. The loading selection box, verification box, and GUI window will also show the most recent experimenter. If the user then changes the experimenter name, that new name will be used when starting the session. 

- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/905

More consistent information about experimenter names

none

- [ ] tested

no

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2998,13 +2998,13 @@ class Window(QMainWindow):
         filepath = os.path.join(self.default_saveFolder,self.current_box)
         now = datetime.now()
         mouse_dirs = os.listdir(filepath)
-        mouse_dirs.sort(reverse=True, key=lambda x: os.path.getmtime(os.path.join(filepath, x)))   # in order of date modified
+        mouse_dirs.sort(reverse=True)
         mice = []
         experimenters = []
         for m in mouse_dirs:
             session_dir = os.path.join(self.default_saveFolder, self.current_box, str(m))
             sessions = os.listdir(session_dir)
-            sessions.sort(reverse=True, key=lambda x: os.path.getmtime(os.path.join(session_dir,x)))
+            sessions.sort(reverse=True)
             for s in sessions:
                 if 'behavior_' in s:
                     json_file = os.path.join(self.default_saveFolder,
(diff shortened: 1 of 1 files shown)


XX-Yin (approved):
looks good!

src/foraging_gui/Foraging.py
```diff
@@ -2998,12 +2998,13 @@ def _Open_getListOfMice(self):
         filepath = os.path.join(self.default_saveFolder,self.current_box)
         now = datetime.now()
         mouse_dirs = os.listdir(filepath)
-        mouse_dirs.sort(reverse=True, key=lambda x: os.path.getmtime(os.path.join(filepath, x)))   # in order of date modified
+        mouse_dirs.sort(reverse=True, key=lambda x: os.path.getmtime(os.path.join(filepath,x))) # in order of date modified
```
micahwoodard, line 3001: Is this just sorting the filename strings alphabetically now and it works out because of the date at the end?

src/foraging_gui/Foraging.py
```diff
@@ -2998,12 +2998,13 @@ def _Open_getListOfMice(self):
...
         mice = []
         experimenters = []
         for m in mouse_dirs:
             session_dir = os.path.join(self.default_saveFolder, self.current_box, str(m))
             sessions = os.listdir(session_dir)
+            sessions.sort(reverse=True)
```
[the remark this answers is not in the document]
alexpiet: This is sorting the sessions based on the name of the folders, which is the name of the session. So this is sorting by session date, newest first

src/foraging_gui/Foraging.py
alexpiet: I'm also confused by this! Its sorting the folders of mouse names by the modification date. So presumably this is sorting by the order in which the mice ran their first sessions.

I tried removing this to sort by mouse_id numerically, and it caused some weird behavior, so I reverted it to how it was working before

alexpiet: To be clear, I didn't change this part

micahwoodard: Yeah I originally did the sorting, and I'm trying to remember why. I think it was so the experimenters were in the right order or something.

PR #1384, opened 2025-02-16
Adding in session, task, opto, and fip schemas and widgets

### LINKED ISSUE #800: Implement AindBehaviorTaskLogicModel

micahwoodard:
  
### Describe changes:
- in retrospect, this probably should have been broken up into three smaller merges, but I was too far before I realized that. This pr incorporates the session, task_logic, fip, and opto schemas into the UI. Now, parameters relating to how sessions are set up come from these created models, not widgets. In order to allow users to still be able to change the schemas from the gui, I created custom widgets that ingest full schemas and produce widgets for each field and subfields in a given schema. This keeps the model up to date. These widgets also have a function called apply_schema so if a schema is loaded in from slims, it can update all of the relevant widgets easily. 

- resolves #800 
- widgets will look different but shouldn't effect session
- [ ] 446/7


XX-Yin:
@micahwoodard @alexpiet There are quite a lot of changes, so before making major changes it is best to explain in detail the logic of the changes and how they relate to the entire picture.

XX-Yin:
also looping in @hanhou

micahwoodard:
@XX-Yin yeah this is quite ready for review. I still need to better document changes and include images of what's changes and also add a few finishes to the code.

micahwoodard:
Outlining major task parameter/logic change here. New task parameter widget looks like this: 

Behavior GUI: 

https://github.com/user-attachments/assets/9573795f-95c5-4cfa-a592-50741a7850e2

Ephys: 
image

The widget needs a AindDynamicForagingTaskParameters object to initialize and then will update that model with changes made by the user. The widget basically groups the sub fields of the [new task logic schema](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging/pull/4). There are checkboxes to set optional fields like auto_water and warmup to None. Based on what session you have configured (i.e. have uncouple reward or reward n checked), the task drop down will update and allow user to select from only the subsection of task that makes sense with what is currently configured. It doesn't disable any widgets depending on what task like the old widget. The gui task logic won't use those parameters any ways and they are still included in the schema. This feels a little weird so maybe that indicates that the schema is not fully capturing the task logic? Let me know what you guys think. 

One thing that needed rewriting was the _WaterVolumnManage1 and _WaterVolumnManage2 functions. Since user changing the spout time was disabled a while ago, I removed _WaterVolumnManage1 and made it so the volume was the only thing that could change the spout time in [update_valve_open_time](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/5add2349595a96db7b1fae5feb413470bddb45e2/src/foraging_gui/Foraging.py#L1669C1-L1684C82). 

Some cosmetic changes also happened as a result of changing the parameters tab. I deleted the RewardPairs2 label in the BehaviorGui. Since that info was already being displayed in the dashboard, I felt okay removing it. Let me know if it is critical to have it in the parameters widget. I also moved the NextBlock button to be next to the start button in both guis. Again. let me know if it is critical to have this in the parameters tab. For the Ephys gui, I also moved the Give Reward buttons above the start button. Let me know if that is not a good placement. Happy to move wherever. 

Another big change was removing keyPress and other functions that iterated through every widget. Since task logic will now rely on model, it's not longer necessary to log and track changes for every key press. Also, I think the checking input type functionality should be replaced with widget validators instead. This is a pretty big change, but I think important to try and reduce code complexity and move away from using the widget as a way to run the session. There might be some unintended consequences to this so let me know concerns. 

Other than what is mentioned above, the logic for how to run the task should be exactly the same, just pulling from the task_logic attribute rather than TP_ parameters or widgets

micahwoodard:
Outline major opto changes here. Widget now looks like this: 

https://github.com/user-attachments/assets/4a8d0d38-a252-414f-b707-859e573dafe3



Same thing as the task parameter widget. The widget needs a Optogenetics object to initialize and then will update that model with changes made by the user. The widget basically groups the sub fields of the [new opto schema](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging/pull/4). You can add laser colors and have the ability to change the laser locations, start and end conditions and protocol. 

Instead of using [numbers to specify condition](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/edf5c31bf0243a1b69091a4b8b7706d319296a30/src/foraging_gui/MyFunctions.py#L1424C9-L1432C28), the logic now sets the[ selected_condition attribute to a LaserColor object](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/5add2349595a96db7b1fae5feb413470bddb45e2/src/foraging_gui/MyFunctions.py#L1400C17-L1400C65). This object contains all of the previous named [CLP_ parameters](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/edf5c31bf0243a1b69091a4b8b7706d319296a30/src/foraging_gui/MyFunctions.py#L1259C9-L1273C64) so those were mostly removed as well. Other than that the logic should be the same. Having a little trouble testing but hopefully will be able to this afternoon. 

I'm still working on the widget sizing because its a little wonky on the rigs.

@XX-Yin I think you know the most about the opto sessions so let me know if I missed any key functionality in the schema

micahwoodard:
Outline major session changes here. Widget now looks like this:

ForagingGUI:

image


ephys: 
image

The auto train button is removed since further in the refactor it won't be used. The streamlit button for the foraging gui is moved up next to the start button. The experiment drop down is now coupled with the experimenter and subject because they are coupled in the models. The notes widget is placed where the old notes widget was though. Refactor here is also very minimal where logic should be the same but is drawing from model instead of widgets.

micahwoodard:
@XX-Yin @alexpiet @hanhou I'm actually going to spend sometime and separate out this pr. I think that will lead to more effective reviews and make it easier

XX-Yin:
Some general questions:

1. How do the new widgets validate incorrect data types?
2. How do the new widgets distinguish between typed parameters and the actual parameters used in the task?
3. Why is it necessary to rewrite these widgets in the Python code? How do we coordinate future development in Qt Designer?

micahwoodard:
@XX-Yin 

> 1. How do the new widgets validate incorrect data types?
     
Numerical values are validated with QIntValidators and QDoubleValidators so only ints or floats can be entered to those fields. Any value that is range restricted should also be restricted within the validator. Any value that has restricted text values should be limited to QComboBoxes. All others are up to the user but since those are fields like notes and experimenter, there are not "wrong" text values that will cause issues in the downstream code. Is that all of the cases you are concerned with or are there more cases?

> 2. How do the new widgets distinguish between typed parameters and the actual parameters used in the task?

I'm not sure what you mean? 

> 3. Why is it necessary to rewrite these widgets in the Python code? How do we coordinate future development in Qt Designer?

I think it's easier to write custom widgets in python, but just a preference. Also, you couldn't use Qt Designer to layout these widgets because they are populated based on a schema. The main plus I see doing it this way is that if/when the schemas need to be changed, the changes will be automatically reflected in the widget with no extra effort. And I don't know about future development. Do you have any ideas?

micahwoodard:
Oh also for number 3, another value add is that there is a function called apply_schema where it will take a schema model and apply the field values to the widget. This will make it really easy to pull session, task_logic, opto, fip models from slims and apply it to widgets. We would have to do mapping of schemas to widgets other wise so this seemed simpler/less time consuming to me.

XX-Yin:
2. How do the new widgets distinguish between typed parameters and the actual parameters used in the task?

I was asking when parameters in the widget will be used to run the task. Previously, you need to press enter to confirm the change of parameters, which is a double check and can avoid unintended use of parameters. 

----------

It's quite challenging to produce a well-organized, clean UI purely in Python code. Typically, you have to run the code, see how the UI looks, then readjust it—an inconvenient and time-consuming process. With Qt Designer, the primary advantage lies in direct UI arrangement and the separation of design from core logic. If you replace everything with Python code, it will complicate future updates in Qt Designer.

micahwoodard:
> I was asking when parameters in the widget will be used to run the task. Previously, you need to press enter to confirm the change of parameters, which is a double check and can avoid unintended use of parameters.

When you press enter or finish editing any of the schema widgets, the schema will be updated with the latest widget value. So, the next iteration of trial that uses the schema will have the latest value. Is this what you mean?



> It's quite challenging to produce a well-organized, clean UI purely in Python code. Typically, you have to run the code, see how the UI looks, then readjust it—an inconvenient and time-consuming process. With Qt Designer, the primary advantage lies in direct UI arrangement and the separation of design from core logic. If you replace everything with Python code, it will complicate future updates in Qt Designer.

Yeah that makes sense. How often are we updating the UI in Qt designer? Since these widgets are all based on schemas, I don't think there would be a need to update them using Qt designer. I think they would be separate enough to not interfere with update using Qt Designer.  Do you have any specific concerns in mind?

PR #1390, opened 2025-02-18
adding in session and task logic models

### LINKED ISSUE #800: Implement AindBehaviorTaskLogicModel

micahwoodard:
  
### Describe changes:
Outlining major task parameter/logic change here. New task parameter widget looks like this: 

Behavior GUI: 

https://github.com/user-attachments/assets/9573795f-95c5-4cfa-a592-50741a7850e2

Ephys: 
image

The widget needs a AindDynamicForagingTaskParameters object to initialize and then will update that model with changes made by the user. The widget basically groups the sub fields of the [new task logic schema](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging/pull/4). There are checkboxes to set optional fields like auto_water and warmup to None. Based on what session you have configured (i.e. have uncouple reward or reward n checked), the task drop down will update and allow user to select from only the subsection of task that makes sense with what is currently configured. It doesn't disable any widgets depending on what task like the old widget. The gui task logic won't use those parameters any ways and they are still included in the schema. This feels a little weird so maybe that indicates that the schema is not fully capturing the task logic? Let me know what you guys think. 

One thing that needed rewriting was the _WaterVolumnManage1 and _WaterVolumnManage2 functions. Since user changing the spout time was disabled a while ago, I removed _WaterVolumnManage1 and made it so the volume was the only thing that could change the spout time in [update_valve_open_time](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/5add2349595a96db7b1fae5feb413470bddb45e2/src/foraging_gui/Foraging.py#L1669C1-L1684C82). 

Some cosmetic changes also happened as a result of changing the parameters tab. I deleted the RewardPairs2 label in the BehaviorGui. Since that info was already being displayed in the dashboard, I felt okay removing it. Let me know if it is critical to have it in the parameters widget. I also moved the NextBlock button to be next to the start button in both guis. Again. let me know if it is critical to have this in the parameters tab. For the Ephys gui, I also moved the Give Reward buttons above the start button. Let me know if that is not a good placement. Happy to move wherever. 

Another big change was removing keyPress and other functions that iterated through every widget. Since task logic will now rely on model, it's not longer necessary to log and track changes for every key press. Also, I think the checking input type functionality should be replaced with widget validators instead. This is a pretty big change, but I think important to try and reduce code complexity and move away from using the widget as a way to run the session. There might be some unintended consequences to this so let me know concerns. 

Other than what is mentioned above, the logic for how to run the task should be exactly the same, just pulling from the task_logic attribute rather than TP_ parameters or widgets

Outline major session changes here. Widget now looks like this:

ForagingGUI:

image

ephys: 
image

The auto train button is removed since further in the refactor it won't be used. The streamlit button for the foraging gui is moved up next to the start button. The experiment drop down is now coupled with the experimenter and subject because they are coupled in the models. The notes widget is placed where the old notes widget was though. Refactor here is also very minimal where logic should be the same but is drawing from model instead of widgets. 
- resolves #800 
None
- [ ] 446/7


src/foraging_gui/schema_widgets/session_parameters_widget.py
```diff
@@ -0,0 +1,93 @@
+from foraging_gui.schema_widgets.schema_widget_base import SchemaWidgetBase, add_border, create_widget
+from aind_behavior_services.session import AindBehaviorSessionModel
+from datetime import datetime
+from PyQt5.QtWidgets import QTextEdit, QLabel
+from PyQt5.QtGui import QIntValidator
+
+class SessionParametersWidget(SchemaWidgetBase):
+
+    """
+    Widget to expose task logic for fiber photometry sessions
```
alexpiet, line 10: this comment seems wrong?

src/foraging_gui/Foraging.py
```diff
@@ -241,6 +278,33 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
...
         logging.info('Start up complete')
 
+    def update_session_task(self, task_type: Literal["coupled", "uncoupled", "rewardN"]):
+        """
+        Configure session task based on task parameters configured
+        :param task_type: task type currently configured
+        """
+
+        if task_type == "coupled":
+            for i in range(2, 5):
```
alexpiet, line 288: whats going on here with "i in range()"?

src/foraging_gui/Foraging.py
```diff
@@ -296,7 +360,7 @@ def _LoadUI(self):
         uic.loadUi(self.default_ui, self)
         if self.default_ui=='ForagingGUI.ui':
             logging.info('Using ForagingGUI.ui interface')
-            self.label_date.setText(str(date.today()))
+            #self.label_date.setText(str(date.today()))
```
alexpiet, line 363: remove commented out code

alexpiet:
@micahwoodard A couple minor things - the big question is how to test

micahwoodard:
@alexpiet yeah I am also not sure how to test robustly. Would there be a test mouse available that could run a couple of session on this branch to fully test out if it's doing what is expected? We could manually set a curriculum for a couple of sessions, maybe put some debug log statements that display the task, session, opto, and fib parameters that are being used to make sure they match up with the widget values. 

Another big question I have is how to deal with post processing. The TransferToNWB file relies solely on the TP parameters. Should I map those values to the schemas? I think because the some of the TP_ parameters are lists and it's iterating through every value at every trial, this is a little more complicated than just mapping values. Since the schema can't hold a record of every value used, should we lock changing parameters during a session?

micahwoodard:
Also @alexpiet I could write up some unit tests for checking that a schema updates as expected and a trial executes as expected if people think that would be useful?

src/foraging_gui/Foraging.py
micahwoodard: It's specifying the index of the items in the self.session_widget.experiment_widget which is a combobox. The list is [
            "Coupled Baiting",
            "Coupled Without Baiting",
            "Uncoupled Baiting",
            "Uncoupled Without Baiting",
            "RewardN"
        ]

So if the task type is coupled, then widget will disable the "Uncoupled Baiting", "Uncoupled Without Baiting", "RewardN" selections  (indexes 2, 3, 4) and enable the "Coupled Baiting", and "Coupled Without Baiting" selections (index 0, 1). In the video, it shows how different configurations of the task parameters allow for different tasks to be selectable in the self.session_widget.experiment_widget and this is the logic that is handling that.

src/foraging_gui/schema_widgets/session_parameters_widget.py
micahwoodard: yep!

PR #1393, opened 2025-02-18
adding back opto logic

micahwoodard:
  
### Describe changes:
Outline major opto changes here. Widget now looks like this: 

https://github.com/user-attachments/assets/4a8d0d38-a252-414f-b707-859e573dafe3

Same thing as the task parameter widget. The widget needs a Optogenetics object to initialize and then will update that model with changes made by the user. The widget basically groups the sub fields of the [new opto schema](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging/pull/4). You can add laser colors and have the ability to change the laser locations, start and end conditions and protocol. 

Instead of using [numbers to specify condition](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/edf5c31bf0243a1b69091a4b8b7706d319296a30/src/foraging_gui/MyFunctions.py#L1424C9-L1432C28), the logic now sets the[ selected_condition attribute to a LaserColor object](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/5add2349595a96db7b1fae5feb413470bddb45e2/src/foraging_gui/MyFunctions.py#L1400C17-L1400C65). This object contains all of the previous named [CLP_ parameters](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/edf5c31bf0243a1b69091a4b8b7706d319296a30/src/foraging_gui/MyFunctions.py#L1259C9-L1273C64) so those were mostly removed as well. Also, instead of using the combo box to determine if opto is run, it checks if the schemas laser_color field is an empty list. Other than that the logic should be the same. Having a little trouble testing but hopefully will be able to this afternoon. 

I'm still working on the widget sizing because its a little wonky on the rigs.

@XX-Yin I think you know the most about the opto sessions so let me know if I missed any key functionality in the schema
- widgets will look different but shouldn't effect session
- [ ] 446/7


ben-at-allen (approved):
That's a pretty massive set of changes hehe.  A lot seem to be style changes which all look good (spacing between ='s, etc).  Is there a specific area that you want me to look over that you think might be more volatile?  It looks good just from a cursory check. 
 
I did look through the new widget (opto_parameters_widget.py) more in-depth but that looks good too.  Without knowing much about the base class (SchemaWidgetBase), which seems to have some very specific controls referenced, I won't know if there's anything wrong in this child class if there's an error directly related to that inheritance.  Resolving widgets by generating their names from loops is a little dicey where a collection of widgets might be better (a QT Supported collection like a groupbox or a parent holder widget or just a manual collection like a dict or list) but it looks like that practice might be an artifact of that base class and how its expected to be used.  So I trust that you're probably following some pattern that is common when inheriting from that base widget.  But in general, in any UI app, if renaming a control with Ctrl+R throughout the project creates a bug, or if a UI element has to have a special "format" for its name because something else depends on it to conform to that format, is bad practice.
 
like: setattr(self, f"laser_colors.{laser}_check_box", check_box)
 
Great job on typing and consistency as well throughout.
 
Just a quick verification that this app is not suppose to pull from ZK for configs or log to SIPE's logserver as of yet correct?

micahwoodard:
Created checks to make sure waveforms are equal. Here are the resultant plots to make it easier but tests will also run. 

Sine waves with both lasers:
image

sine wave laser 1:
image

sine wave laser 2:
image

sine wave laser 2 offset:
image

pulse with both lasers:
image

pulse with laser 1:
image

pulse with laser 2: 
image

pulse with start offset:
image

pulse with 10 frequency:
image

constant both lasers: 
image

constant laser 1:
image

constant laser 2:
image

constant with offset:
image

constant with rampdown 2:
image

PR #1397, opened 2025-02-18
adding in fip model

micahwoodard:
  
### Describe changes:
Outline major fip changes here. Widget now looks like this:

ForagingGUI:
image

Ephys gui: 
image

There is a check box to enable/disable fip instead of a combo box. The logic instead of using the combobox is now to check if the fip_model.mode is set to None or not. Other than that, logic should be exactly the same.  This refactor is pretty minimal. 

- [ ] 446/7

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2458,8 +2458,6 @@ class Window(QMainWindow):
                     json.dump(Obj2, outfile, indent=4, cls=NumpyEncoder)
             elif self.SaveFile.endswith('.json'):
                 with open(self.SaveFile, "w") as outfile:
-                    from pprint import pprint
-                    pprint(Obj)
                     json.dump(Obj, outfile, indent=4, cls=NumpyEncoder)
 
         # Toggle unsaved data to False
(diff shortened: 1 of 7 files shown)


micahwoodard:
@XX-Yin I was wondering if you had any suggestions on how to test that the optogenetics  is working how it's supposes to? On this branch, I can run a session with lasers configured and not get any errors, but I wanted to make sure that the lasers are doing what I think they are doing

XX-Yin:
Typically, I turn on the laser or LED and configure the optogenetics parameters to verify whether the outputs match expectations.

@micahwoodard @alexpiet @hanhou 

To be honest, I have some concerns about the updates related to the behavior model—not just limited to this PR. I know @micahwoodard has put a lot of effort into these updates, and I truly appreciate that. However, I do have a few reservations:

1. I'm not sure if this is the best and simplest way to add the new curriculum. This logic used here seems to introduce additional complexities.
2. There’s a risk of hidden bugs going undetected. Our current codebase has undergone extensive testing. but these updates are quite significant, and I’m unsure how to comprehensively validate them.
3. The changes might disrupt our current UI, which is built with Qt Designer.

alexpiet:
This seems like a good update to me.

PR #1403, opened 2025-02-20
Add data_lock to calculate bias thread

### LINKED ISSUE #1394: startTimer cannot create timer in thread
> i think this may have caused another issue
> 
> my session randomly closed out and this is the message on the GUI terminal
> 
> Image 

 _Originally posted by @mlee10201 in [#926](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/926#issuecomment-2666745068)_

micahwoodard:
  
### Describe changes:
- adding data_lock to bias calculation to try and stop threading errors
- resolves #1394 
- None
- None
- [ ] 446/7
- No

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4388,12 +4388,10 @@ class Window(QMainWindow):
                     # add data to bias_indicator
                     if not self.bias_thread.is_alive():
                         logger.debug('Starting bias thread.')
-                        self.bias_thread = threading.Thread(target=self.bias_indicator.calculate_bias,
+                        self.bias_thread = threading.Thread(target=self.calculate_bias,
                                                        kwargs={'trial_num': len(formatted_history),
                                                                'choice_history': choice_history,
-                                                               'reward_history': np.array(any_reward).astype(int),
-                                                               'n_trial_back': 5,
-                                                               'cv': 1})
+                                                               'reward_history': np.array(any_reward).astype(int)})
                         self.bias_thread.start()
                     else:
                         logger.debug('Skipping bias calculation as previous is still in progress. ')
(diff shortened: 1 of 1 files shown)


XX-Yin:
@micahwoodard Thank you for adding this. Could you also add the data_lock here ? https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1403/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aL4507

alexpiet:
Makes sense to me. I don't know anything about data locks, so I'll defer to Xinxin

XX-Yin:
@micahwoodard @alexpiet 
This change looks good to me. I will approve it, and I think we can merge it to the develop and the production_testing this week after a brief test. 

There is a potential issue when using the data_lock across multiple threads, as the second code block will be blocked and wait for the first one to finish and release the lock. In our case, this could increase the ITI due to the running time of other threads, except for the **_GetAnimalResponse** . A better solution would be to move the data_lock inside the **_GetAnimalResponse** , where the data is modified, since this is the time-limiting step. We can submit a separate PR to optimize the entire threading process.

*****************************************************
Here is a short summary how data_lock works:

1. A data_lock is a mutex (mutual exclusion lock). When a thread enters with data_lock, it acquires the lock and no other thread can do the same until the lock is released.
This guarantees that only one thread at a time can safely read or write shared data.
2.When the thread exits the with block, it releases the lock, letting other threads proceed in turn.
3. If any thread accesses the shared data without using with data_lock:, you risk race conditions (corrupted or inconsistent data).
4. When a thread tries to enter the with data_lock: block and the lock is already held by another thread, the second thread cannot run the code inside the with data_lock: block until the first thread releases the lock.

XX-Yin (approved):

XX-Yin:
> @micahwoodard @alexpiet This change looks good to me. I will approve it, and I think we can merge it to the develop and the production_testing this week after a brief test.
> 
> There is a potential issue when using the data_lock across multiple threads, as the second code block will be blocked and wait for the first one to finish and release the lock. In our case, this could increase the ITI due to the running time of other threads, except for the **_GetAnimalResponse** . A better solution would be to move the data_lock inside the **_GetAnimalResponse**  , **bias check** and **back up save**, where the data is modified, since this is the time-limiting step. We can submit a separate PR to optimize the entire threading process.
> 
> Here is a short summary how data_lock works:
> 
> 1. A data_lock is a mutex (mutual exclusion lock). When a thread enters with data_lock, it acquires the lock and no other thread can do the same until the lock is released.
>    This guarantees that only one thread at a time can safely read or write shared data.
>    2.When the thread exits the with block, it releases the lock, letting other threads proceed in turn.
> 2. If any thread accesses the shared data without using with data_lock:, you risk race conditions (corrupted or inconsistent data).
> 3. When a thread tries to enter the with data_lock: block and the lock is already held by another thread, the second thread cannot run the code inside the with data_lock: block until the first thread releases the lock.

micahwoodard:
Tested it this morning and seems to be running fine. I did a test with and without simulating responses and nothing seemed to be blocking but agree would be nice to refactor threading

XX-Yin:
Great. Then it’s ready to merge

alexpiet:
Thanks @XX-Yin, @micahwoodard for working to implement this. Lets see if this solves the timer issue

XX-Yin:
@alexpiet @micahwoodard The timer might be a separate issue from the crash. Let's see.

PR #1406, opened 2025-02-21
Sound button

### LINKED ISSUE #1324: Make button to easily play sound for checking audio levels

micahwoodard:
  
### Describe changes:
- Add button on toolbar to trigger beep
- resolves #1324 
- Button in toolbar to trigger beep 
None
- [x] 446/7
- No


micahwoodard:
Here's a picture of the behavior gui. The button is a speaker in the tool bar 
image

and Ephys:
image

XX-Yin:
@micahwoodard Looks good! 

1. Can you disable the play sound button when the session is started (start button is checked) to avoid interaction with the behavior? 
2. Have you tested it by running the behavior (give sound first and then start the behavior)?

XX-Yin:
And the bonsai layout file has a lot of changes, can you make sure we make minimal changes for the layout file?

alexpiet:
What is the behavior of the button? Is it a toggle? (sound plays until you press again?) Or does it play for some interval?

micahwoodard:
> What is the behavior of the button? Is it a toggle? (sound plays until you press again?) Or does it play for some interval?
I made it not a toggle button so it emits only one go cue. I can ask RA's what they prefer though

micahwoodard:
> Can you disable the play sound button when the session is started (start button is checked) to avoid interaction with the behavior? 
> And the bonsai layout file has a lot of changes, can you make sure we make minimal changes for the layout file?

Yes can do!

> Have you tested it by running the behavior (give sound first and then start the behavior)?

Not yet people are still running mice. Will do probably around 3pm

src/foraging_gui/Foraging.py
```diff
@@ -2404,6 +2405,14 @@ def _Camera(self):
         else:
             self.Camera_dialog.hide()
 
+    def play_beep(self):
+        """
+        Convenience function to play tone
+        """
+
+        self.Channel3.TriggerGoCue(1)
+        #self.Channel3.receive() # clear buffer
```
XX-Yin, line 2414: I think you intuition here to receive the go cue timestamp is correct. It's Channel 1. Like example shown here
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/MyFunctions.py#L1719


I was not sure about this before, because the manual go cue trigger and the go cue from the behavior happen on different threads. I don't know if the timestamps related to the manual go cue trigger will be removed in a different thread. From the error you had, the answer is not removed.

micahwoodard: It stalled and crashed when I tried it so I thought there might be no messages to receive? Should I try with Channel1 instead?

XX-Yin: The sound card timestamp was sent through self.Channel (was renamed as Channel 1 in the MyFunctions.py). so just change self.Channel3.receive() to self.Channel.receive() should work.

micahwoodard: Oh sorry didn't read your first comment closely enough. Thanks for your help! It's been tested and ready to merge

micahwoodard:
The button plays once. Ella said it would probably be better for a continual tone so I'll update for that and re-request review before next week

micahwoodard:
Okay updated to play continual tone! Will test probably this afternoon

micahwoodard:
Okay tested in rig room and now will play continuous sound when button is pressed down

micahwoodard:
@XX-Yin @alexpiet ready for review

src/foraging_gui/Foraging.py
```diff
@@ -316,6 +316,11 @@ def connectSignalsSlots(self):
         '''Define callbacks'''
         self.action_About.triggered.connect(self._about)
         self.action_Camera.triggered.connect(self._Camera)
+
+        # create QTimer to deliver constant tone
+        self.beep_loop = QtCore.QTimer(timeout=self.play_beep, interval=10)
+        self.action_Sound.toggled.connect(lambda checked: self.beep_loop.start() if checked else self.beep_loop.stop())
+
```
XX-Yin, line 323: What does this continuous sound mean? Are the beeps with an interval?

alexpiet: What is the interval between beeps? We should verify that the measured sound level from this button matches the measured sound level from how it was being measured before.

micahwoodard: It's plays the beep every 10ms which is specified by the interval part of the QTimer. It's using the same bonsai node that the actual session uses so it should be the same tone. How was it being measured before and would @mlee10201 be the person to ask for help testing?

micahwoodard:
@alexpiet @XX-Yin Confirmed with @mlee10201 that the sound button works and plays correct tone. However, there is no way to tune or save settings within the gui which maybe helpful for calibration. That is not in the scope of [this ticke](https://apps.powerapps.com/play/e/9084dabc-d707-e1bf-8113-685be25cac15/a/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&ItemID=3963)t though so if that is the next step we should probably create a new ticket for that

micahwoodard:
If you guys are good, then yes ready to merge. 

We were confused what settings change the sound levels. Changing the Attenuation in the Settings_box.csv didn't seem to do anything and changing the go_cue_decibel only marginally changed the level. Is that supposed to be the case?

XX-Yin:
To be clear, 
**go_cue_decibel_box#** in the **foraging settings** file is only used for metadata. This value is the measured decibel 
**AttenuationLeft** and **AttenuationRight** in the **Settings_box#.csv** are used for changing the decibel. During the calibration, we changed the value to get the intended decibel.

src/foraging_gui/Foraging.py
XX-Yin: Is this intended "play the beep every 10ms"? I'm not sure if this will influence the measurement of the decibel. If this is for calibration/check purpose, can we increase the interval to like 5s?

micahwoodard:
Huh that's weird when @mlee10201 and I were testing, we changed AttenuationLeft and AttenuationRight in the Settings_box#.csv to a couple of different values, including 0 and nothing seemed to change. I can go test and confirm later today but that is not what is supposed to happen?

micahwoodard:
> > Huh that's weird when @mlee10201 and I were testing, we changed AttenuationLeft and AttenuationRight in the Settings_box#.csv to a couple of different values, including 0 and nothing seemed to change. I can go test and confirm later today but that is not what is supposed to happen?
> 
> Yes, changing Attenuation should change the volume. You would need to re-open the GUI to see an effect

Yeah I changed the csv file, saved, closed the gui, then reopened. I can double check, but I'm fairly certain that the changes to the csv file were saved when we reopened the gui

alexpiet:
@micahwoodard If you can replicate that behavior, see if it effects `main` or just your branch.

micahwoodard:
@alexpiet @XX-Yin when I set the AttenuationLeft and AttenuationRight to 0 in the Settings_box#.csv file, saved, and opened the gui, and ran a session, there was no noticeable difference in the sound cue volume to me. Here is a photo of the csv file. Is this the expected behavior? 
image

Update also tried 200 but no difference.

src/foraging_gui/Foraging.py
micahwoodard: It's to repeatedly play the beep at a fast enough speed so the sound is just a continuous beep. There is no pause anymore which is what @mlee10201 and @ellahiltonvano requested. It didn't effect measuring the decibels but if you think it will be a problem elsewhere, I can try and change the bonsai node

alexpiet (approved):
Since the attenuation issue is also on main, I figure we can merge this in this week

PR #1412, opened 2025-02-26
[skip email] Checking for saved sessions with no trials

alexpiet:
  
- https://github.com/AllenNeuralDynamics/dynamic_foraging_error_tracking/issues/4

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -3161,11 +3161,8 @@ class Window(QMainWindow):
 
                         # loading_parameters_type=0, get the last value of saved training parameters for each trial; 
                         # loading_parameters_type=1, get the current value for single value data directly from the window. 
-                        if 'TP_{}'.format(key) in CurrentObj:
-                            if len(CurrentObj['TP_'+key])< 2:
-                                # Adding debugging code for dynamic_foraging_error_tracking/issues/4
-                                logging.info('key {}: {}'.format(key,CurrentObj['TP_'+key]))
-                            value=np.array([CurrentObj['TP_'+key][-2]])
+                        if ('TP_{}'.format(key) in CurrentObj) and (len(CurrentObj['TP_{}'.format(key)]) > 1):
+                            value=np.array([CurrentObj['TP_{}'.format(key)][-2]])
                             loading_parameters_type=0
                         else:
                             value=CurrentObj[key]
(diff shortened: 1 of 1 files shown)


alexpiet:
I could not replicate the most recent instance of this happening because the offending session was already deleted. I think this PR will help up debug future instances.

src/foraging_gui/Foraging.py
```diff
@@ -3159,8 +3159,12 @@ def _Open(self,open_last = False,input_file = ''):
                             continue
                         widget = widget_dict[key]
 
-                        # loading_parameters_type=0, get the last value of saved training parameters for each trial; loading_parameters_type=1, get the current value for single value data directly from the window. 
+                        # loading_parameters_type=0, get the last value of saved training parameters for each trial; 
+                        # loading_parameters_type=1, get the current value for single value data directly from the window. 
                         if 'TP_{}'.format(key) in CurrentObj:
```
XX-Yin, line 3164: @alexpiet I think the [issue](https://github.com/AllenNeuralDynamics/dynamic_foraging_error_tracking/issues/4) might be associated we only have parameters associated with one trial available. 

I think this issue can be fixed by adding an extra check

if len(CurrentObj['TP_'+key])<= 2:
                            value=CurrentObj[key]
                            loading_parameters_type=1
else:
                            value=CurrentObj[key]
                            loading_parameters_type=1

alexpiet: I agree thats most likely the problem, I wasn't sure what `loading_parameters_type` does, so I wanted to log information before making the change.

What does `loading_parameters_type` do?

Do you mean this:
```
if len(CurrentObj['TP_'+key])<= 2:
    value=CurrentObj['TP_'+ key] # Note I changed this part from CurrentObj[key]
    loading_parameters_type=1
else:
    value=np.array([CurrentObj['TP_'+key][-2]])
    loading_parameters_type=0
```

XX-Yin: loading_parameters_type=1, it will use parameters extracted from the GUI widgets. 

loading_parameters_type=0, it will use the last training parameters (stored in TP_ ). The training parameters are generated before the running of the next trial in most cases, so here we use index -2.

It should like the code shown below
```
 if len(CurrentObj['TP_'+key])<= 2:
    value=CurrentObj[key]
    loading_parameters_type=1
else:
    value=np.array([CurrentObj['TP_'+key][-2]])
    loading_parameters_type=0
```

alexpiet: Ok, I've implemented this change.

alexpiet:
@XX-Yin @micahwoodard This is ready for final review

micahwoodard (approved):
Makes sense to me.  Is this check for if a TP_ is missing from the behavior.json?

PR #1413, opened 2025-02-27
fix the warmup

XX-Yin:
  
### Describe changes:
Excluding the empty widgets from text check. 

Fixed https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/840#issuecomment-2584676842

I can repeat this issue in 447 with the simulation function in the GUI. And found out that the problem was the empty widget introduced by the AIND stage triggering the text change and changing UpdateParameters to zero which would not allow the parameters to change.

No

No

Tested in 447
No


XX-Yin:
@alexpiet @micahwoodard @hanhou This is ready for review

alexpiet:
Thanks for getting to the bottom of this issue! 

@micahwoodard can you review? It looks good to me, but I'm less familiar with how widgets work

micahwoodard:
I don't really understand what's going on here either. @XX-Yin are you saying that the stage_widget widgets don't have a set objectName so they are now skipped in the  _CheckTextChange function? And that allows the parameters to be updated since  UpdateParameters is still 1 after the stage is moved? Also, do you know what recent changes to the stage_widget or code would have introduced this bug?

XX-Yin:
@micahwoodard It's some empty widgets triggered the text check, which will change the UpdateParameters to zero (zero means avoiding update of parameters). These empty widgets are from the AIND stage widgets, which is the reason we only see this issue in 447 (AIND stage widgets) instead of 446 (mostly new scale stage widgets). However, I don't know why AIND stage widgets have these empty widgets.

XX-Yin:
@micahwoodard @alexpiet Can we review and merge this tomorrow so we can merge this into the production_testing?

micahwoodard:
I still don't understand what caused this is the first place since it seems to be happening recently. I'm assuming there was a new release to stage_widget? Would be good to understand for future issues like this

micahwoodard (approved):

PR #1417, opened 2025-03-03
Hotfix/stagewidget issue

jessy-liao:
  
### Describe changes:

This PR implements the following two features: 
1. Stagewidget now initializes before bonsai. I suspect bonsai may be connecting to all devices in its code which is causing issues with the stage widget connecting. 
2. New alert box that shows up in addition to the exception box that shows up. This new alert box has a message for scientists to help add some additional information in this issue: #925. This will help me debug this issue since I still haven't been able to recreate it. 

#925 

On stagewidget errors, a new alert box should display prompting the user on what to do to help diagnose this issue. 

If possible, this PR shouldn't be distributed to all rigs. There should be a control group of rigs that is running without this pr and a testing group with this PR. This way if we continue seeing errors we can potentially rule out bonsai being the root cause. 

Yes

No

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -112,9 +112,6 @@ class Window(QMainWindow):
         # Load Rig Json
         self._LoadRigJson()
 
-        # Stage Widget
-        self.stage_widget = None
-
         # Load User interface
         self._LoadUI()
 
@@ -154,6 +151,10 @@ class Window(QMainWindow):
         self.unsaved_data = False   # Setting unsaved data to False
         self.to_check_drop_frames = 1 # 1, to check drop frames during saving data; 0, not to check drop frames
         self.session_run = False    # flag to indicate if session has been run or not
+
+        # Stage Widget
+        self.stage_widget = None
+        self._load_stage()
 
         # Connect to Bonsai
         self._InitializeBonsai()
@@ -207,7 +208,6 @@ class Window(QMainWindow):
         self._WaterCalibration()# to open the water calibration panel
         self._Camera()
         self._InitializeMotorStage()
-        self._load_stage()
         self._Metadata()
         self.RewardFamilies=[[[8,1],[6, 1],[3, 1],[1, 1]],[[8, 1], [1, 1]],[[1,0],[.9,.1],[.8,.2],[.7,.3],[.6,.4],[.5,.5]],[[6, 1],[3, 1],[1, 1]]]
         self.WaterPerRewardedTrial=0.005
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -154,7 +154,20 @@ class Window(QMainWindow):
 
         # Stage Widget
         self.stage_widget = None
-        self._load_stage()
+        try:
+            self._load_stage()
+        except IOError as e:
+            logging.error(e)
+            logging.error(f"ERROR..."
+                          f"Dear scientist, please perform the following to document this issue:"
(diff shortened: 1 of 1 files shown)


micahwoodard (approved):
This looks good to me!

src/foraging_gui/Foraging.py
```diff
@@ -155,6 +152,23 @@ def __init__(self, parent=None,box_number=1,start_bonsai_ide=True):
         self.to_check_drop_frames = 1 # 1, to check drop frames during saving data; 0, not to check drop frames
         self.session_run = False    # flag to indicate if session has been run or not
 
+        # Stage Widget
+        self.stage_widget = None
```
alexpiet, line 156: @jessy-liao Our code automatically opens github issues for unique error strings, so if you just do:
```logging.error('stage widget failure <unique identifier for session/rig/mouse>')```

Then the error will open with the log attached, and all the information you requested. This will get you more information than asking a user to document things

File: src/foraging_gui/Foraging.py [hunk 2]
```diff
@@ -4977,6 +4990,16 @@ def log_git_hash():
     return git_hash, git_branch, repo_url, repo_dirty_flag, dirty_files, version
 
 
+def show_msg_box(window_title, title, msg):
```
alexpiet, line 4993: Seems odd to make a function that is only used once. I would prefer to not make a function here. But can you at least add a doc string

alexpiet (changes requested):
- I highly recommend using the automatic issue opening approach instead of asking users to document
- optional: clean up show_msg_box() function

jessy-liao:
**Could you elaborate on why you want this update only on some rigs?**

I guess it doesn't matter if its partially deployed or fully deployed. I mentioned partially deployed because of the show_msg_box function which is not something I would want added to the production code. It is more of a one-off to help me debug.

**Code automatically opens github issues for unique error strings. Should we use automatic issues?**
The idea behind user response is to catch what was going on before the gui is restarted. Example: (user does something, exits gui, starts gui - error occurs). 
Does the auto issue send logs from the last session of the gui? My intentions with this is if the user can provide the last 2 log files I could perhaps see if a pre-existing error (or a preexisting exit step) is leading to stage widget crashing at the next restart. 

**Clean up show_msg_box() function**
I added this one-off show_msg_box function to help notify users to help me debug. 

If we stick to auto issues then we can remove this OR...

I could make the existing show_exception_box more general so I can use it to display custom messages as well.

alexpiet:
> **Code automatically opens github issues for unique error strings. Should we use automatic issues?** The idea behind user response is to catch what was going on before the gui is restarted. Example: (user does something, exits gui, starts gui - error occurs). Does the auto issue send logs from the last session of the gui? My intentions with this is if the user can provide the last 2 log files I could perhaps see if a pre-existing error (or a preexisting exit step) is leading to stage widget crashing at the next restart.

It will not include the log from the previous session in the auto-issues. That makes sense to have users give you that information. I would clarify the message to the user that you want information on the previous session. That is not clear from your current message. I would also still use the auto-issue approach to automatically log as much information as possible. 

> I could make the existing show_exception_box more general so I can use it to display custom messages as well.

Don't spend much time on this. I would just add a doc strong, and a note that this is temporary code that will be removed when the stagewidget issue is closed (link to appropriate issue or SIPE ticket)

jessy-liao (commented):
Review Updates:

- Changed message to scientist to clarify that I want the last two log files. 
- Added docstring for show_msg_box

alexpiet (approved):

PR #1428, opened 2025-03-11
Updating Project Name information

### LINKED ISSUE #1427: Upload manifest always has "Behavior Platform" as project name
- All QC reports are showing up under "Behavior Platform"
- QC reports inherit this from data_description.json, which in turn inherits from the upload manifest
- The upload manifest uses `self.project_name`
- I believe the problem is that after getting the approved list of project names, we try to set the Metadata_dialog.ProjectName index. This index is based on an old list of project names, so it always fails and we use the default project name.

To fix:
- We need to have the ability to manually set the project name because of ephys rigs
- Update the list of project names in the GUI to use the dynamically generated AIND project names rather than the static csv file
- Throw an error if we can't set the project name in the metadata
- Remove CSV of project names, since its outdated
- Remove Data description fields of metadata UI

alexpiet:
  
### Describe changes:
- Previously, the list of project names was gathered from a static CSV file. 
- Recently, aind-metadata-service has provided a list of project names that are strictly enforced
- Because the schedule was using the aind-metadata-service project names, but the UI was using the static CSV list, it was causing a mis-match, and all sessions were being uploaded with "Behavior Platform" which is the default. 
- This PR updates the UI to use the AIND-metadata-service list of project names, instead of the static csv file on an institute server. 
- Removes unused metadata labels for funding source, fundee, grant number. These are tracked via project names, so we don't need to set them
- Users can still manually set project names using a combo-box
- Removes unused functions in GenerateMetadata.py

- resolves #1427 

- The metadata dialog box has some fields removed, and the project list will be up to date
Capture

- The setting `project_info_file` in `ForagingSettings.json` is being depreciated. Remove it if you have set it. I don't think any rigs have set this manually. 

- tested

- no

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -2016,7 +2016,7 @@ class MetadataDialog(QDialog):
         '''get the widgets used for saving/loading metadata'''
         exclude_widgets=self._get_children_keys(self.Probes)
         exclude_widgets+=self._get_children_keys(self.Microscopes)
-        exclude_widgets+=['EphysProbes','RigMetadataFile','StickMicroscopes']
+        exclude_widgets+=['EphysProbes','RigMetadataFile','StickMicroscopes', 'ProjectName']
         widget_dict = {w.objectName(): w for w in self.findChildren(
             (QtWidgets.QLineEdit, QtWidgets.QTextEdit, QtWidgets.QComboBox))
             if w.objectName() not in exclude_widgets}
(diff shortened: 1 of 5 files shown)


alexpiet:
@hanhou @micahwoodard @XX-Yin This is ready for review

alexpiet:
Just kidding, not ready to test. It sets the project name correctly, but the upload manifest doesn't work

alexpiet:
@hanhou @micahwoodard @XX-Yin This is ready for review again, I've tested it.

alexpiet:
@XX-Yin @micahwoodard @hanhou A reminder to review this PR

alexpiet:
@XX-Yin @micahwoodard @hanhou Please review this, so we can include it in this weeks testing cycle

micahwoodard (approved):
This one is going over my head a bit because I haven't worked with this part of the code extensively but nothing looks overtly wrong.

Issue #1414, opened 2025-02-27
AttenuationLeft and AttenuationRight Problem


micahwoodard:
I set the AttenuationLeft and AttenuationRight to 0 in the Settings_box#.csv file, saved, and opened the gui, and ran a session, there was no noticeable difference in the sound cue volume to me. Also tried 200 and also 2000 with no change. Confirmed that the values are being set correctly in bonsai but doesn't seem to be doing what intended. Detail in #1406

XX-Yin:
I found when I start the behavior GUI after removing the "--no-bonsai-ide" (the bonsai window will open), the sound decibel will change with the AttenuationLeft and AttenuationRight. However, when I start the behavior GUI with the "--no-bonsai-ide" (the bonsai window will not open), it may require multiple times of close and open of the GUI to make the AttenuationLeft and AttenuationRight change effective. 

I think there is something related to the update/initialization on the bonsai side when there is (or no) bonsai window open. I can change the code to make the value update more robust.

alexpiet:
Note you can use the desktop links for "Box.bat" (has --no-bonsai-ide) and "Configure Box.bat" (does not have the --no-bonsai-ide) for easy testing.

That being said - its very strange that the IDE window would make a difference! Thanks for documenting this @micahwoodard

alexpiet:
To be clear @micahwoodard this happens on the `main` branch?

micahwoodard:
@alexpiet yes this was on the main branch. @XX-Yin was also there with me yesterday afternoon to confirm

ellahiltonvano:
@alexpiet @micahwoodard if we need to recalibrate the gocue this week can that be done? Or do we need to wait until this issue is resolved?

micahwoodard:
@ellahiltonvano I think we'll have to wait until this is fixed. Although the behavior seems inconsistent so I'm not sure

XX-Yin:
@micahwoodard @ellahiltonvano I have updated the code and will test it this afternoon!

XX-Yin:
It works. I have a PR waiting for review
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1429

ellahiltonvano:
@alexpiet does this change anything for Bruno's investigation?

micahwoodard:
[...]

PR #1429, opened 2025-03-13
Update go cue

XX-Yin:
  
### Describe changes:
Update the bonsai code to set the attenuation value each trial (using GoCue to trigger the update). 
image

Resolved https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1414#issuecomment-2719041592

The issue is when we start the bonsai without the ide by adding --no--bonsai-ide to the bat file, the go cue decibel is not updated with the change of the AttenuationLeft and AttenuationRight values in the Settings_box#.csv, but it works when starting with the bonsai ide. 

In this update, the attenuation value is forced to be updated each trial. 

No

No

Tested in 447 8D

No


XX-Yin:
@micahwoodard @alexpiet @hanhou This is ready for review

alexpiet:
Why does using the IDE change the behavior of how settings are applied? Are other settings impacted by not using the IDE?

Are you saying that the Attenuation parameters are set correctly when starting Bonsai (without the IDE), but changing the parameter while bonsai is already running does not change the value in Bonsai (without the IDE)? That was, to me, the expected behavior.

XX-Yin:
"Why does using the IDE change the behavior of how settings are applied? Are other settings impacted by not using the IDE?" 

I have no idea about why IDE matters. 
------
The parameter should be read from the settings file and set when we open the bonsai workflow. We previously didn't change the parameter when the bonsai is running. In this update, I forced the parameter to be updated each trial to make sure it's correctly set.

XX-Yin:
Other parameters including COM ports, camera serial numbers, codec and running wheel, current box are not affected by the IDE.

alexpiet:
@bruno-f-cruz Do you have any insight into why this parameter would not be set correctly when we do not use the IDE? I'd like to understand the problem before using a quick fix.

src/workflows/foraging.bonsai
```diff
@@ -1530,7 +1530,7 @@
                       <harp:VisualIndicators>On</harp:VisualIndicators>
                       <harp:Heartbeat>Disabled</harp:Heartbeat>
                       <harp:IgnoreErrors>false</harp:IgnoreErrors>
-                      <harp:PortName>COM12</harp:PortName>
+                      <harp:PortName>COM5</harp:PortName>
                     </Combinator>
```
micahwoodard, line 1534: Does this com port number matter? Is it updated for each box?

alexpiet: Its over-written by the COM ports in the settings file

micahwoodard:
@XX-Yin what rig did you test this on? I was testing on 8C this morning and was still seeing this behavior but I'm wondering if there is something wrong with 8C specifically since that's the rig I was using last time

XX-Yin:
> @XX-Yin what rig did you test this on? I was testing on 8C this morning and was still seeing this behavior but I'm wondering if there is something wrong with 8C specifically since that's the rig I was using last time

I tested it on 8D.

XX-Yin:
Sounds good! The changing logic is the same!

PR #1430, opened 2025-03-13
Attenuation fix

### LINKED ISSUE #1414: AttenuationLeft and AttenuationRight Problem

micahwoodard:
  
### Describe changes:
- changed bonsai workflows to set attenuation after initialization of sound device 
image

- resolves #1414 
- None
- None

- [ ] 446/7

### DIFF
--- a/src/workflows/foraging.bonsai
+++ b/src/workflows/foraging.bonsai
@@ -1536,8 +1536,13 @@
                   <Expression xsi:type="rx:BehaviorSubject">
                     <Name>SoundCardEvents</Name>
                   </Expression>
-                  <Expression xsi:type="SubscribeSubject">
-                    <Name>SoundCardCommands</Name>
+                  <Expression xsi:type="rx:BehaviorSubject">
+                    <Name>SoundCardEvents</Name>
+                  </Expression>
+                  <Expression xsi:type="Combinator">
+                    <Combinator xsi:type="rx:Delay">
+                      <rx:DueTime>PT2S</rx:DueTime>
+                    </Combinator>
                   </Expression>
                   <Expression xsi:type="Combinator">
                     <Combinator xsi:type="rx:Take">
@@ -1569,8 +1574,13 @@
                   <Expression xsi:type="MulticastSubject">
                     <Name>SoundCardCommands</Name>
                   </Expression>
-                  <Expression xsi:type="SubscribeSubject">
-                    <Name>SoundCardCommands</Name>
+                  <Expression xsi:type="rx:BehaviorSubject">
+                    <Name>SoundCardEvents</Name>
+                  </Expression>
+                  <Expression xsi:type="Combinator">
+                    <Combinator xsi:type="rx:Delay">
+                      <rx:DueTime>PT2S</rx:DueTime>
+                    </Combinator>
                   </Expression>
                   <Expression xsi:type="Combinator">
                     <Combinator xsi:type="rx:Take">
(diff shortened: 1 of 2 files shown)


micahwoodard:
@bruno-f-cruz I'm still getting inconsistent results with this workflow for setting the attenuation via a csv file. Is there something immediately obvious to you that I did wrong from what we talked about yesterday?

bruno-f-cruz:
Nop. What you did looks correct to me and, assuming I understand the rest of the workflow, should set the register to the value in the CSV. Have you looked at the logged harp data to check if you see the corresponding write message being logged? May be worth just doing what I tried to avoid in the first place and just do a live debug session at the rig.

micahwoodard:
Okay thank you for this input. I will try this afternoon when the rigs are free

micahwoodard:
Updating from convo with Bruno, the fix pushed should work but because the firmware of the soundboard is very old, there are some bugs when sending multiple commands so the attenuation was probably lost. Need to update with delay for fix to work. Longterm fix would be to update the firmware

bruno-f-cruz:
To add a bit more detail:
1. In the workflow micah sent previously, the subjects that trigger the command should be SoundCardEvents and not SoundCardCommands. 
2. Even after fixing 1., the behavior of the board did not look deterministic. I looked around a bit and saw that the core fw version of the board is 1.4, which is VERY old. Specifically, prior to 1.11, concurrent messages to the board were known to be randomly dropped. If you want to keep using 1.4, please make sure that all messages sent are acknowledged online (you can do this by waiting for the echo of the write message), or that you stagger messages with some arbitrary small delay. I would encourage you to also look at the read dumps previously recorded to make sure what attention sessions ran with because I don't think that there is any guarantees given the previous workflow and firmware combination.

alexpiet:
Thanks @bruno-f-cruz. If we upgrade the firmware on the harp boards, do you (1) have a recommended version to pin to? (2) expect any breaking changes to our bonsai workflow?

XX-Yin:
> Thanks @bruno-f-cruz. If we upgrade the firmware on the harp boards, do you (1) have a recommended version to pin to? (2) expect any breaking changes to our bonsai workflow?

@alexpiet Please check if we can get the sound and let the behavior running after updating the firmware. Previously, there are different issues in different versions of soundcard firmware. One version can get sound, but can't get the event timestamp. The other version can get the event timestamp, but no sound.

bruno-f-cruz:
I think the latest version is this one: https://github.com/harp-tech/device.soundcard/releases/tag/fw2.2-harp1.13

To @XX-Yin point it is possible that some interfaces changed. However, for harp I would encourage people to try to stay up-to-date with the latest releases, within reason (it should be noted that the 1.13 release its from Aug 10, 2023). The ecosystem is small, and everything is published "as is", bugs are often detected and promptly fixed in new firmware versions. This is especially important when it comes to versions of the core. 
The way I see it, you have a few options:
1. Keep everything as is and run offline QC on the read dump of the boards to make sure that settings were correct
2. Update the firmware to latest version, and possibly update the workflow logic if the board interface changed
3. Update the Bonsai workflow to mitigate the problem with the current firmware (e.g. make sure write messages are acknowledged by the board).

I believe 2. to be the best long-term solution. Yes, it will require periodic maintenance, but at the same time allows us to keep using the latest features/bug fixes from the ecosystem (for instance, the[ upcoming 1.14 ](https://github.com/harp-tech/core.atxmega/releases/tag/1.14-preview3)core release shall include: https://github.com/harp-tech/protocol/issues/48
https://github.com/harp-tech/protocol/issues/62
https://github.com/harp-tech/protocol/pull/45). 

Short-term solution starts with 1. and, if qc is failing often, implement 3.

alexpiet:
Thanks @bruno-f-cruz. I made a separate ticket to consider firmware updates (https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1435)

@micahwoodard Do you have a quick fix to implement today?

micahwoodard:
New audio workflow: 
image

micahwoodard:
Okay @alexpiet @XX-Yin I tested this in 446 2C and sound was changing based on settings.csv number

alexpiet (approved):

PR #1436, opened 2025-03-14
Widget to change attenuation of go cue

micahwoodard:
  
### Describe changes:
- sound button now has a drop down allowing users to change attenuation level of go cue. Changing the level in the gui will also save in settings file. 
- resolves [#986](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/986)
- allows user to change attenuation 
- None
- [ ] 446/7
No


micahwoodard:
Here is photo of the widget 
image

micahwoodard:
@ellahiltonvano does this widget for the sound look okay? Users can use the computer mouse to scroll the slider value up and down by hovering on it and the volume will update. I thought that would be better than a text edit but I can change if you think it would be better to let users input values! Also it's arbitrarily set to a maximum of 500 dbs. Let me know if that is not realistic!

ellahiltonvano:
@micahwoodard why is there a left and right value? Is it pretty easy to scroll to the exact value you want? Our range is 73 +/-2 dB so I'd say we don't want to be able to increase to 500 - maybe 100 for deaf mice? @ZhixiaoSu what value do you usually increase the tone to?

alexpiet:
Yes 500db is way too loud! 90 is even very high.

alexpiet:
I can't remember, but right now we are only use either left or right, correct? We should just have one slider that's sets both. Implementing left and right was just future proofing

alexpiet:
> @alexpiet @ellahiltonvano sorry I got confused this does not actually set the decibels, it sets the attenuation levels found in the settings_box csv file. Is there a way to set the decibels directly or would we need to have some sort of calibration curve?

Ah gotcha. We could probably calculate the mapping between attenuation and decibel, since I think it's just linearly scaling the sound file, and then decibels are a logarithmic scale. 

Grab some time on my calendar on Monday and we can do the math and test with a sound meter

micahwoodard:
> @micahwoodard why is there a left and right value? Is it pretty easy to scroll to the exact value you want? Our range is 73 +/-2 dB so I'd say we don't want to be able to increase to 500 - maybe 100 for deaf mice? @ZhixiaoSu what value do you usually increase the tone to?

I'll change to a spin box to you can enter a number or scroll. And I'll remove the right and left.

micahwoodard:
This is what the new bonsai code looks like. Still need to test but Bruno helped with it so hopeful. 
image

and this is what the gui looks like:
image

@ellahiltonvano does this widget work for ra's or can I change anything to make it better?

micahwoodard:
Okay tested in 447 and it works! @XX-Yin @galenlynch  this branch should work for calibration if you haven't already done it. Sorry it took me a while, I'm still learning bonsai. @ellahiltonvano I'm going to merge in now so I don't forget, but let me know if there are updates you would like and I can make them. As a note the spinbox value can be scrolled using the mouse, typed into, or you can use the arrow keys.

PR #1441, opened 2025-03-19
Checking for no generated trials when ending session

alexpiet:
  
### Describe changes:
I believe this error (https://github.com/AllenNeuralDynamics/dynamic_foraging_error_tracking/issues/266) is caused when an error happens during the photometry baseline period that doesn't crash the session, but prevents the normal generation of trials. Then the user stops the session, but the session end tasks can't run because the normal generation of trials didn't happen. This PR just checks for the existence of the GeneratedTrials before trying to access them. I tested this doesn't interfere with normal operations. 

- resolves https://github.com/AllenNeuralDynamics/dynamic_foraging_error_tracking/issues/266

- none

- none

- tested

- no

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4315,28 +4315,31 @@ class Window(QMainWindow):
         """
         Data cleanup and saving that needs to be done at end of session.
         """
+        if hasattr(self, 'GeneratedTrials'):
+            # If the session never generated any trials, then we don't need to perform these tasks
 
-        # fill out GenerateTrials B_Bias
-        last_bias = self.GeneratedTrials.B_Bias[-1]
-        b_bias_len = len(self.GeneratedTrials.B_Bias)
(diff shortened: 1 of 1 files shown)


XX-Yin:
@alexpiet Does this PR addressed the same issue as https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1443?

alexpiet:
> @alexpiet Does this PR address the same issue as #1443?

No, they are separate. This PR handles cleaning up a case where an error prevents trials from being generated and that in turn causes a new error

#1443 handles a case where a local variable is not set when loading

alexpiet:
Sorry I merged it without review, I didn't mean to do that. I think I got confused about which PR I was looking at

XX-Yin:
> Sorry I merged it without review, I didn't mean to do that. I think I got confused about which PR I was looking at

No worries!

PR #1455, opened 2025-03-24
Delaying upload time for sessions that start after 5:30pm

### LINKED ISSUE #1450: Delay upload time for sessions that start after 630pm
I had set the upload time as 830pm because I thought that was the latest we could be running mice. But mice can run later. So for mice that start after 630pm, we could delay the upload time to at least two hours after the session start time?

One problem with is that earlier sessions could start uploading in the middle of a late running session. I dont want to delay all upload times

alexpiet:
  
### Describe changes:
- Previously session upload time was always 8:30 plus a random offset. However, sessions that run late will not upload until the next day.
- now upload time for sessions that start after 5:30pm will be delayed to 3 hours after the start time (plus random offset)

- resolves #1450 

- Data should always upload the day it was run

- none

- [ ] tested

- no

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4871,11 +4871,17 @@ class Window(QMainWindow):
 
         logging.info('Generating upload manifest')
         try:
-            # Upload time is 8:30 tonight, plus a random offset over a 30 minute period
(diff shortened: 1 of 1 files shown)


micahwoodard (approved):
Makes sense and looks good to me

PR #1456, opened 2025-03-25
Fix drop of frame in fib session

XX-Yin:
  
### Describe changes:
Fixed drop of frame associated with the fiber session. 

- Improved Camera Trigger Handling in Logging:
The fiber session previously began logging before the behavioral session started and did not account for cases where the camera trigger remained active. In this update, the camera trigger on/off check has been moved into the logging function itself. This ensures the trigger status is verified every time logging is initiated.

- Fixed Logging Initialization Bug:
Resolved a potential issue where fiber photometry logging could start even when logging was already started. Logging will now only begin if it hasn’t already been started. This fix can also address the issue of duplicate session folders observed in some fiber sessions.

- Fixed drop of frame associated with the fiber session. 
- Fixed Logging Initialization Bug in the fiber session. 

No

No

Not yet. 

No


alexpiet:
@hagikent can you review this PR to make sure it addresses your concerns about fip logging?

src/foraging_gui/Foraging.py
```diff
@@ -1057,6 +1057,15 @@ def _restartlogging(self,log_folder=None):
...
         else:
             self._StopCurrentSession()
+
+        # Turn off the camera recording in case it is on
+        self.Camera_dialog.StartRecording.setChecked(False)
+        # Turn off the preview if it is on and the autocontrol is on, which can make sure the trigger is off before starting the logging. 
+        if self.Camera_dialog.AutoControl.currentText()=='Yes' and self.Camera_dialog.StartPreview.isChecked():
+            self.Camera_dialog.StartPreview.setChecked(False)
+            # sleep for 1 second to make sure the trigger is off
+            time.sleep(1)
```
alexpiet, line 1067: Do we need to sleep for a whole second? Thats a long delay. Is there a better way to check if the trigger is off?

src/foraging_gui/Foraging.py
```diff
@@ -3494,7 +3503,8 @@ def _StartFIP(self):
                 logging.warning('FIP workflow already started, user restarts',extra={'tags': [self.warning_log_tag]})
 
         # Start logging
-        self.Ot_log_folder=self._restartlogging()
+        if self.logging_type!=0:
```
alexpiet, line 3506: What change does this make? What is logging_type?

src/foraging_gui/Foraging.py
XX-Yin: 1 second seems to be ok and sufficient. It will only sleep 1s when the preview is started.

alexpiet:
@ellahiltonvano Can you test this PR?

ellahiltonvano:
@alexpiet I can test it with the current order of operations

hagikent:
It looks ok to me... but hard to picture how this behaves when combined with the restart-logger node in the Bonsai, which by default behaves in a complicated manner.

@XX-Yin 
Are you sure all FIP sync pulses generated before behavior starts are recorded properly as they are now?

@ellahiltonvano if timing allows, it would be safer if I can join testing at the rigs.

ellahiltonvano:
@hagikent if you want to stop by 447 at 2pm we could test then

XX-Yin:
@hagikent 

"Are you sure all FIP sync pulses generated before behavior starts are recorded properly as they are now?"

Yes

ellahiltonvano:
Tested with Kenta, did not see drop of frames. Worked as expected, but we only used a ~3min test session so if dropped frames are accumulating throughout the session our test would not have captured that.

XX-Yin:
@ellahiltonvano @hagikent Thanks! Sounds great! 3m is enough

alexpiet (approved):
Thanks for fixing this critical bug @XX-Yin

PR #1152, opened 2024-12-10
Metadata not saved

micahwoodard:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
- changing variables from type np.float64 to float type
### What issues or discussions does this update address?
- resolves [#799](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/799)
### Describe the expected change in behavior from the perspective of the experimenter
- metadata can now be saved without error on 6D
### Describe any manual update steps for task computers
- None
### Was this update tested in 446/447?
- [ ] 477 specifically on 6D

PR #1461, opened 2025-03-27
Hotfix 1461

### LINKED ISSUE #1394: startTimer cannot create timer in thread

galenlynch:
These changes fix a threading bug #1394 that is preventing me from doing experiments. @micahwoodard I do not understand the context of these changes, and what reverting your commits would do. Please feel free to commit to this branch as you see fit.

Fixes #1394


galenlynch:
Also feel free to replace this PR with one that addresses the actual root of the bug, if it is not this.

micahwoodard:
@galenlynch just trying to get caught up here

> what reverting your commits would do

What commits of mine are you reverting in this pr? I only see removal of typing the foraging efficiency as float which I believe @XX-Yin was the author of? He might be a better person to evaluate any downstream consequences of this. 


> Also feel free to replace this PR with one that addresses the actual root of the bug, if it is not this.

I think you are right that there was a computation bug somewhere that was being hidden behind all the try/excepts. If this is the fix that works for your bug, I'm betting that this is the root cause. I'm assuming the calculation for these variables returned a None type or something that can't be converted to a float. 


@alexpiet @XX-Yin kinda weird that the data locks worked in 446/7? Are we still thinking that it was a race condition solved by the locks or do we think it was caused by the bug Galen found but something to do with timing made it not appear?

galenlynch:
@micahwoodard Each of my commits is reverting one from your account, with your original commit referenced in each reversion. Not sure if you were the actual author, but it came from your account.

The original commits changed the data type from np.float to a python float. Notably dividing by zero with numpy floats produces a run-time warning, while dividing by zero with a python float produces a `ZeroDivisionError`.

alexpiet:
@micahwoodard I'm not sure if the data locks fixed it, or if its a mouse/session/curriculum specific issue and we just didnt have any more problem combinations. It was never a wide-spread issue that impacted all mice

galenlynch:
Actually it looks like you were converting to python float after the division...

galenlynch:
Looks like the commits were from #1152

micahwoodard:
I'm going to push a fix to account for the metadata bug in [#799](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/799) if that's cool and then I'm on board to approve this

galenlynch:
I can test on 323 ephys 2.

micahwoodard:
@galenlynch added a check for all np types so hopefully that _get_field function is a little more robust. Works on my laptop, but interested to see if it works for you

galenlynch:
@micahwoodard Your commits do not re-introduce the timer bug. However, on each trial I get a "metadata not saved!" warning in the GUI message area.

Is that expected? How do I fix that warning?
```
08:56:42 AM:ERROR:Foraging:Foraging.py:_Save:line 2844:Traceback (most recent call last):
  File "C:\Users\svc_aind_ephys\Documents\GitHub\dynamic-foraging-task\src\foraging_gui\Foraging.py", line 2825, in _Save
    generated_metadata=generate_metadata(Obj=Obj)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\GenerateMetadata.py", line 111, in __init__
    self._session()
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\GenerateMetadata.py", line 489, in _session
    self._get_stimulus()
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\GenerateMetadata.py", line 714, in _get_stimulus
    self._get_behavior_stimulus()
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\GenerateMetadata.py", line 745, in _get_behavior_stimulus
    output_parameters=self._get_output_parameters(),
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\GenerateMetadata.py", line 796, in _get_output_parameters
    'foraging_efficiency': _get_field(self.Obj, 'B_for_eff_optimal'),
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\svc_aind_ephys\.conda\envs\foraging-test\Lib\site-packages\foraging_gui\TransferToNWB.py", line 53, in _get_field
    if value in reject_list:
       ^^^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```
Log:
[W10DTB6VBSD9-A_gui_log_2025-03-27_08-55-11.txt](https://github.com/user-attachments/files/19489880/W10DTB6VBSD9-A_gui_log_2025-03-27_08-55-11.txt)

micahwoodard:
Posting here for visibility. So with changing the foraging efficient to np.float64, timer bug is solved but saving metadata doesn't work. Fixing the metadata bug reintroduces the timer bug. I'm wondering if there is a bug further down saving metadata and the metadata crashing is covering that up. Will continue to investigate.

XX-Yin:
> @galenlynch just trying to get caught up here
> 
> > what reverting your commits would do
> 
> What commits of mine are you reverting in this pr? I only see removal of typing the foraging efficiency as float which I believe @XX-Yin was the author of? He might be a better person to evaluate any downstream consequences of this.
> 
> > Also feel free to replace this PR with one that addresses the actual root of the bug, if it is not this.
> 
> I think you are right that there was a computation bug somewhere that was being hidden behind all the try/excepts. If this is the fix that works for your bug, I'm betting that this is the root cause. I'm assuming the calculation for these variables returned a None type or something that can't be converted to a float.
> 
> @alexpiet @XX-Yin kinda weird that the data locks worked in 446/7? Are we still thinking that it was a race condition solved by the locks or do we think it was caused by the bug Galen found but something to do with timing made it not appear?

I don't think the data_lock fixed the Timer issue, the data_lock is for fixing lagging.

micahwoodard:
Okay I think this startTimer error is not related to the QTimers. @XX-Yin was able to confirm by commenting all the QTimers out. I think its probably related to the clean up/completion of the Worker timer that is saving the metadata for each trial. If that thread crashes due to metadata generation error or the thread is commented out, we do not see these errors. I think next step we should maybe try using a different threading method? Or take a closer look at how the Worker and QThreadpool are being used. I can resume testing after lunch if the rig is still free

micahwoodard:
I actually don't have time to further debug this afternoon but will try tomorrow morning when I first get it. @galenlynch if you have more time today, it would be interesting to try and turn off saving every trial in the Setting.json and see if the error pops up again.

galenlynch:
Yesterday @XX-Yin commented out the saving in every trial in branch `temp_run` which seems to eliminate the issue.

micahwoodard:
@XX-Yin could you also achieve the same results by setting the `save_each_trial` entry in the Settings.json to false?

XX-Yin:
Micah, I believe so, which actually a better option without changing the code

alexpiet:
@micahwoodard @XX-Yin @galenlynch I haven't been following the details of this discussion - do we have a settled fix that should get merged in? Or should Galen continue to use his hotfix branch while we find a stable solution?

galenlynch:
OK, it's specifically related to generating metadata in the save function. According to QT documentation, threads outside the main thread should not interact with QObjects or widgets (except through slots), and I suspect that might be happening somewhere in the save function.

Commenting lines 2819 through 2832 (related to metadata) allows the data to be saved after every trial without the QTimer bug. One of those lines must somehow interact with the GUI state, which is not allowed.

@alexpiet  let's not merge yet.

galenlynch:
Yeah it accesses a bunch of widgets

galenlynch:
From my current understanding of #1394 I think some refactoring will be required, since threads other than the main one should _not_ access GUI state but currently the save worker directly gets the state of the metadata dialog by checking the value of each of its widgets.

@micahwoodard I think this PR has become unwieldy since it no longer fixes #1394. However, you have added a bunch of useful changes that might belong on a separate PR. Do you want me to create a new PR against `develop` and cherry pick these commits?

micahwoodard:
> From my current understanding of #1394 I think some refactoring will be required, since threads other than the main one should _not_ access GUI state but currently the save worker directly gets the state of the metadata dialog by checking the value of each of its widgets.

This could be solved by the curriculum refactor I'm doing. The task will be run by pydantic models instead of widget values and so saving them doesn't require checking widgets. However, this saving function still checks and saves widget values outside of the model to try and keep downstream processing the same. We could rethink this approach and only save values that are in a pydantic model and maybe that would solve the issue? 

> @micahwoodard I think this PR has become unwieldy since it no longer fixes #1394. However, you have added a bunch of useful changes that might belong on a separate PR. Do you want me to create a new PR against `develop` and cherry pick these commits?
Yeah that would be helpful! I think the relevant ones would be the nwb checking if the value is a np.array type and using the self.session_model.commit_hash instead of subprocess for the commit.

galenlynch:
Closing this PR in favor of #1470

PR #1468, opened 2025-03-30
Fix linter complaints

galenlynch:
I ran `ruff check --fix src/foraging_gui/*.py` which [fixes linter warnings that are safe to fix](https://docs.astral.sh/ruff/linter/#fixes)


galenlynch:
I also checked over the easily fixable 'unsafe changes' that made them.

alexpiet (approved):

alexpiet:
Thanks for doing that!

PR #1469, opened 2025-03-31
Adjust formatting to be like the rest of AIND

galenlynch:
@alexpiet for whatever reason your black output doesn't look like mine. I added black configuration to match the rest of AIND in the `pyproject.toml`  (taken from the library template), which changes the results of black.


alexpiet (approved):

alexpiet:
Thanks!

PR #1471, opened 2025-03-31
Adding subject ID when logging error about setting default platform

alexpiet:
  
### Describe changes:
- Updates error message to include subject ID
- resolves an edge case when "check_schedule" is True, but the mouse is not on the schedule

- improves the ability to track data

- none

- none

- [x] tested

- no

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -5587,10 +5587,12 @@ class Window(QMainWindow):
         ]
         index = np.where(np.array(projects) == "Behavior Platform")[0]
         if len(index) > 0:
+            logging.error('here 1')
             index = index[0]
             self.Metadata_dialog.ProjectName.setCurrentIndex(index)
             self.Metadata_dialog._show_project_info()
         else:
+            logging.error('here')
             self.Metadata_dialog.ProjectName.addItems([project_name])
         return project_name
 
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1622,12 +1622,7 @@ class Window(QMainWindow):
 
         # If we have a valid name update the metadata dialog
         if project_name is not None:
-            projects = [
-                self.Metadata_dialog.ProjectName.itemText(i)
-                for i in range(self.Metadata_dialog.ProjectName.count())
-            ]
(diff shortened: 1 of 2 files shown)


alexpiet:
@micahwoodard @hanhou @XX-Yin This is ready for review

src/foraging_gui/Foraging.py
```diff
@@ -5577,19 +5581,30 @@ def _set_default_project(self):
...
         ]
         index = np.where(np.array(projects) == "Behavior Platform")[0]
-        if len(index) > 0:
-            index = index[0]
-            self.Metadata_dialog.ProjectName.setCurrentIndex(index)
-            self.Metadata_dialog._show_project_info()
-        else:
+
+        # If not, add it
+        if len(index) == 0:
```
micahwoodard, line 5597: Some code simplification could be doing something like

```
if self.Metadata_dialog.ProjectName.findText("Behavior Platform") == -1:    # Behavior Platform is not in combobox
    self.Metadata_dialog.ProjectName.addItem("Behavior Platform")              # add Behavior Platform to combobox
```

not a huge deal just makes it a little more readable

src/foraging_gui/Foraging.py
```diff
@@ -5577,19 +5581,30 @@ def _set_default_project(self):
...
             self.Metadata_dialog.ProjectName.itemText(i)
             for i in range(self.Metadata_dialog.ProjectName.count())
         ]
         index = np.where(np.array(projects) == "Behavior Platform")[0]
-        if len(index) > 0:
-            index = index[0]
-            self.Metadata_dialog.ProjectName.setCurrentIndex(index)
-            self.Metadata_dialog._show_project_info()
-        else:
+
```
micahwoodard, line 5595: Tiny bit confusing to switch between "Behavior Platform" and project_name in a couple of places. Made me have to think a little harder and double check when reviewing but not a big deal.

micahwoodard (approved):
Some simplification that you can take or leave but makes sense and looks good to me

src/foraging_gui/Foraging.py
alexpiet: good catch

src/foraging_gui/Foraging.py
alexpiet: good catch

PR #1472, opened 2025-03-31
fix timer issue

XX-Yin:
  
### Describe changes:
Delete self.bias_indicator.clear(), which caused the timer issue. self.bias_indicator.clear() is redundant in the _Save function, as the _NewSession will run this code as well. 

 self.bias_indicator.clear() modifies a lot of widgets, which causes the timer issue when widgets are modified from a different thread. Generally, it's ok to get access to widgets from a different thread, but modifying widgets from different threads should be avoided. 
 
The Save button was checked in that specific session that we can replicate the issue.  

Fixes https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1394
No

No

Tested in 323_Ephys2

No


galenlynch:
I will run a test session on ephys rig 2! Thank you xinxin!

alexpiet:
Thanks for working on this!

Are you intentionally merging to "main" ?

XX-Yin:
> Thanks for working on this!
> 
> Are you intentionally merging to "main" ? There are so many linting updates its very hard to understand what changes this PR makes

Yes, I would like to make a hotfix to the main. There is one intended to the develop for record. https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1473

alexpiet:
@micahwoodard Any issue with removing the bias_indicator.clear() from every "_Save()" function call?

See isolated code here: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1473

micahwoodard:
No I don't think so. It should probably be called in _NewSession though to make sure it's up to date for the next session. I think that's the only reason why it's being called in the save function at all.

galenlynch:
@micahwoodard  it is already called in `_NewSession`:

https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/a3003e8b4a58dad2f34de7a9cc948d2de5f7cf5a/src/foraging_gui/Foraging.py#L3817

galenlynch:
@XX-Yin @alexpiet I'll make a copy of this branch off of the current main, to simplify the merge.

XX-Yin:
> @XX-Yin @alexpiet I'll make a copy of this branch off of the current main, to simplify the merge.

Sounds good

micahwoodard:
> @micahwoodard Any issue with removing the bias_indicator.clear() from every "_Save()" function call?
> 
> See isolated code here: #1473

I'm kinda confused as to why this would cause an issue at all in Galen's case because this is only called if the save button is clicked and that would not be the case in saving by trial

galenlynch:
I think the save button was clicked. The session loads with the save button highlighted, which might explain why the bug was session-specific.

galenlynch:
Can confirm that it fixed the bug.

Closed in favor of #1476

micahwoodard:
> I think the save button was clicked. The session loads with the save button highlighted, which might explain why the bug was session-specific.

Ah okay sorry for missing this!

PR #1473, opened 2025-03-31
Fix timer 3

XX-Yin:


XX-Yin:
a copy of https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1472, just for record

src/foraging_gui/Foraging.py
```diff
@@ -3998,7 +3998,6 @@ def _Save(self, ForceSave=0, SaveAs=0, SaveContinue=0, BackupSave=0):
                     ]
                 ):
                     self._AddWaterLogResult(session)
-                self.bias_indicator.clear()  # prepare for new session
```
micahwoodard, line 4001: @XX-Yin I'm kinda confused why this was causing errors in Galen's case because this is only called when the save button is clicked and that wouldn't be the case in the saving data per trial? Am I misunderstanding something?

galenlynch: The session was loaded with the save button depressed. Not sure it should be stateful like that, maybe that is a separate and related bug.

galenlynch:
I would suggest closing this in favor of #1477

src/foraging_gui/Foraging.py
XX-Yin: @micahwoodard Please see https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1472. 

In that specific session we can replicate this error, the Save button was checked. This is also why we can't easily replicate this issue.

micahwoodard: Ahh okay that makes sense thanks!

PR #1474, opened 2025-03-31
Fix more linter complaints

galenlynch:
After these changes the only linter complaints I'm getting are `E722` and `C901`, 'bare except' and 'function too complex' respectively.

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -3413,7 +3413,7 @@ class AutoTrainDialog(QDialog):
                 self.MainWindow,
                 "Box {}, Error".format(self.MainWindow.box_letter),
                 "AWS connection failed!\n"
-                "Please check your AWS credentials at ~\.aws\credentials and restart the GUI!\n\n"
+                "Please check your AWS credentials at ~/.aws/credentials and restart the GUI!\n\n"
                 "The AutoTrain will be disabled until the connection is restored.",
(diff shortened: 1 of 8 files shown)


src/foraging_gui/Dialogs.py
```diff
@@ -3413,7 +3413,7 @@ def _connect_auto_training_manager(self):
                 self.MainWindow,
                 "Box {}, Error".format(self.MainWindow.box_letter),
                 "AWS connection failed!\n"
-                "Please check your AWS credentials at ~\.aws\credentials and restart the GUI!\n\n"
+                "Please check your AWS credentials at ~/.aws/credentials and restart the GUI!\n\n"
```
alexpiet, line 3416: This is wrong. Since we only run on windows it will be '\' not '/'

alexpiet (approved):
Looks good other than the windows/linux path. Which doesn't really matter.

src/foraging_gui/Dialogs.py
galenlynch: `'\c'` is an invalid escape sequence. I changed it to `/` since i'm lazy. Would you prefer `\\c`?

galenlynch:
The issue was the invalid escape sequence, happy to use `'\\c'` if you prefer.

galenlynch:
@alexpiet I escaped the backslashes

PR #1477, opened 2025-03-31
Hotfix/qt timer bug

galenlynch:
#1476 but into `develop` instead of `main`


galenlynch:
I will resolve the merge conflict with `develop` if/when #1476 is merged.

XX-Yin (approved):

PR #1484, opened 2025-04-01
Feat 1447 slims stage coords

### LINKED ISSUE #1447: Stage coordinates tracked and saved to slims
Track last used stage coordinates in slims behavior model

micahwoodard:
  
### Describe changes:
- stage coordinates are stored in slims in the offset for mouse models.  When stage is moved, offset is updated if mouse is loaded. 
- resolves #1447 
- None
- None
- [x] 446/7
- None


micahwoodard:
Do we need this to work for the new scale stage too? If so we need to figure out a way to be able to tell what stage coordinate system the offest is in because newscale and aind stage have different ones.

alexpiet:
> Do we need this to work for the new scale stage too? If so we need to figure out a way to be able to tell what stage coordinate system the offest is in because newscale and aind stage have different ones.

I think it needs to, because the ephys rigs will continue to use the newscale stages

micahwoodard:
> I think it needs to, because the ephys rigs will continue to use the newscale stages

Should we have some sort of check to make sure the offset is being applied to the same stage since they will have different coordinate systems or just trust the users?

alexpiet:
> > I think it needs to, because the ephys rigs will continue to use the newscale stages
> 
> Should we have some sort of check to make sure the offset is being applied to the same stage since they will have different coordinate systems or just trust the users?

We should absolutely check, since the differences in coordinate systems are huge and could cause problems

micahwoodard:
Any ideas on how? We maybe could add an attachment to the mouse model that says what stage?  But also, does it make sense to store the offsets in slims if it's from the newscale stage? The way Alex Cahoon was describing the offset seemed like it's not stage coordinates but rather the ideal position for mouse licking from the monument. The newscale stage rigs don't have monuments set do they?

alexpiet:
> Any ideas on how? We maybe could add an attachment to the mouse model that says what stage? But also, does it make sense to store the offsets in slims if it's from the newscale stage? The way Alex Cahoon was describing the offset seemed like it's not stage coordinates but rather the ideal position for mouse licking from the monument. The newscale stage rigs don't have monuments set do they?

The stage will be rig dependent, so we don't want to store that on SLIMS. I think when we load the session on the rig we want to check what stage is on the rig. This logic already exists when determining whether to load the stage widget. The newscale stage does not have a monument. So I think the logic is

Newscale stage: position = saved position in the previous session
AIND stage: position = monument (rig specific) offset + mouse offset + saved position in the previous session

micahwoodard:
logic for loading stage positions:
- Store last used x, y, z coords in the operational control model as well as what stage these coordinates relate to
- Store aind stage coords used in slims as well
- if operational control stage used matches the rig stage, the use coordinate found in operational control
- If coords are read in and operational control stage is newscale but rig stage is aind stage, check slims for offset. If nothing on slims, do nothing.
- If coords are read in and operational control stage is aind stage but rig stage is newscale, do nothing 
- If coords are read in and operational control stage is newscale stage and rig stage is newscale but rig id is different, do nothing

PR #1494, opened 2025-04-09
Manually reverting ruff changes

alexpiet:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
- #1468 used ruff to fix linter complaints. This introduced bugs. 
-  I could not revert the commit since other commits had already been merged, and I didn't want to revert everything
- I believe the issues are caused by the following mistake
    - `1 == True` evaluates to True
    - `1 is True` evaluates to False
    - `1` evaluates to True
- I went through the commits in #1468 and corrected the instances where I believe this was causing a problem

### What issues or discussions does this update address?
- resolves #1492 
- resolves https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/1051

### Describe the expected change in behavior from the perspective of the experimenter
- none

### Describe any manual update steps for task computers
- none

### Was this update tested in 446/447?
- [ ] tested

### Does this update impact downstream processing by adding new saved files, or changing their format? If so, have you documented changes?
- none

PR #1497, opened 2025-04-10
Revert linter fixes from #1468

galenlynch:
This is an alternative to #1494 where I have completely reversed the contents of #1468. For what it's worth, I much prefer #1494.

### DIFF
--- a/src/foraging_gui/Dialogs.py
+++ b/src/foraging_gui/Dialogs.py
@@ -2,6 +2,7 @@ import json
 import logging
 import math
 import os
+import shutil
 import subprocess
 import time
 import webbrowser
@@ -17,7 +18,7 @@ from aind_auto_train.schema.task import TrainingStage
 from matplotlib.backends.backend_qt5agg import (
     NavigationToolbar2QT as NavigationToolbar,
 )
-from PyQt5 import QtWidgets, uic
+from PyQt5 import QtGui, QtWidgets, uic
 from PyQt5.QtCore import (
     QAbstractTableModel,
     QItemSelectionModel,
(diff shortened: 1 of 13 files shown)


galenlynch:
yes

src/foraging_gui/Dialogs.py
```diff
@@ -786,7 +789,7 @@ def _CalibrateLeftOne(self, repeat=False):
                     "All measurements have been completed. Either press Repeat, or Finished"
                 )
                 return
-            next_index = np.where(self.left_measurements is False)[0][0]
+            next_index = np.where(self.left_measurements != True)[0][0]
```
XX-Yin, line 792: They are not identical in some cases even most likely it's correct in this case.

src/foraging_gui/Dialogs.py
```diff
@@ -973,7 +976,7 @@ def _CalibrateRightOne(self, repeat=False):
                     "All measurements have been completed. Either press Repeat, or Finished"
                 )
                 return
-            next_index = np.where(self.right_measurements is False)[0][0]
+            next_index = np.where(self.right_measurements != True)[0][0]
```
XX-Yin, line 979: They are not identical in some cases even most likely it's correct in this case.

src/foraging_gui/Dialogs.py
```diff
@@ -1459,7 +1462,7 @@ def _SpotCheck(self, valve: Literal["Left", "Right"]):
                 "If this is a typo, please press cancel."
                 '<br><br><span style="color:purple;font-weight:bold">IMPORTANT</span>: '
                 "If the measurement was correctly entered, please press okay and repeat"
-                "spot check once.",
+                "spot check once.".format(np.round(result, 2)),
```
XX-Yin, line 1465: Is this correct here?

galenlynch: This is restoring the code to how it was before #1468

src/foraging_gui/Dialogs.py
galenlynch: This is restoring the code to how it was before #1468

alexpiet:
Closing because we instead merged: https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1494

Decision was made on teams

PR #1502, opened 2025-04-14
redo open function

### LINKED ISSUE #1489: Update open function
The open function ingest a behavior json and updates all widget found. The new load function should ingest the pydantic models and update accordingly

micahwoodard:
  
### Describe changes:
- rewrites loading function to load in task, session, fip, opto, and operational control model instead of behavior.json
- resolves #1489 
- user will select a folder that contains models instead of selection behavior json
- none
- [ ] 446/7
No


micahwoodard:
@alexpiet @XX-Yin Main thing to review is the _Open function. There are some other things that are small bug fixes

XX-Yin:
@micahwoodard Can we still load parameters and perform visualization from the behavior json? It would be better to ensure backward compatibility as we don't have multiple session model parameters for data collected so far.

micahwoodard:
> @micahwoodard Can we still load parameters and perform visualization from the behavior json? It would be better to ensure backward compatibility as we don't have multiple session model parameters for data collected so far.

Good point about the visualization, I believe I can add that in. I can also create a function to map parameters for backwards compatibility.

micahwoodard:
@XX-Yin could you point me to where exactly in the code the gui pulls information from the bahavior json and adds it to the graphs?

micahwoodard:
> @XX-Yin could you point me to where exactly in the code the gui pulls information from the bahavior json and adds it to the graphs?

Never mind think I might've found it. Is it _LoadVisualization?

XX-Yin:
> > @XX-Yin could you point me to where exactly in the code the gui pulls information from the bahavior json and adds it to the graphs?
> 
> Never mind think I might've found it. Is it _LoadVisualization?

Yes, it's _LoadVisualization

micahwoodard:
@XX-Yin @alexpiet should opened mice be treated like loaded mice? Should we write the next session to slims? I feel like no since we can't tell where they are in the curriculum based on the loaded files

micahwoodard:
@XX-Yin @alexpiet this is also ready for review. Gui will now update visuals and can load in old behavior jsons

XX-Yin:
@micahwoodard My understanding is if the "on curriculum is selected", the session should be written to slims no matter it's opened mice or loaded mice.

alexpiet:
@micahwoodard In general I don't think it should matter if the data was loaded from the local drive or from slims

micahwoodard:
I guess I have some questions about what the utility is for the Open function. Why would you open a mouse rather than load a mouse? Are we assuming that the session opened has already been written to slims? That's little weird because the last written session to slims should be what is run in the next session. How do we handle that? Is opening a mouse a continuation of the session? How do we relay that in slims if so?

alexpiet:
I assume the use case is something like SLIMS is down or internet access is down.

alexpiet:
I'm not sure I'm tracking the details of your question. I think I was imaging the mouse was already in SLIMS, but maybe SLIMS is down and we want to run the mouse by opening local data

micahwoodard:
Based on in person chat: 

Need two functionalities from loading local sessions

1. Open local sessions and load in behavior from this session. This will update GenerateTrials and plots and act as a continuation of this session. 
2. Open local sessions and create a new session using model/behavior json from local json to populate parameters. This will leave GenerateTrials and plots empty. 

I'm debating with whether or not to query slims for if mouse has session information written.  If anything, I think there could be a functionality that checks slims for the last written behavior session, deletes if found, uploads the session loaded and marks it off curriculum. That seems like we're assuming a lot though and could be problematic if we're envisioning this only to be used when slims is down. I will think harder and maybe ask Bruno what he thinks

@XX-Yin does this align with what you are thinking or do you have anything to add?

micahwoodard:
This pr does a couple things

1. Open local sessions and load in behavior from this session. This will update GenerateTrials and plots and act as a continuation of this session.
2. Open local sessions and create a new session using model/behavior json from local json to populate parameters. This will leave GenerateTrials and plots empty.
3. Allows sessions to run if gui is not connected to slims
4. Updates slims attachments with the json that actually ran during the session

Issue #1046, opened 2024-11-19
What is the lickspout stage "reference position"


alexpiet:
- This is only implemented for the NewScale stage, and not the AIND stage.
- What is it? How is it used?
- Do we need to maintain it?

I have a vague memory this is used in the AIND metadata, but I dont know in what manner. 

@XX-Yin @jessy-liao

ellahiltonvano:
@alexpiet not sure if this is helpful but I don't think any RAs have used this function.

micahwoodard:
@XX-Yin just trying to figure out how this is used to I can best update the functionality of track and set the aind stage coordinates. See #1028

XX-Yin:
@Micah ***@***.***>@Alex ***@***.***> This is used in the metadata to set the lickspout reference position. We move to the lickspout to a reference place (e.g. nose tip), then click the “set reference position” button to record this position.

We will discuss with Saskia today about the metadata. And this function might need to be updated. Will get feedback to you soon.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

micah:
I believe you have @'d the wrong person ***@***.*** is me, but I have no
idea what this repository is about).

XX-Yin:
> I believe you have @'d the wrong person ***@***.*** is me, but I have no idea what this repository is about).

I mistakenly tagged the wrong person. This seems to be an issue related to the email sent from my phone.

alexpiet:
Closing as redundant with https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/465

PR #1507, opened 2025-04-21
Habituation period before session 

micahwoodard:
  
### Describe changes:
- adds a QSpinBox with range 0 to 90. Start loop will wait specified time before starting session 
- resolves [#1046](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/1046)
- ability to specify wait time before session starts
- None
- [x] 446/7
No

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -5461,7 +5461,7 @@ class Window(QMainWindow):
 
         self.unsaved_data = False
         self.ManualWaterVolume = [0, 0]
-        self.baseline_min_elapsed = 0
+        self.baseline_min_elapsed = 0   # variable to track baseline time elapsed before session for start/stop
 
         # Clear Plots
         if hasattr(self, "PlotM") and self.clear_figure_after_save:
@@ -6296,6 +6296,7 @@ class Window(QMainWindow):
         elapsed = self.baseline_min_elapsed
         while self.baseline_min_elapsed < self.hab_time_box.value() and self.behavior_baseline_period.is_set():
             QApplication.processEvents()
+            # update baseline time elapsed before session for start/stop logic
             self.baseline_min_elapsed = ((time.time() - start_time) / 60) + elapsed
 
         update_hab_timer.stop()
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -6334,8 +6334,7 @@ class Window(QMainWindow):
         stall_iteration = 1
         stall_duration = 5 * 60
 
-        if self.Start.isChecked():
-            logging.info(f"Starting session.")
+        logging.info(f"Starting session.")
 
         while self.Start.isChecked():
             QApplication.processEvents()
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2761,6 +2761,7 @@ class Window(QMainWindow):
                         "laser_2_calibration_power",
                         "laser_1_calibration_voltage",
                         "laser_2_calibration_voltage",
+                        "hab_time_box"
                     }:
                         continue
                     if child.objectName() == "UncoupledReward":
@@ -6266,16 +6267,35 @@ class Window(QMainWindow):
             logging.info("No active session logger")
 
     def _StartTrialLoop(self, GeneratedTrials, worker1, worker_save):
-        if self.Start.isChecked():
-            logging.info("starting trial loop")
-        else:
+        if not self.Start.isChecked():
             logging.info("ending trial loop")
+            return
+
+        else:
+            logging.info("starting trial loop")
+
+        # pause for specified habituation time
+        start_time = time.time()
+
+        # create habituation timer label and update every minute
+        hab_timer_label = QLabel(f"Time elapsed: {(time.time() - start_time) / 60}")
+        hab_timer_label.setStyleSheet(f"color: {self.default_warning_color};")
+        self.warning_widget.layout().insertWidget(0, hab_timer_label)
+        update_hab_timer = QtCore.QTimer(
+            timeout=lambda: hab_timer_label.setText(f"Time elapsed: {(time.time() - start_time) / 60}"),
+            interval=60000)
+        update_hab_timer.start()
+
+        logging.info(f"Waiting {self.hab_time_box.value()} min before starting session.")
+        while time.time() - start_time < self.hab_time_box.value() * 60:
+            QApplication.processEvents()
+        update_hab_timer.stop()
+        logging.info(f"Starting session.")
 
         # Track elapsed time in case Bonsai Stalls
         last_trial_start = time.time()
         stall_iteration = 1
         stall_duration = 5 * 60
-
         while self.Start.isChecked():
             QApplication.processEvents()
             if (
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -6278,12 +6278,13 @@ class Window(QMainWindow):
         start_time = time.time()
 
         # create habituation timer label and update every minute
-        hab_timer_label = QLabel(f"Time elapsed: {(time.time() - start_time) / 60}")
+        hab_timer_label = QLabel()
         hab_timer_label.setStyleSheet(f"color: {self.default_warning_color};")
         self.warning_widget.layout().insertWidget(0, hab_timer_label)
         update_hab_timer = QtCore.QTimer(
-            timeout=lambda: hab_timer_label.setText(f"Time elapsed: {(time.time() - start_time) / 60}"),
-            interval=60000)
+            timeout=lambda: hab_timer_label.setText(f"Time elapsed: {round((time.time() - start_time) // 60)} minutes"
+                                                    f" {round((time.time() - start_time) % 60)} seconds"),
+            interval=1000)
         update_hab_timer.start()
 
         logging.info(f"Waiting {self.hab_time_box.value()} min before starting session.")
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -6158,6 +6158,21 @@ class Window(QMainWindow):
         :returns: boolean indicating if workflow is running or not. If None, fip is not configured
         """
 
+        if self.PhotometryB.currentText() == "on" and (
+                not self.FIP_started
+        ):
+            reply = QMessageBox.critical(
+                self,
+                "Box {}, Start".format(self.box_letter),
+                'Photometry is set to "on", but the FIP workflow has not been started',
+                QMessageBox.Ok,
+            )
+
+            logging.info(
+                "Cannot start session without starting FIP workflow"
+            )
+            return False
+
         # Check if photometry excitation is running or not
         if self.PhotometryB.currentText() == "on" and not self.StartExcitation.isChecked():
             logging.warning('photometry is set to "on", but excitation is not running')
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -5913,42 +5913,6 @@ class Window(QMainWindow):
                 )
                 return
 
-            # Check if photometry excitation is running or not
-            if self.PhotometryB.currentText() == "on" and (
-                not self.StartExcitation.isChecked()
-            ):
-                logging.warning(
-                    'photometry is set to "on", but excitation is not running'
-                )
-
-                reply = QMessageBox.question(
-                    self,
-                    "Box {}, Start".format(self.box_letter),
-                    'Photometry is set to "on", but excitation is not running. Start excitation now?',
-                    QMessageBox.Yes | QMessageBox.No,
-                    QMessageBox.Yes,
-                )
-                if reply == QMessageBox.Yes:
-                    self.StartExcitation.setChecked(True)
-                    logging.info("User selected to start excitation")
-                    started = self._StartExcitation()
-                    if started == 0:
-                        reply = QMessageBox.critical(
-                            self,
-                            "Box {}, Start".format(self.box_letter),
-                            "Could not start excitation, therefore cannot start the session",
-                            QMessageBox.Ok,
-                        )
-                        logging.info(
-                            "could not start session, due to failure to start excitation"
-                        )
-                        self.Start.setChecked(False)
-                        return
-                else:
-                    logging.info("User selected not to start excitation")
-                    self.Start.setChecked(False)
-                    return
-
             # disable sound button
             self.sound_button.setEnabled(False)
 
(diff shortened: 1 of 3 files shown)


micahwoodard:
What ephys gui looks like:

image

what behavior looks like: 

image

Also countdown label will show time elapsed in warning widget: 

image

micahwoodard:
@galenlynch @ellahiltonvano is the functionality of the pr what you were hoping for? It just delays the behavior session for however many minutes you have specified in the habituation box. That's how I understood the conversation and issue but lemme know if that is not the case.

@XX-Yin I put the logic in the Start loop function so it would be very generic and not need to have to keep track of a timer or anything. I don't think that will affect the photometry timer since it's downstream but let me know. Also didn't see any errors with saving thread but let m know if you see anything that would cause issues

ellahiltonvano:
@micahwoodard yep! Maybe title it 'Baseline'? Are we able to type in a number or just use the up/down arrows?

micahwoodard:
> @micahwoodard yep! Maybe title it 'Baseline'? Are we able to type in a number or just use the up/down arrows?

Sounds good and you're able to scroll the number up or down with the mouse, click the arrows, or type in a number to change. The range is 0 to 90 min. Let me know if it would be better to change that range.

src/foraging_gui/Foraging.py
```diff
@@ -6266,16 +6271,42 @@ def end_session_log(self) -> None:
...
+        if not self.Start.isChecked():
             logging.info("ending trial loop")
+            self.behavior_baseline_period.clear()
+            return
+
+        else:
+            logging.info("starting trial loop")
+            self.behavior_baseline_period.set()
+
+        # pause for specified habituation time
```
XX-Yin, line 6283: We only need to start the timer once for each session, which means we should check if this is a new session (Self.StartANewSession == 1).

XX-Yin:
It would be better to move the main update to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) (after the photometry timer) for easier organization.

src/foraging_gui/Foraging.py
```diff
@@ -6266,16 +6271,42 @@ def end_session_log(self) -> None:
...
+        start_time = time.time()
```
XX-Yin, line 6284: Could you package the code to an relatively independent function to make it modularized?

src/foraging_gui/Foraging.py
micahwoodard: I made some logic so if the baseline time has elapsed before a session, it will skip. This way, users can stop during the baseline waiting period and resume again.

XX-Yin:
> It would be better to move the main update to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) (after the photometry timer) for easier organization.

@micahwoodard This looks good. Can you move the "[wait](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6318)" to  this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) instead of the _StartTrialLoop?https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6306

micahwoodard:
> > It would be better to move the main update to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) (after the photometry timer) for easier organization.
> 
> @micahwoodard This looks good. Can you move the "[wait](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6318)" to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) instead of the _StartTrialLoop?https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6306

Why would it be better there? I think we should keep it in the start loop since it uses the same QApplication.processEvents() as the start loop. Hopefully in the future we can move away from using QApplication.processEvents() and I think it would be easier to do if we have all of the blocking code in one function.

XX-Yin:
> > > It would be better to move the main update to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) (after the photometry timer) for easier organization.
> > 
> > 
> > @micahwoodard This looks good. Can you move the "[wait](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6318)" to this [place](https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/main/src/foraging_gui/Foraging.py#L6170) instead of the _StartTrialLoop?https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1507/files#diff-b3cf4ca4fc9cb05d0a854615b8178bb10c1805e102fa1685fd60bd66bec9d34aR6306
> 
> Why would it be better there? I think we should keep it in the start loop since it uses the same QApplication.processEvents() as the start loop. Hopefully in the future we can move away from using QApplication.processEvents() and I think it would be easier to do if we have all of the blocking code in one function.

Just for organization. Moving this new timer after the photometry timer would make the code easier to read and organize.

micahwoodard:
Ah okay. I guess it doesn't really matter to me because _Start function is the blocking function anyways since it's calling StartLoop. Just to double check, do we want the baseline to happen before the fip baseline? That makes the most since to me but let me know if that's not the case

micahwoodard:
@XX-Yin @galenlynch I'm a little unclear on the fip baseline part of the session. It seems like how I have the logic  now, if you are running a fip session when you press start, the baseline will be recorded during the habituation period as well. Do we want that to be the case?

XX-Yin:
> @XX-Yin @galenlynch I'm a little unclear on the fip baseline part of the session. It seems like how I have the logic now, if you are running a fip session when you press start, the baseline will be recorded during the habituation period as well. Do we want that to be the case?

@micahwoodard We don't want the baseline to be recorded for fiber photometry.

XX-Yin:
> @XX-Yin could we move both timers into the _StartTrialLoop instead? It's a little easier to keep track of the behavior_baseline_period Event that way?

@micahwoodard Can you expand a little more about how it helps to keep track of the behavior_baseline_period? The _StartTrialLoop is more related to "trials". Here the timer is only running once and I think it's better to put it outside of the _StartTrialLoop. In general, we may need to reorganize the _Start functions to make it more modularized, which is not the scope of this PR.

micahwoodard:
> @micahwoodard Can you expand a little more about how it helps to keep track of the behavior_baseline_period? The _StartTrialLoop is more related to "trials". Here the timer is only running once and I think it's better to put it outside of the _StartTrialLoop. In general, we may need to reorganize the _Start functions to make it more modularized, which is not the scope of this PR.

It's just easier programmatically to handle the the set and clearing of the threading Event that breaks the loop if a user presses stop. I can change it, I don't think it'll be that hard

XX-Yin:
> > @micahwoodard We don't want the baseline to be recorded for fiber photometry.
> 
> Any idea on how to avoid this? I don't have a huge understanding on how the fip baseline is taken but it seems like the workflow is started before the user presses start so I'm not sure how to do the habituation timer before the fip workflow is started

We can trigger the following function only after the baseline period. 

            # Check if photometry excitation is running or not
            if self.PhotometryB.currentText() == "on" and (
                not self.StartExcitation.isChecked()
            ):
                logging.warning(
                    'photometry is set to "on", but excitation is not running'
                )

                reply = QMessageBox.question(
                    self,
                    "Box {}, Start".format(self.box_letter),
                    'Photometry is set to "on", but excitation is not running. Start excitation now?',
                    QMessageBox.Yes | QMessageBox.No,
                    QMessageBox.Yes,
                )
                if reply == QMessageBox.Yes:
                    self.StartExcitation.setChecked(True)
                    logging.info("User selected to start excitation")
                    started = self._StartExcitation()
                    if started == 0:
                        reply = QMessageBox.critical(
                            self,
                            "Box {}, Start".format(self.box_letter),
                            "Could not start excitation, therefore cannot start the session",
                            QMessageBox.Ok,
                        )
                        logging.info(
                            "could not start session, due to failure to start excitation"
                        )
                        self.Start.setChecked(False)
                        return
                else:
                    logging.info("User selected not to start excitation")
                    self.Start.setChecked(False)
                    return
                    
                    
However, we can probably simplify the logic by not considering the interaction of the baseline with the fiber photometry mice at this moment. We don't have a fiber photometry mouse that will perform the baseline period for ephys recording right now (may have in a few months).

micahwoodard:
> > > @micahwoodard We don't want the baseline to be recorded for fiber photometry.
> > 
> > 
> > Any idea on how to avoid this? I don't have a huge understanding on how the fip baseline is taken but it seems like the workflow is started before the user presses start so I'm not sure how to do the habituation timer before the fip workflow is started
> 
> We can trigger the following function only after the baseline period.
> 
> ```
>         # Check if photometry excitation is running or not
>         if self.PhotometryB.currentText() == "on" and (
>             not self.StartExcitation.isChecked()
>         ):
>             logging.warning(
>                 'photometry is set to "on", but excitation is not running'
>             )
> 
>             reply = QMessageBox.question(
>                 self,
>                 "Box {}, Start".format(self.box_letter),
>                 'Photometry is set to "on", but excitation is not running. Start excitation now?',
>                 QMessageBox.Yes | QMessageBox.No,
>                 QMessageBox.Yes,
>             )
>             if reply == QMessageBox.Yes:
>                 self.StartExcitation.setChecked(True)
>                 logging.info("User selected to start excitation")
>                 started = self._StartExcitation()
>                 if started == 0:
>                     reply = QMessageBox.critical(
>                         self,
>                         "Box {}, Start".format(self.box_letter),
>                         "Could not start excitation, therefore cannot start the session",
>                         QMessageBox.Ok,
>                     )
>                     logging.info(
>                         "could not start session, due to failure to start excitation"
>                     )
>                     self.Start.setChecked(False)
>                     return
>             else:
>                 logging.info("User selected not to start excitation")
>                 self.Start.setChecked(False)
>                 return
> ```
> 
> However, we can probably simplify the logic by not considering the interaction of the baseline with the fiber photometry mice at this moment. We don't have a fiber photometry mouse that will perform the baseline period for ephys recording right now (may have in a few months).

Great thanks for this!

micahwoodard:
@XX-Yin made necessary changes but this would fail if users pressed the Start FIP Workflow button before pressing start. Should we remove that button or disable it?

XX-Yin:
> @XX-Yin made necessary changes but this would fail if users pressed the Start FIP Workflow button before pressing start. Should we remove that button or disable it?

@micahwoodard I believe you mean this would fail when user pressed **Start excitation** button before pressing start. Let's not change the button in this PR as it involves changing of fiber SOP. 

We can move on to add the baseline timer without considering the interaction with the fiber photometry in this PR.

micahwoodard:
Okay sounds good. I'll test on rig tomorrow morning and then if that works I'll merge for next week's updates. @XX-Yin thanks for all the help!

micahwoodard:
@XX-Yin tested on rig with multiple stops during baseline and tested with interactions between stating fip workflows and  baselines

micahwoodard:
@XX-Yin could you approve this so it can be merged in next week?

src/foraging_gui/Foraging.py
```diff
@@ -6265,17 +6289,54 @@ def end_session_log(self) -> None:
...
+            return
+        
+        logging.info("starting trial loop")
 
         # Track elapsed time in case Bonsai Stalls
         last_trial_start = time.time()
         stall_iteration = 1
         stall_duration = 5 * 60
 
+        if self.Start.isChecked():
```
XX-Yin, line 6337: Can you move this to the if not self.Start.isChecked(): else structure?

micahwoodard: Yep

micahwoodard:
@XX-Yin ready for review again

src/foraging_gui/Foraging.py
```diff
@@ -6136,12 +6095,21 @@ def _Start(self):
...
+                self.Start.isChecked()
+                and self.PhotometryB.currentText() == "on"
+                and self.PhotometryRun == 0
         ):
+            # check if workflow is running and start photometry timer
+            if not self.photometry_workflow_running():
+                self.Start.setChecked(False)
+                return
+
             logging.info("Starting photometry baseline timer")
```
XX-Yin, line 6113: @micahwoodard Could you also package the Start photometry baseline timer to an independent function? I think it would be nicer and more organized to do so. That also works for me If you prefer to do it in another PR.

micahwoodard: @XX-Yin yes I can do that. Could I tackle that next week after we've tested that the baseline timer works well?

XX-Yin: Sure!

XX-Yin:
@micahwoodard One final question. Have you tested the fiber photometry with or without fiber workflow opened?

micahwoodard:
@XX-Yin Yes I have. It will warn you if the workflow has not been started once the baseline timer is finished and will act the same as the gui does now if you press yes or no. I tested both not having the workflow pressed and the start excitation pressed and behavior if you declined to turn either on and they all seem like they behaved how they do currently. Again, we will have problems if you start excitation before the baseline timer is finished though

XX-Yin:
"Again, it seems like problems will arise if excitation is started before the baseline timer is finished though"

@micahwoodard What's the problem?

micahwoodard:
> > @XX-Yin made necessary changes but this would fail if users pressed the Start FIP Workflow button before pressing start. Should we remove that button or disable it?
> 
> @micahwoodard I believe you mean this would fail when user pressed **Start excitation** button before pressing start. Let's not change the button in this PR as it involves changing of fiber SOP.
> 
> We can move on to add the baseline timer without considering the interaction with the fiber photometry in this PR.

As we discussed here, It would start the excitation during the baseline timer

XX-Yin:
> > > @XX-Yin made necessary changes but this would fail if users pressed the Start FIP Workflow button before pressing start. Should we remove that button or disable it?
> > 
> > 
> > @micahwoodard I believe you mean this would fail when user pressed **Start excitation** button before pressing start. Let's not change the button in this PR as it involves changing of fiber SOP.
> > We can move on to add the baseline timer without considering the interaction with the fiber photometry in this PR.
> 
> As we discussed here, It would start the excitation during the baseline timer

It's okay now.

PR #1510, opened 2025-04-22
120 character limit

micahwoodard:
  
### Describe changes:
- changes linter 120
-  I find the 80 character limit to be a little strict and it makes the code harder to read. It also expands the Foraging.py file by 1000 lines which makes it more difficult to navigate and find things. Just a suggestion though! I know 80 is what the rest of aind does but from a programming stance, 120 might work better for this project. 
- None
- None
- [ ] 446/7
- [ ]


galenlynch:
I don't have a strong opinion. Line breaks make it hard to read code, but code so complicated that a statement doesn't fit in 80 characters is also intrinsically hard to understand. I think 80 characters might be a good nudge to reduce function and statement complexity, but if we're past that point then 120 characters might make it easier to read.

galenlynch:
I often have many code panes open side-by-side which is also easier with modest line lengths, but I think 120 is fine for that.

alexpiet:
I would prefer to keep things consistent with the rest of AIND - and clean up code to reduce long lines instead of bumping the linter tolerance

PR #1512, opened 2025-04-24
adds warmup parameters to curriculum branch

micahwoodard:
- add warmup_setting for curriculum branch. When stage_1_warmup is run, trials look like this: 
image


alexpiet:
This makes sense, but I want to defer to @hanhou.

hanhou (approved):
Looks good to me. Thanks!

PR #1518, opened 2025-04-29
Feat 1106 baseline load in bug

micahwoodard:
  
### Describe changes:
- skips loading in baseline timer value since the prefix and suffix make it not an int. I tried to change it so it saves as a int using the .value() call but can't find where in the code to change. This timer will ultimately be set from somewhere in the new curriculum so I didn't want to spec to much time digging. Jsons can now be loaded in without error
- resolves [#1106](https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/1106)
- None
None
- [ ] 446/7
- Adds TP_hab_time_box but should effect pipeline


src/foraging_gui/Foraging.py
```diff
@@ -3098,9 +3098,14 @@ def _GetTrainingParameters(self, prefix="TP_"):
             self.Metadata_dialog,
         ]:
             # Iterate over each child of the container that is a QLineEdit or QDoubleSpinBox
+            for child in container.findChildren(QtWidgets.QLineEdit):
```
alexpiet, line 3101: What is this code block doing?

alexpiet:
This is a bigger PR that I expected. Can you just ignore the baseline timer value?

PR #1524, opened 2025-05-02
remove filter for schedule without assigned box

alexpiet:
  
### Describe changes:
- Does not filter out schedule rows without a box assignment

- https://github.com/AllenNeuralDynamics/dynamic_foraging_error_tracking/issues/609#issuecomment-2845356968

Box assignments are not required for `check_schedule` functionality to run

- none

- [ ] tested

- none

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -1561,7 +1561,7 @@ class Window(QMainWindow):
                 for x in schedule["Mouse ID"].unique()
                 if isinstance(x, str) and (len(x) > 3) and ("/" not in x)
             ]
-            self.schedule = schedule.dropna(subset=["Mouse ID", "Box"]).copy()
+            self.schedule = schedule.dropna(subset=["Mouse ID"]).copy()
             logging.info("Loaded behavior schedule")
         else:
             self.schedule_mice = None


micahwoodard:
@alexpiet I think this is working but there might me unintended consequences to doing this? Based on convos with @ellahiltonvano, the schedule contains mouse training for the current week and next week and so sometimes mice will be on the schedule twice. Seems like maybe filtering by box was a way to only get the currently trained mice? I don't know the logic behind some of these decisions or the ins and outs of the schedules but worth considering.

alexpiet:
> @alexpiet I think this is working but there might me unintended consequences to doing this? Based on convos with @ellahiltonvano, the schedule contains mouse training for the current week and next week and so sometimes mice will be on the schedule twice. Seems like maybe filtering by box was a way to only get the currently trained mice? I don't know the logic behind some of these decisions or the ins and outs of the schedules but worth considering.

Hmm, maybe you are right. I thought the code filtered for the correct week, but I guess not. I will delay merging this

micahwoodard:
When we tested, it seemed like it picked the mouse from the next week based on the trainer provided. It was on Friday though so maybe something with the filtering is going on there?

PR #1534, opened 2025-05-07
Production testing

micahwoodard:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
- merging production_testing to main for sw3 hotfix and adding hab_timer back

PR #1535, opened 2025-05-08
Revert "Production testing"

micahwoodard:
Reverts AllenNeuralDynamics/dynamic-foraging-task#1534

@ellahiltonvano found errors so need to roll back and merge next week


micahwoodard:
@XX-Yin unfortunately this needs to be rolled back. @ellahiltonvano tried running this on an ephys rig and there are still some lingering merge errors that I didn't catch. If we could revert this merge as soon as possible so that ephys rig can be run that would be great. All other rigs should be fine as this was merged in after the update this morning.

XX-Yin (approved):

PR #1543, opened 2025-05-22
Feat fast retract

micahwoodard:
  
### Describe changes:
- adds fast retract feature
- SIPE [ticket ](https://apps.powerapps.com/play/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&ItemID=4130)
- if fast retraction is enabled, lickspout will fast retract and unretract at specified speed and wait time
- None
- [x] 446/7
- No


micahwoodard:
Users can configure fast retract using widgets in parameters tab
image

Enabling fast retract is at the bottom of the tab
image

src/foraging_gui/MyFunctions.py
```diff
@@ -2917,6 +2921,13 @@ def _GetAnimalResponse(self, Channel1, Channel3, data_lock):
...
                         self.B_CurrentRewarded[0] = False
                         self.B_CurrentRewarded[1] = False
+
+                if self.B_AnimalCurrentResponse == 0:
+                    self.mouseLicked.emit("Left")
+                    logging.info("Left emit")
+                elif self.B_AnimalCurrentResponse == 1:
+                    self.mouseLicked.emit("Right")
+                    logging.info("Right emit")
                 B_CurrentRewarded = self.B_CurrentRewarded
```
micahwoodard, line 2931: @XX-Yin was wondering if self.B_AnimalCurrentResponse in the _GetAnimalResponse was a good place to check what the mouse's response is for each trial for the fast retract? I'm not sure if we need to catch as soon as the mouse licks. Any help as to where to move it would be much appreciated!

micahwoodard:
@XX-Yin would be nice to get your eyes on how I've implemented this

src/foraging_gui/Foraging.py
```diff
@@ -619,6 +617,88 @@ def _insert_stage_widget(self, widget_to_replace: str) -> None:
...
+        at_origin = list(self._GetPositions().values())[1:3] == [0, 0]
+        if tp.lick_spout_retraction and self.stage_widget is not None and not at_origin:
+            logger.info(f"Retracting {lick_spout_retract} lick spout.")
+            motor = 1 if lick_spout_licked == "Left" else 2                             # TODO: is this the correct mapping
+            curr_pos = self.stage_widget.stage_model.get_current_positions_mm(motor)    # TODO: Do I need to set rel_to_monument to True?
+            self.stage_widget.stage_model.quick_move(motor=motor, distance=pos-curr_pos, skip_if_busy=True)
+
+            # configure timer to un-retract lick spout
+            timer.timeout.disconnect()
+            timer.timeout.connect(lambda: self.un_retract_lick_spout(lick_spout_licked, curr_pos))
```
alexpiet, line 646: shouldn't this be `lick_spout_retract`, not `lick_spout_licked` ?

alexpiet: oh nvm, I see that function wants `lick_spout_licked`, a little confusing to me that you don't just pass that function which lick spout to un-retract, but this works too

alexpiet (approved):
looks good to me

XX-Yin:
> @XX-Yin would be nice to get your eyes on this since it alters the bonsai code. Putting a signal in the RigClient was the only place that I could figure out how to get the lick message soon enough. I'm going to try and get some videos later today to show functionality. @ellahiltonvano we should probably test with a real mouse? Not sure how we want to handle testing this so open to suggestions.

@micahwoodard sorry for this delay. Can you show how the bonsai code is changed?

src/foraging_gui/Foraging.py
micahwoodard: yeah that does seems a little unintuitive. That is easy to rework

micahwoodard:
@XX-Yin No worries! This is blocked any way with the new curriculum role out but wanted to get feedback on how I got it to work and if people had other ideas on where to best catch when the mouse has licked

Here is current bonsai code: 
image
image
image


Here is the new bonsai code. I'm usingthe  a LeftLick and RightLick subject for the JaneliaLickDetector and added a LeftLickLicketySplit  and RightLickLicketySplit subject for the NewLickDetector. Then I merge the respective subjects and write an osc message to channel 2 if the spout was licked. 

image
image
image

src/foraging_gui/rigcontrol.py
```diff
@@ -1,12 +1,17 @@
...
-
+from PyQt5 import QtCore
 from pyOSC3.OSC3 import OSCMessage
 
 
-class RigClient:
+class RigClient(QtCore.QObject):
+    mouseLicked = QtCore.pyqtSignal(str)
+
     def __init__(self, client):
```
micahwoodard, line 11: @alexpiet @XX-Yin does anybody have concerns about making the RigClient a QObject and using a signal here? I couldn't find a way to catch the lick fast enough using a thread in the main gui but it feels clunky catching the lick in the rig client.

XX-Yin:
@micahwoodard 

You only need to send a copy of licks to a different address. Is this what you changed? 

image

XX-Yin:
@micahwoodard What will happen if the Lick spout retraction is selected for the newscale stage?

src/foraging_gui/rigcontrol.py
XX-Yin: @micahwoodard This should be ok

micahwoodard:
> @micahwoodard What will happen if the Lick spout retraction is selected for the newscale stage?

Nothing. The newscale stage can't fast retract so if there is not an aind stage, there will be a log message that says fast retract is not available. Should we maybe disable the user from enabling fast retract if the rig does not have the capabilities? But also if fast retract is in the curriculum, that would mean mice would be off curriculum which seems weird?

XX-Yin:
@micahwoodard  Can we set other parameters (speed and when to retract) like we discussed before?

micahwoodard:
> Users can configure fast retract using widgets in parameters tab image
> 
> Enabling fast retract is at the bottom of the tab image

@XX-Yin yes you can configure speed and wait time. Are there other parameters people want to be able to control as well?

micahwoodard:
@XX-Yin @alexpiet this is a vid showing what this looks like. I'm testing with @ellahiltonvano today with an actual mouse. 

https://github.com/user-attachments/assets/dcd73428-9a12-4f19-be60-7235651e07a3

micahwoodard:
@ellahiltonvano and I tested this last week with a mouse and there were some unusual bugs. Still working on investigating and fixing.

PR #1555, opened 2025-05-30
Log temperature

micahwoodard:
  
### Describe changes:
- Adds bonsai node to interface with the environment sensor on rigs. Also saves B_EnvironmentSensorTemperature, B_EnvironmentSensorHumidity, B_EnvironmentSensorPressure, and B_EnvironmentSensorTimestamp to the behavior.json. 
SIPE ticket [#4132](https://apps.powerapps.com/play/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&ItemID=4132)
- None
- com port needs to be added to the Setting_box#.csv file. Example, for 1A EnvSen,COM3 needs to be added. 
- [x] 446/7
- Adds B_EnvironmentSensorTemperature, B_EnvironmentSensorHumidity, B_EnvironmentSensorPressure, and B_EnvironmentSensorTimestamp to the behavior.json


micahwoodard:
@Ahad-Allen this will add B_EnvironmentSensorTemperature, B_EnvironmentSensorHumidity, B_EnvironmentSensorPressure, and B_EnvironmentSensorTimestamp to the behavior.json. I don't think that should break anything but wanted to double check

micahwoodard:
@XX-Yin here are the bonsai workflows added 
image
image

XX-Yin:
@micahwoodard what will happen if we don't have the environment sensor connected?

micahwoodard:
@XX-Yin 

> @XX-Yin here are the bonsai workflows added. I'm not sure if all rigs have environment sensors, and I'm unsure how to check in bonsai if a key exists or not. We could just skip the set up if there is no environment sensor configured

I'm not sure how to check if a key is in a dictionary in bonsai. Do you?

XX-Yin:
@micahwoodard 

Some general questions

1. what is the TrialEnd doing here?
2. When do we get the Temperature, Humidity, Pressure? How are they aligned with the harp timestamp? 
3. What is the frequency to get these parameters?

XX-Yin:
> @XX-Yin
> 
> > @XX-Yin here are the bonsai workflows added. I'm not sure if all rigs have environment sensors, and I'm unsure how to check in bonsai if a key exists or not. We could just skip the set up if there is no environment sensor configured
> 
> I'm not sure how to check if a key is in a dictionary in bonsai. Do you?

@micahwoodard You can use the CameraPreview node as an example

micahwoodard:
@XX-Yin 

> 1. what is the TrialEnd doing here?

It triggers the sensor to collect environment data after every trial end. Without it, the sensor would never be prompted to collect data after initialization. We could also use a timer or other events. TrialEnd made sense to me since other data is emited after each trial end

> 2. When do we get the Temperature, Humidity, Pressure? How are they aligned with the harp timestamp?

They are aligned with the TrialEnd signal. I will double check they are aligned with the harp timestamps

> 3. What is the frequency to get these parameters?

After every TrialEnd. We can change this if we want more or less.

micahwoodard:
@XX-Yin I've addressed concerns about the com port not being in the settings json with this check. 

image


This is the code for the timestamps. Looking at the values in the json that are saved, it seems like they are synced to the behavior timestamps. I'm not sure though. what do you think?
image
image

micahwoodard:
@nathan-mars that sounds good to me! Will update

@XX-Yin how are other data points aligned with the harp timestamps? I'm unsure of how to verify if the environment sensor timestamps are synced

Ahad-Allen:
I can update the capsules with the new fields and starting testing later this week

micahwoodard:
@XX-Yin so I should use the Behavior.TimestampedDigitalInputStat instead?

micahwoodard:
> @micahwoodard This is not precise. You should use code like this
> image

@XX-Yin would be nice to be able to merge this in for next weeks testing. I don't think I can use TimestampedDigitalInputStat because it seems to me this only happens when a lickspout is licked? What specific part of this workflow are you referencing and why is using the timestamps imprecise if that is what other data uses? It seems like the two are coming from similar sources. 
image

XX-Yin:
> > @micahwoodard This is not precise. You should use code like this
> > image
> 
> @XX-Yin would be nice to be able to merge this in for next weeks testing. I don't think I can use TimestampedDigitalInputStat because it seems to me this only happens when a lickspout is licked? What specific part of this workflow are you referencing and why is using the timestamps imprecise if that is what other data uses? It seems like the two are coming from similar sources. image

@micahwoodard To directly pair the environment sensor event with a timestamp is not precise as they are not synchronized. 

1. Is the environment sensor a HARP device? If so, could you add it to the logging node?

2. It would be preferable to collect measurements at a fixed frequency rather than at the end of each trial, since trial durations are varied. That said, perhaps this doesn’t matter much if variables like temperature and humidity remain relatively stable. Do you have any data showing the stability of these measurements?

nathan-mars:
>@micahwoodard It’s not accurate to directly pair environment sensor events with timestamps, as they are not synchronized.
> 1. Is the environment sensor a HARP device? If so, could you add it to the logging node?

It is a Harp device with a 3.5mm sync line. I don't understand the comment that env sensor events aren't synchronized with timestamps, but I'll leave that to others to sort out. 

> 2. It would be preferable to collect measurements at a fixed frequency rather than at the end of each trial, since trial durations are varied. That said, perhaps this doesn’t matter much if variables like temperature and humidity remain relatively stable. Do you have any data showing the stability of these measurements?

Multiple measurements should be made, not just the start/end temperature. Ideally once a second or something convenient to implement. One of the reasons for adding this to the workflow is so that we can test how stable temperature/etc are.

bruno-f-cruz:
I already talked to @micahwoodard about this. This is a harp-device as such it can be logged like ANY harp device (https://allenneuraldynamics.github.io/Bonsai.AllenNeuralDynamics/articles/core-logging.html#with-metadata). Is there a reason to not follow the standard?

The device is indeed a flully fledged harp device, it collects data at a periodic rate (https://allenneuraldynamics.github.io/Bonsai.AllenNeuralDynamics/articles/environment-sensor.html#parsing-sensor-data) and is synchronized with all other harp devices in the rig provided a synch cable is provided.

micahwoodard:
Great thanks so much for everyone's input!

> 1. If so, could you add it to the logging node?

> I already talked to @micahwoodard about this. This is a harp-device as such it can be logged like ANY harp device (https://allenneuraldynamics.github.io/Bonsai.AllenNeuralDynamics/articles/core-logging.html#with-metadata). Is there a reason to not follow the standard?

@XX-Yin where is the logging node? Is it the same as what @bruno-f-cruz is referencing? Do we not need to save values in the behavior json if we save them in the log?

> The device is indeed a flully fledged harp device, it collects data at a periodic rate (https://allenneuraldynamics.github.io/Bonsai.AllenNeuralDynamics/articles/environment-sensor.html#parsing-sensor-data) and is synchronized with all other harp devices in the rig provided a synch cable is provided.

@bruno-f-cruz does that mean that the timestamps coming from the sensor are already synchronized so I can just use those directly? I do not have a lot of experience with harp devices or bonsai or hardware on the behavior rigs

micahwoodard:
@XX-Yin anything else I need to do to get this merged in?

XX-Yin:
> @XX-Yin anything else I need to do to get this merged in?

@micahwoodard  Have you tested the updated version?

src/foraging_gui/MyFunctions.py
```diff
@@ -3033,6 +3037,19 @@ def _get_irregular_timestamp(self, Channel2):
                 self.B_AutoRightWaterStartTime = np.append(
                     self.B_AutoRightWaterStartTime, Rec[1][1][0]
                 )
+            elif Rec[0].address == "/EnvironmentSensorTemperature":
+                value = Rec[1][1][0] if type(Rec[1][1][0]) != float else round(Rec[1][1][0], 1)
```
XX-Yin, line 3041: is it necessary to round?

micahwoodard: In earlier comments, it was decided to round 

from @alexpiet 
> @micahwoodard @nathan-mars The raw data from the sensors look like they have 16 digits after the decimal point. What is the actual precision of these sensors? I'm not sure if its best practice to just log everything and deal with it later, or round to the actual precision now.

from @nathan-mars 
> The datasheet says ±0.5ºC and ±1.5% RH accuracy ([BME688](https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bme688-ds000.pdf)). (Side note - we haven't calibrated the relative humidity, so I would not trust that value yet).
> Keeping 1 past decimal makes sense to me (23.45678 recorded as 23.5)

micahwoodard:
had to change layout file since I was getting this error with old one. Double checked on rig that layout still looks good 
image

PR #1563, opened 2025-06-22
BonsaiGUI windows optimization

hagikent:
### Describe changes:
Lick, Camera windows optimization for 446/447 RAs' operations.
Pure visualization update in the `bonsai.layout` file.

https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/641#issuecomment-2937361472

N/A. No need to drug windows anymore.

N/A. (maybe Ephys rigs need their own optimisation)

Yes, tested/developed in 447, 3D.

N/A

### DIFF
--- a/src/workflows/foraging.bonsai.layout
+++ b/src/workflows/foraging.bonsai.layout
@@ -64,12 +64,12 @@
     <EditorDialogSettings>
       <Visible>true</Visible>
       <Location>
-        <X>4</X>
-        <Y>5</Y>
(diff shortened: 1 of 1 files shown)


hagikent:
Thanks @micahwoodard . True, I could have added a screenshot, my bad.
By the way, I realized that .bonsai workflow file is now somehow protected so we can not overwrite by`save`. I guess this is designed/intentional?

PR #1577, opened 2025-07-21
Safer Ephys saving.

yonibrowning:
We have, in cases of high computer load, had the gui crash during ephys recordings. If the crash happens while the file is saving (which it seems to do more often than not), the result is a corrupt JSON that seems to be unrecoverable. This bug has caused us to lose valuable segments of ephys+behavior data.

Here, I have modified the save function to (1) save a new file each trial called "tmp" (2) once tmp is safely saved, delete the old log and (3) rename tmp to the name of the desired json log file. This way only the last trial of the task will be lost if something gets corrupted during the saving process. This does not increase the saving load on any trial, it simply avoids directly overwriting data to prevent loss.

I been running this solution since May on on rig 2 and it seems to be stable. Because it is hard to get the task to crash in a controlled way it is difficult to confirm that it is totally ephys safe, but it should be much more robust. In the event of a crash, what should (theoretically) happen is that the "tmp" file should be corrupted leaving a health json file from the last trial.

A couple of notes
(1) A better way to do this would be to save only information for the last trial each trial, but this solution was vetoed in the past due to fears about extra effort to maintain. The proposed fix here achieves a similar goal, albeit with higher "save" load on each trial.

(2) I accidently committed these changes on a computer with Galen's github logged on.

Thanks!

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -4067,8 +4067,17 @@ class Window(QMainWindow):
                     with open(self.SaveFile, "w") as outfile:
                         json.dump(Obj2, outfile, indent=4, cls=NumpyEncoder)
                 elif self.SaveFile.endswith(".json"):
-                    with open(self.SaveFile, "w") as outfile:
+                    # Crashses during save can corupt a json file.
+                    # Make tmp file to save to
+                    tmp_file_name = self.SaveFile.split('.json')[0]
+                    tmp_file_name = tmp_file_name + '_tmp.json'
+                    with open(tmp_file_name, "w") as outfile:
                         json.dump(Obj, outfile, indent=4, cls=NumpyEncoder)
+                    # After file is safely saved, remove the old save file
+                    # and rewrite the new one.
+                    if os.path.isfile(self.SaveFile):
+                        os.remove(self.SaveFile)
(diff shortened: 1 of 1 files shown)


yonibrowning:
I have changed to merge with develop. Thanks!

XX-Yin (approved):

alexpiet:
Sorry you've had data loss - sounds like the crashes were because the computer crashed not the GUI, but I understand you want the data to be safe in case the computer crashes. 

I don't think this will significantly change the CPU/memory use during saving, but that might be a good thing to verify. 

I'll defer to @micahwoodard for review.

alexpiet:
@micahwoodard Can you please review this PR?

micahwoodard (approved):
seems a little hacky but if it works for you and has been preserving data, it seems like a good solution

micahwoodard:
@yonibrowning @galenlynch can we merge this?

yonibrowning:
Yes please!Sent from my iPhoneOn Sep 18, 2025, at 2:07 PM, Micah Woodard ***@***.***> wrote:﻿micahwoodard left a comment (AllenNeuralDynamics/dynamic-foraging-task#1577)
@yonibrowning @galenlynch can we merge this?

PR #1585, opened 2025-08-06
Automatic bias correction

### LINKED ISSUE #1426: Better automate manual water rewards
We'd like to improve the algorithm of autowater delivery for motivation and bias
1. Motivation
    - Current: give autowater when `N` **_ignore trials_** happen in a row, where `N` is defined in the curriculum: 
      Image
      Image
    - New proposal:
        -  TBD
    
3. Bias
    - Current: implicitly addressed by giving autowater when `N` **_Unrewarded trials_** happen in a row
      Image
      Image
    - New proposal:
        - Inspired by Bowen's comment (and Han's experience)
            > Start with manual water to overcome bias. Continue pressing manual water for a couple of trials on the choice side.
            >If animals switch, then monitor closely their choices and block transitions in next a few blocks.
            >If animals don't switch and continue sticking to one side, move lickspout at a step size of 100µm. And only move lickspount for 1 step size per trial away from the biased side.
            >Move the lickspouts until animals switch to the other side while pressing manual water on the previously biased side in a couple of trials to ensure animals can still reach the biased side.
        ```
         (autowater only)
         while auto_water_for_bias == True:
             If |bias| >= `threshold_1` and P_L != P_R and last `N` choices are all on the side with lowest probability:
                  give `M` drops of water on the better side
                  wait for `X` trials
        ```

- related https://github.com/AllenNeuralDynamics/aind-behavior-blog/discussions/891

micahwoodard:
  
### Describe changes:
- adds logic to give water or move stage based on bias calculated. Bias correction parameters can be controlled through the ui and can be found [here ](https://github.com/AllenNeuralDynamics/Aind.Behavior.DynamicForaging/blob/38d781cc2f13f92e9a3ed54c1531707a8218739e/src/aind_behavior_dynamic_foraging/DataSchemas/operation_control.py#L39C1-L56C97). I put the water attempt parameter in the operational control model for now since adding this parameter to the curriculum seems like a larger effort/discussion. 
- resolves #1426 
- sipe ticket [4115](https://apps.powerapps.com/play/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&ItemID=4115)
- water and stage movement for bias will automatically be handled 
- None
- [x] 446/7
- no


micahwoodard:
Here is a preview of the gui so bias parameters can be configured

image

micahwoodard:
implemented based on @hanhou input found below 

I agree with Xinxin Yin that being able to turn auto water and lick spout on and off separately may be desirable, but I don't think their logic should be independent or we should only rely on lick spout even in early sessions. To me, using water as an incentive to shape behavior is better than moving the lickspout, since it does not change the nature of the task (i.e., tracking water reward) and thus doesn't make the mice mistakenly think it's a different task (i.e. tracking the lick spouts). 
 
In summary, we may want:
(Bowen) try water first, and if it doesn't work, move the lick spout
(Han) move the lick spouts to the biased side if |bias| > threshold_bad and bring them back to center if |bias| < threshold_good
(Xinxin) being able to control the relative strength of autolickspout and autowater over the training
I can think of a simple algorithm that implements all these points. It essentially uses a parameter antibias_water_first_max to control, once |bias| crosses threshold_bad, the number of auto-water attempts before we start to move the lick spouts:

```
antibias_water_count = 0
if bias_check_interval <= trial_number - last_bias_check:
    last_bias_check = trial_number
    if |bias| > threshold_bad:   # red state
        # do intervention
        if antibias_water_count < antibias_water_first_max:
             give_water_correction()
             antibias_water_count += 1
        else:
             move_lickspout_to_the_biased_side()
        return
    # Reset antibias_water_count once not in red state
    antibias_water_count = 0
    
    # If in green state, always bring lick spouts back to center
    if |bias| < threshold_good: 
        move_lickspouts_back_to_center()
       
    # If in yellow state, do nothing
```
 
In this implementation, antibias_water_first_max=0 gives us pure lickspout movement (my original proposal)
 
image

antibias_water_first_max=inf gives us pure autowater 
 
image

and antibias_water_first_max=1 gives us Bowen's proposal
 
image

we can gradually increase the parameter antibias_water_first_max over the curriculum (from 1 to inf?).

ellahiltonvano:
@hanhou @XX-Yin @bwtan-allen @ZhixiaoSu could you please review this PR for Micah so we can continue to testing?

ellahiltonvano:
@hanhou @XX-Yin @bwtan-allen @ZhixiaoSu could you please review this PR for Micah so we can continue to testing? I can schedule a meeting to discuss if needed.

JeremiahYCohen:
Sounds good, thanks. I think it's fine as a UI input.

PR #1619, opened 2025-09-26
catching up "develop" with "Ading_second_stimulus"

hagikent:
`ading_second_stimulus` is derived from very old `main`

`adding_CS2_catchup` is newly branched from the latest `develop` 

I would like to merge them first and then `adding_CS2_catchup` to `develop` to be safer.
@micahwoodard would be great if you could take a look at the conflicts as I'm not sure how to keep only relevant changes.

`ading_second_stimulus` is working but missing two important functions:
- automatic lick spout location across days
- missing some metadata(?) and not triggering the pipeline properly?


micahwoodard:
@hagikent sorry I misunderstood the point of this pr. I thought adding _CS2_catchup was another functionality branch that you eventually wanted to merge into develop but it looks like its just an empty branch that you were wanting to use as a way to update ading_second_stimulus. Sorry for my confusion! You are right, this pr is redundant so I will close

PR #1620, opened 2025-09-29
Second stimulus merge

micahwoodard:
  
### Describe changes:
- updates branch ading_second_stimulus branch with develop 
- none
- none


hagikent:
Thanks @micahwoodard . I will test the `second-stimulus-merge` branch on the rig. And get back to you. 

While solving the conflict, any specific non-sencond-stimulus function that should be explicitly tested?

I guess I can close this PR as redundant?  https://github.com/AllenNeuralDynamics/dynamic-foraging-task/pull/1619

alexpiet:
@hagikent is there any extra information that needs to be added to the nwbs?

hagikent:
@alexpiet thanks for the heads up. should metadata/behavior-json be also modified in a coordinated manner?

rachelstephlee:
This was not processed with the correct aind-bonsai-behavior-nwb package. The CS+/CS- is not being packaged in the NWB. Will follow up with Ahad and Kenta separately.

rachelstephlee:
We are waiting for a job_type so we can trigger the right pipeline; that is waiting on me to finish validating one last PR...

micahwoodard:
> One question, can we switch job_type (=parameters to be fed into watch-dog) flexibly session-by-session within the same branch...? (if not we cannot merge this into main)

@alexpiet is the job_type currently being set by the setting file under the transfer_service_job_type key? That would mean a user would have to update the settings file which does not seem like a good solution

alexpiet:
> > One question, can we switch job_type (=parameters to be fed into watch-dog) flexibly session-by-session within the same branch...? (if not we cannot merge this into main)
> 
> @alexpiet is the job_type currently being set by the setting file under the transfer_service_job_type key? That would mean a user would have to update the settings file which does not seem like a good solution

Yes, its set right now via the settings file. I think its a very bad idea to have a user modify the settings file to run a specific mouse. We could move the job type to the curriculum, or just set logic that determines what the job type is based on task or something.

Why does the second stimulus need a different job type?

Ahad-Allen:
That is correct. We are just waiting for Rachel's validation and then we can mark the current PR for the DF + Fiber pipeline as a release that can be triggered with the job type for older sessions. We will then move over to using the fiber + DF split pipelines.

hagikent (approved):

PR #1632, opened 2025-10-07
Use saturation visualizer as the default visualizer for cameras

bruno-f-cruz:

### DIFF
--- a/bonsai/Bonsai.config
+++ b/bonsai/Bonsai.config
@@ -6,6 +6,7 @@
     <Package id="AForge.Video.DirectShow" version="2.2.5" />
     <Package id="AllenNeuralDynamics.AindBehaviorServices" version="0.8.0" />
     <Package id="AllenNeuralDynamics.Core" version="0.4.0" />
+    <Package id="AllenNeuralDynamics.Core.Design" version="0.4.0" />
     <Package id="AllenNeuralDynamics.EnvironmentSensor" version="0.3.0" />
     <Package id="AllenNeuralDynamics.LicketySplitLickDetector" version="0.1.0-preview2024020101" />
     <Package id="AssimpNet" version="4.1.0" />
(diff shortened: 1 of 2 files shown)


bruno-f-cruz:
This PR replaces the previous visualizer for cameras by:
image

Each one of those combinator operators has 3 properties (max, min, alpha) that can be used to dynamically adjust what is getting shown.

rachelstephlee:
Notes: 
- layout file will need to be updated, or will crash 
- alternatively, need to update latest bonsai version

alexpiet:
Please merge to `develop` not `main`

ellahiltonvano:
[Here ](https://apps.powerapps.com/play/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&ItemID=4475)is the related SIPE ticket

micahwoodard:
@bruno-f-cruz @rachelstephlee @ellahiltonvano it looks like the camera recording visualizer was replaced with the saturation visualizer. Would it make more sense for the camera preview window to be the saturation visualizer since that is what's used during set up?

ellahiltonvano:
@micahwoodard most RAs that run FIP do not use the preview window outside of the first session. It is used session 1 to set lickspout location before starting, but after that we just use the recording window. 

I suppose per this new saturation window, we would need to be checking before starting the session, is that right @rachelstephlee?

ellahiltonvano:
@micahwoodard we can start recording before starting behavior, but its in the SOP to use autocontrol (where the recording starts 5sec after behavior). 

Correct about preview and recording, we are currently only recording in FIP sessions (all other sessions use preview).

ellahiltonvano:
@rachelstephlee I think I'm still concerned that this is adding an additional step to the RA workflow when the likelihood that Sci intervention will be done before the session is very low. It seems like this is something that would be better to live in the QC portal - scientists can check their mice daily, and find time to resolve the issue in that box. If scientists are checking their data every day, then we wouldn't have more than a few sessions with an issue - even if RAs identify the issue before a session, it doesn't seem like this is something that would be able to be resolved before the mouse is run given time constraints.

micahwoodard:
decided not to implement visualizer check

PR #1651, opened 2025-11-03
Update  Environment Sensor Logging

micahwoodard:
  
### Describe changes:
- aligns logging of env sen with aind standards
- [#137](https://github.com/AllenNeuralDynamics/aind-fip-nwb-base-capsule/issues/137#issuecomment-3474891887)
- None
- None
- [x] 446/7
- None

### DIFF
--- a/src/workflows/foraging.bonsai.layout
+++ b/src/workflows/foraging.bonsai.layout
@@ -40,12 +40,12 @@
     <EditorDialogSettings>
       <Visible>true</Visible>
       <Location>
-        <X>3</X>
-        <Y>3</Y>
+        <X>4</X>
+        <Y>5</Y>
       </Location>
       <Size>
-        <Width>512</Width>
-        <Height>743</Height>
+        <Width>1843</Width>
+        <Height>842</Height>
       </Size>
       <WindowState>Normal</WindowState>
     </EditorDialogSettings>
@@ -498,12 +498,12 @@
     <EditorDialogSettings>
       <Visible>true</Visible>
       <Location>
-        <X>3</X>
-        <Y>3</Y>
+        <X>4</X>
+        <Y>5</Y>
       </Location>
       <Size>
-        <Width>512</Width>
-        <Height>743</Height>
+        <Width>1843</Width>
+        <Height>842</Height>
       </Size>
       <WindowState>Normal</WindowState>
     </EditorDialogSettings>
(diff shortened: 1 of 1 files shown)


bruno-f-cruz:
Some notes: 
1 - If no one is using the former logger, you may as well deprecate it ASAP in favor of the new one
2 - If I understand correctly, the read dump will NOT be captured by the logger. The read-dump is asked 1 second after the workflow starts, but the logger is only started "on request" from the GUI.

image

Otherwise, the pattern seems implemented correctly!

micahwoodard:
> Wouldnt it make more sense to trigger the read dump every time you initiate a new logger instance from the incoming OSC message?

I remember when I was doing this, I was having a hard time figuring out how to periodically write data and this ended up working so I went with it. Can I replace the timer with a start logging event and then the device will emit periodic events of all registers?

micahwoodard:
@bruno-f-cruz updated this to be better :) Now instead of triggering read dump every second, I just parse the event. Then when logging is starting, I trigger a read dump 1 second later. Files look like everything is writing correctly

image
image
image

image

bruno-f-cruz:
This makes sense to me. The only issue I can see is that `Parse.Take(1)`. This means that yo will only allow one message to be sent by each time the bonsai is opened. Is this the intended behavior?

micahwoodard:
> This makes sense to me. The only issue I can see is that `Parse.Take(1)`. This means that yo will only allow one message to be sent by each time the bonsai is opened. Is this the intended behavior?

That was how it was originally coded, I just copy and pasted to make a subject. @XX-Yin any insight here?

micahwoodard:
This makes sense to me. The only issue I can see is that Parse.Take(1). This means that yo will only allow one message to be sent by each time the bonsai is opened. Is this the intended behavior?

@XX-Yin is this right for the logging?

micahwoodard:
@XX-Yin can you confirm this is right? I don't know your original intent behind this

XX-Yin:
> @XX-Yin can you confirm this is right? I don't know your original intent behind this

@micahwoodard Sorry for this delay. Yes, the "Parse.Take(1)" is intended. We trigger it for each session. For the parts related to the environmental sensor, they make sense to me, but I’m not very familiar with that.

alexpiet:
I think the next steps are to adopt the tools Bruno suggested

PR #1656, opened 2025-11-07
Add documentation for block structure to README.md

hanhou:
- Add description for block structure
- Reorder the sections
~~(merging to main directly)~~


XX-Yin:
It’d be better to merge into the develop branch as well, since we need to keep different branches synchronized.

alexpiet (approved):

micahwoodard:
@hanhou I think it got reverted when main had to be reverted due to the second stimulus and bug and got lost in all of the merge conflicts that had to be fixed

Issue #1689, opened 2025-12-18
Data retention policy after changes to on-rig compression of videos


rachelstephlee:
This is a ticket to discuss data retention needs after we make changes on on-rig compression of videos. Rachel will fill out. 

Additional notes: 

Adam's JPOD is still being actively used (Login with Kenta). If we increase data size significantly, the time it takes to transfer to JPOD will change.

alexpiet:
> i can file a ticket with SIPE to allow us to _only_ keep the data we want (lick spout location) and nothing else.

We want all behavior data other than video. All the settings are required to run properly

> Note that with the current on-rig compression, data size is NOT an issue. it's the same as before the PR went in.

In that case, lets skip making any changes right now. When the refactor is done, we won't need to keep any data on the rig, so this problem will go away. If you agree @rachelstephlee, can you close this ticket?

rachelstephlee:
I'd prefer we put this request in because I'd rather we use the AIND file standard compression for 446/447. The rigs should be able to keep up, and this is the best quality video we can have. 

Added ticket here: https://apps.powerapps.com/play/e/9084dabc-d707-e1bf-8113-685be25cac15/a/cb338490-e23d-48e6-9e6d-e600b8d17dd4?tenantId=32669cd6-737f-4b39-8bdd-d6951120d3fc&Hidenavbar=true&

rachelstephlee:
Latest version of AIND file standard test states that the on-rig save would be 120GB for a 90 minute video (with fiber). This is with mp4 (not mkv, but galen stated that we should have around the same size) 

See a 45 minute video in scratch/rachel/446_6D_AINDFILESTANDARD_test/behavior_823819_2026-03-12_11-59-17

Given this, I think we should be OK with data retention. Data will be deleted once it's uploaded to CO, which would take ~24, at most 48 hours. 

We can save around 15 sessions/rig (estimated from 1900 GB), which should be more than enough for 2 days of sessions. This is ONLY for 446/447 rigs.

PR #1593, opened 2025-09-02
Implement FFMPEG logger

bruno-f-cruz:
This PR implements an ffmpeg logger for all cameras. The following caveats should be considered:
- I tried to touch the workflow as little as possible. A better approach would be to adopt the same pattern for dynamically creating cameras as needed like we do in VrForaging but this would require substantial changes.
- The codecs are currently hard-coded according to https://allenneuraldynamics.github.io/aind-file-standards/file_formats/behavior_videos/#acquisitionrawprimary-data-format.
- I did not want to implement them flexibly since I don't have visibility into how the CSV settings are maintained and the currently reserved keywords. If you need it, it should be an easy addition.
- Despite enabling encoding for all cameras, i can make no guarantees as to whether the hardware will be able to keep up with the 4 cameras currently specified at 500fps each (I highly doubt it)
- There was an undocumented prefix on top of the root logging path (i.e. <path>/../../behavior-videos) that I have no idea why was necessary. Perhaps the dynamic foraging GUI is considering something else as being the "root path" than what bonsai is? Lmk if this is an issue, and I can easily fix it.
- The rest of the workflow is still responsible for making sure the triggering only starts AFTER the cameras have started and, ideally, make sure that cameras stop before the recording stops. This will ensure that you get both the start and end of the stream in your data.

PR #1693, opened 2025-12-20
FFMPEG on-rig parameter

rachelstephlee:

Sets the video on-rig compression parameters to the parameters that will give us 1. efficiently compressed videos 2. videos we can use reliably for video analysis (**NOT** determined by eye, but by rigorous benchmarking!) 3. ensure that we do not cause problems with data footprint. 

3. is also documented in this issue: #1689 
To address 3, we may not need to change these parameters, but instead change the data deletion/retention strategy. This will need to be documented as a SIPE ticket that will be delegated to SIPE to handle for project management. 

We will have videos compressed efficiently, able to be used in video analyses, and also not cause huge data footprint on the rigs themselves. 

Once this PR is finished, we will need to edit the ffmpeg_input_args and ffmpeg_output_args in the setting.csv file (this should be "C:\Users\svc_aind_behavior\Documents\ForagingSettings\Settings_boxX.csv"
)

Note that this PR is dependent on this PR: #1593 to even make sense. 

Yes. It also needs to be tested in ephy rigs with 3 cameras and 4 cameras. 

Yes it does. This is why this PR was created in the first place. I will add notes down below. 

### Current Suggested on-rig parameters: 

For ALL rigs: Dynamic Foraging default FFMPEG parameters, but a bit better
```
ffmpeg_input_args, -colorspace bt709 -color_primaries bt709 -color_range full -color_trc linear
ffmpeg_output_args, -vcodec h264_nvenc -rc vbr -cq 23 -preset fast -v warning
```

Galen's christmas parameters are preferred, once they are debugged: 

## Action Items
- [ ] Test combination of on-rig parameter, check sizing, check if this will impact data retention plan
- [ ] make sure second stage (job_type) is changed to gamma_encoding_fix_colorspace by default. "GAMMA_ENCODING_FIX_COLORSPACE". Currently it's without fix_colorspace.


rachelstephlee:
this PR shouldn' be closed, yet-- more parameters need to be debugged before we settle on the final set of parameters.

micahwoodard:
right now it doesn't look like there are any code changes from develop in the FFMPEG-PR-on-rig-parameter? The way github works, that makes a pr impossible which is why it was autoclosed. When there are code changes we can open back up!
image

rachelstephlee:
Galen thinks that this is a versioning issue for why his christmas parameters turn the videos green. 

this is the ffmpeg version on his computer:
image

this is the ffmpeg version on the ephys 4 rig 
image

PR #1718, opened 2026-03-17
Explicitly define Behavior Board event emission at startup

bruno-f-cruz:
### Describe changes:
The workflow now explicitly defines the events being propagated by the Harp Behavior Board instead of relying on external and asynchronous configuration. This should ALWAYS be the pattern when dealing with hardware (cameras, harp devices, etc...) where in developers should not have to rely on the existing (and unknown) configuration being set manually by users and instead "force" it to be what it is needed for the task.

Fixes https://github.com/AllenNeuralDynamics/aind-behavior-blog/issues/1411#issuecomment-4071278967

Events from Digital inputs, Cameras triggers and ADC reads should now be correctly propagated.

### DIFF
--- a/src/workflows/foraging.bonsai
+++ b/src/workflows/foraging.bonsai
@@ -2896,6 +2896,23 @@
             <Expression xsi:type="SubscribeSubject">
               <Name>BehaviorEvents</Name>
             </Expression>
+            <Expression xsi:type="Combinator">
+              <Combinator xsi:type="rx:Take">
+                <rx:Count>1</rx:Count>
+              </Combinator>
+            </Expression>
+            <Expression xsi:type="beh:CreateMessage">
+              <harp:MessageType>Write</harp:MessageType>
+              <harp:Payload xsi:type="beh:CreateEventEnablePayload">
+                <beh:EventEnable>PortDI PortDIO AnalogData Camera0 Camera1</beh:EventEnable>
+              </harp:Payload>
+            </Expression>
+            <Expression xsi:type="MulticastSubject">
+              <Name>BehaviorCommands</Name>
+            </Expression>
+            <Expression xsi:type="SubscribeSubject">
+              <Name>BehaviorEvents</Name>
+            </Expression>
             <Expression xsi:type="beh:Parse">
               <harp:Register xsi:type="beh:TimestampedAnalogData" />
             </Expression>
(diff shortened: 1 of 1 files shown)


micahwoodard:
tested on 6c and fixes issue

micahwoodard (approved):

XX-Yin:
@bruno-f-cruz thanks, can you post a screenshot to help understand the change?

bruno-f-cruz:
image

```xml
<?xml version="1.0" encoding="utf-8"?>
<WorkflowBuilder Version="2.8.1"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xmlns:rx="clr-namespace:Bonsai.Reactive;assembly=Bonsai.Core"
                 xmlns:beh="clr-namespace:Harp.Behavior;assembly=Harp.Behavior"
                 xmlns:harp="clr-namespace:Bonsai.Harp;assembly=Bonsai.Harp"
                 xmlns="https://bonsai-rx.org/2018/workflow">
  <Workflow>
    <Nodes>
      <Expression xsi:type="SubscribeSubject">
        <Name>BehaviorEvents</Name>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="rx:Take">
          <rx:Count>1</rx:Count>
        </Combinator>
      </Expression>
      <Expression xsi:type="beh:CreateMessage">
        <harp:MessageType>Write</harp:MessageType>
        <harp:Payload xsi:type="beh:CreateEventEnablePayload">
          <beh:EventEnable>PortDI PortDIO AnalogData Camera0 Camera1</beh:EventEnable>
        </harp:Payload>
      </Expression>
      <Expression xsi:type="MulticastSubject">
        <Name>BehaviorCommands</Name>
      </Expression>
    </Nodes>
    <Edges>
      <Edge From="0" To="1" Label="Source1" />
      <Edge From="1" To="2" Label="Source1" />
      <Edge From="2" To="3" Label="Source1" />
    </Edges>
  </Workflow>
</WorkflowBuilder>
```

micahwoodard:
@alexpiet @XX-Yin @ellahiltonvano do we want to merge this? I believe there are still concerns that some settings will be different in the current workflow because they are not applied at the beginning of the session and just assumed to be true and something might be different that isn't obvious. I don't have a solution to this, but wanted to address for visibility

XX-Yin:
@micahwoodard  Can you explain more about your concerns? What are your testing results?

bruno-f-cruz:
> @micahwoodard Can you explain more about your concerns? What are your testing results?

Obvious ones are:

- Camera configurations (we are overriding all we need at the start of the session)
- Sound card waveforms (we are uploading at the start of every session)
- Manipulator (we are setting it at the start of every session)
There may be other stuff that I am not aware of that could still be an issue. We may not need to change anything, but please make sure that, in that box, everything else (especially the items mentioned above) are still working as intended

XX-Yin:
I don't catch up the recent updates, and don't know some details.

micahwoodard:
> @micahwoodard Can you explain more about your concerns? What are your testing results?

My testing results are that when I pressed start, about 15 trials were able to be completed and the graph was updated correctly. @ellahiltonvano do you think other issues would be caught at integration testing?

ellahiltonvano:
@micahwoodard that's the testing I would do, on Monday we could have someone run a mouse in 6C for the morning session and see if they catch any issues during the 75min session?

micahwoodard:
@ellahiltonvano I am comfortable with that if everyone else approves

PR #1719, opened 2026-03-18
FIP metadata update

alexpiet:
  
### Describe changes:
- Changes name of FIP objective in the metadata from "Objective" to "FIP Objective"

- Its unclear what the objective is used for

- none

- none

- no

- It creates an inconsistency in the metadata. @hagikent How is the metadata for the VR foraging structured? Should we fix old metadata, or just correct going forward?


micahwoodard (approved):
This makes sense to me! Is any software dependent on this naming convention and would be broken if merged?

alexpiet:
> This makes sense to me! Is any software dependent on this naming convention and would be broken if merged?

I don't think so, but thats a good question for @hagikent

PR #1723, opened 2026-03-19
Pin numpy and pydantic versions in dependencies

DaniAlcAlv:
### Pull Request instructions:
- Please follow the [update protocol](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki/Software-Update-Procedures)
- Answer the questions below in detail. Your responses will be emailed to experimenters. 
- If the experimenters must do anything new, provide detailed step by step instructions on the [wiki](https://github.com/AllenNeuralDynamics/aind-behavior-blog/wiki)
- If computer maintainers need to manually update anything, provide detailed step by step instructions
- Use markdown syntax in order for your comments to be rendered reliably in the email: "1." instead of "1)", use four spaces for indents.
- If you use the keyword "skip email" in the title, it will skip the email updates
- Merges from "develop" into "production_testing" should use the keyword "production merge" in the title for reliable indexing of updates
- Merges from "production_testing" into "main" should use the keyword "update main"
  
### Describe changes:
Updated dependency versions in pyproject.toml to resolve environment setup and runtime errors.
Pinned:
 - pydantic==2.10.6
 - numpy==2.3.4 
These versions reflect a known working configuration validated during 8D setup.

### What issues or discussions does this update address?
Resolves a PydanticUserError caused by incompatible pydantic versions:
`model_config cannot be used as a model field name`
Aligns repo dependencies with versions confirmed to work by @micahwoodard during troubleshooting setting up rig 8D.

### Describe the expected change in behavior from the perspective of the experimenter
No changes to task functionality or workflow.

### Describe any manual update steps for task computers
Optional> Update dependencies for consistency rinning:
`pip install pydantic==2.10.6 numpy==2.3.4`

### Was this update tested in 446/447?
Tested in Rig 8D

### Does this update impact downstream processing by adding new saved files, or changing their format? If so, have you documented changes?
No

PR #1726, opened 2026-03-20
Fix numpy crash when calling float() on 1D array

### LINKED ISSUE #1725: Behavior crashes with numpy 2.4.3 on main.
with numpy 2.4.3 this line, as well as 533-536  errors:
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/653293091179fa284c22c6dccff4f0bd49848b1e/src/foraging_gui/MyFunctions.py#L522-L524

Because `float()` requires scalars, not 1D arrays in numpy 2.4. The fix is to not set `size=1` when calling `np.random.exponential` to get a scalar instead of a 1D array.

related: AllenNeuralDynamics/dynamic_foraging_error_tracking#1618
possibly related: #1723 (numpy should not be pinned IMO)

galenlynch:
Fixes #1725

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "deepdiff >=8, <9",
     "aind-data-schema==1.1.0",
     "aind-data-schema-models==0.5.6",
-    "pydantic >=2.9.2, <3",
+    "pydantic==2.10.6",
     "stagewidget==1.0.5",
     "python-logging-loki >=0.3.1, <2",
     "pykeepass >=4.0.7, <5",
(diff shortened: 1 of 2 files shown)


alexpiet:
Thanks for finding the bug and implementing the fix!

Can you merge into `develop` and not `main`?

micahwoodard (approved):
this makes sense to me, and I can't see a way this would break anything.  @galenlynch need to switch the branch to develop though.

galenlynch:
I’ll change the target branch soon. FYI you can change the default branch under repo settings to set the develop branch as the default PR target, which would be appropriate for this project: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/changing-the-default-branch

galenlynch:
It looks like there might be more instances of the same bug elsewhere. Let me make sure I have things running locally.

galenlynch:
OK this is working for me on master after fixing the same bug in one more location (hotfix/1725). The pydantic pin is also on this PR, so it competes with #1723.

PR #1731, opened 2026-03-24
adds user_email to manifest

### LINKED ISSUE #1728: Include user email in watchdog manifest
aind-data-transfer-service will soon require a user's email to be included in the upload request, so they know who to send failure notifications to. This means that the watchdog manifest now requires an email. You can fetch this email pretty easily by querying the active directory, see this PR for a clean example:

https://github.com/AllenNeuralDynamics/meso-prepare-transfer/pull/20

micahwoodard:
  
### Describe changes:
- adds user_email field to email 
- resolves #1728 
- None
- None but the schedule PI column will need to be changed to the PI's allen username
- [x] 446/7 
- No

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -7331,9 +7331,10 @@ class Window(QMainWindow):
 
 def get_user_email(username: str) -> str:
     domain = ms_active_directory.ADDomain("corp.alleninstitute.org")
+    domain_username = getpass.getuser()
     session = domain.create_session_as_user(
-        "svc_mpe",
-        authentication_mechanism=ldap3.ASL,
+        domain_username,
+        authentication_mechanism=ldap3.SASL,
         sasl_mechanism=ldap3.GSSAPI,
     )
 
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -7277,7 +7277,7 @@ class Window(QMainWindow):
 
             # Determine sci email
             sci = self._GetInfoFromSchedule(self.behavior_session_model.subject, "PI")
-            sci_email = None if not sci else get_user_email(sci)
+            sci_email = "svc_aind_behavior@alleninstitute.org" if not sci else get_user_email(sci)
 
             # Define contents of manifest file
             contents = {
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -7275,6 +7275,10 @@ class Window(QMainWindow):
                     self.VideoFolder.replace("\\", "/")
                 ]
 
+            # Determine sci email
+            sci = self._GetInfoFromSchedule(self.behavior_session_model.subject, "PI")
+            sci_email = None if not sci else get_user_email(sci)
+
             # Define contents of manifest file
             contents = {
                 "acquisition_datetime": self.behavior_session_model.date,
@@ -7299,6 +7303,7 @@ class Window(QMainWindow):
                 "schedule_time": schedule_time,
                 "project_name": self.Metadata_dialog.ProjectName.currentText(),
                 "script": {},
+                "user_email": sci_email 
             }
 
             # Define filename of manifest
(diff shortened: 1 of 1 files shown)


micahwoodard:
@patricklatimer

micahwoodard:
@ellahiltonvano @alexpiet @jackieswapp this will require changing the PI column in the schedule to have the username of the PI rather than name. I can't see anywhere in the df code where this would be an issue to change so this could be done anytime before this merge. Unless anyone can think of a reason why this would be an issue?

src/foraging_gui/Foraging.py
```diff
@@ -7275,6 +7275,10 @@ def _generate_upload_manifest(self):
                     self.VideoFolder.replace("\\", "/")
                 ]
 
+            # Determine sci email
+            sci = self._GetInfoFromSchedule(self.behavior_session_model.subject, "PI")
+            sci_email = None if not sci else get_user_email(sci)
```
alexpiet: If we don't get an email, do we still (1) run the session, (2) create a manifest, but (3) the email on the manifest will be None?

If the email is None, will the data still upload?

micahwoodard: Seems to me like the manifest has to have an email to upload? Is that true @patricklatimer?

src/foraging_gui/Foraging.py
```diff
@@ -7324,6 +7329,21 @@ def _generate_upload_manifest(self):
                 + "Please alert the mouse owner, and report on github.",
             )
 
+def get_user_email(username: str) -> str:
+    domain = ms_active_directory.ADDomain("corp.alleninstitute.org")
+    domain_username = getpass.getuser()
```
patricklatimer, line 7334: There might actually be differences in what different users can access here: Mekhla and Jon seemed to not be able to do this using svc_aind_scicomp, but svc_sipe and svc_mpe seem to work just fine. Maybe make it a configured argument instead of hardcoding, that way it's always the same?

src/foraging_gui/Foraging.py
patricklatimer: Yes, data-transfer-service won't accept datasets without emails. Maybe we could just set it to a default?

micahwoodard: Any ideas for a good default? maybe svc_aind_behavior? svc_mpe?

patricklatimer: Maybe svc_aind_behavior. 

You know, another possibility is that it could always go to the svc email and we could set up a Teams workflow to post to a channel upon email receipt.

micahwoodard: Do you think the svc email approach would facilitate more scientist engagement than a direct email? @alexpiet has an automated email that is sent out every day already

jackieswapp: @micahwoodard the PI field on the schedule comes from what's entered into the surgical work request. So if we wanted to go that route, we would have to change it there. I think that would be doable, but I'd probably prefer the default email route

alexpiet: I think its a good idea to email whatever scientist is listed on the schedule, and if there is no scientist listed or we can't get an email for whatever reason, then email a svc account and post on teams

src/foraging_gui/Foraging.py
micahwoodard: this has been working for validating user names so I might just leave it hard coded for now

micahwoodard:
@jackieswapp I think we are blocked merging this in the schedule is updated and people who fill out the surgical request form are notified. or we could merge and just default to svc_aind_behavior@alleninstitute.org for a while. I don't have a good sense of how urgent this is. @patricklatimer @alexpiet ?

alexpiet:
> @jackieswapp I think we are blocked merging this in the schedule is updated and people who fill out the surgical request form are notified. or we could merge and just default to [svc_aind_behavior@alleninstitute.org](mailto:svc_aind_behavior@alleninstitute.org) for a while. I don't have a good sense of how urgent this is. @patricklatimer @alexpiet ?

I'm not sure I understand the problem. The schedule contains the experimenter name, which should be the same name from the surgical request form. We can use that, and set svc_aind_behavior@allen... as a default. What is the problem?

As far as urgency, thats a question for @patricklatimer since they are waiting to upgrade the transfer service until all uploads have an email

patricklatimer:
I think the problem was that the schedule contained names instead of usernames. Although @micahwoodard I think the ms-active-directory might work with either usernames or full names, might be worth trying out!

To answer the urgency question, @jtyoung84 is waiting to deploy a transfer-service update that makes this email field required, and it is desirable to close this out by the end of April.

micahwoodard:
@patricklatimer I think you're right we can just use names for scis! Thanks for the suggestions. @jackieswapp @alexpiet Nothing in the schedule needs to change so I think we can just merge if I get approval

alexpiet (approved):

Issue #1717, opened 2026-03-10
Update DF waterlog functionality to call waterlog.exe


patricklatimer:
Waterlog exe will have new cli args for --mouse-id and --weight and stuff. Instead of the slims api in dynamic foraging, it should just call waterlog in a subprocess and let it take care of everything.

Depends on https://github.com/AllenNeuralDynamics/waterlog/issues/19

PR #1742, opened 2026-05-07
Replace SLIMS calls with subprocess call to waterlog app

patricklatimer:
### Describe changes:

Replaces SLIMS waterlog post with subprocess call to waterlog app running on the computer. Waterlog will be automatically opened if it isn't already running, and will be filled with the experimenter's name, mouse_id, weight, and earned_water.

https://github.com/AllenNeuralDynamics/dynamic-foraging-task/issues/1717

Experimenters will have to move over to waterlog and hit submit. (and potentially deliver water)

Requires being deployed alongside waterlog >4.1.0

Not yet

No, but waterlog database information (currently SLIMS, soon dataverse) will look slightly different.

### DIFF
--- a/src/foraging_gui/Foraging.py
+++ b/src/foraging_gui/Foraging.py
@@ -2105,25 +2105,10 @@ class Window(QMainWindow):
             for k, v in water_json
         }
 
-        # extract software information - TODO: Add this to waterlog cli
+        # extract software information to send to waterlog once it can accept it
         logging.info("Extracting software information from first data stream")
         software = session.stimulus_epochs[0].software[0]
-
-        # TODO: validate user first
-        # experimenter_name = self.behavior_session_model.experimenter[0]
-        # try:
-        #     resp = requests.get("http://aind-metadata-service/api/v2/active_directory/{experimenter_name}")
-        #     resp.raise_for_status()
-        #     validated_username = resp.json()['username']
-        # except Exception:
-        #     logging.warning(
-        #         "Could not validate experimenter name against aind-metadata-service", 
-        #         exc_info=True
-        #     )
-        #     validated_username = experimenter_name
-
-        # TODO: Should we remove the suggested water calculation from DF?
-        # That would mean removing the last two arguments below
+        # Access sw name/version with (software.url, software.version)
 
         waterlog_args = [
             self.Settings['waterlog_exe_path'],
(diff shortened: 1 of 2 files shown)


src/foraging_gui/Foraging.py
```diff
@@ -2171,53 +2105,63 @@ def _AddWaterLogResult(self, session: Session):
...
+        #     resp = requests.get("http://aind-metadata-service/api/v2/active_directory/{experimenter_name}")
+        #     resp.raise_for_status()
+        #     validated_username = resp.json()['username']
+        # except Exception:
+        #     logging.warning(
+        #         "Could not validate experimenter name against aind-metadata-service", 
+        #         exc_info=True
+        #     )
+        #     validated_username = experimenter_name
+
```
micahwoodard: experimenter name is already validated so we shouldn't need to do this here :)

src/foraging_gui/Foraging.py
```diff
@@ -2171,53 +2105,63 @@ def _AddWaterLogResult(self, session: Session):
...
+        # TODO: Should we remove the suggested water calculation from DF?
```
micahwoodard: I think it's fine? Might be nice to have if people can't/don't want to use waterlog

src/foraging_gui/Foraging.py
```diff
@@ -2171,53 +2105,63 @@ def _AddWaterLogResult(self, session: Session):
...
-        else:
-            logging.info("Adding waterlog to Slims.")
-            self.slims_client.add_model(model)
+        ###  Leftover fields that were previously sent to SLIMS
+        ###  TODO: validate that it's okay to leave them out
+        # total_water_ml=water["water_in_session_total"]+water["water_after_session"],
+        # workstation=session.rig_id,
+        # sw_source=software.url,
+        # sw_version=software.version,
+
```
micahwoodard: Leaving out total_water, rig id, and software from dataverse would be a good question for @jackieswapp. TO me, it seems fine because that is captured in the metadata

src/foraging_gui/Foraging.py
```diff
@@ -2171,53 +2105,63 @@ def _AddWaterLogResult(self, session: Session):
...
+        logging.info("Sending water info to waterlog")
+        process = subprocess.run([str(arg) for arg in waterlog_args])
+        
+        # TODO: Add message to user to to over to waterlog and hit submit
+
```
micahwoodard, line 2134: could be something like 

QMessageBox.information(self, "Waterlog", "Go to waterlog app to submit water information.")

micahwoodard (changes requested):
Looks good to me, some things just need to be clarified/finished up

micahwoodard (approved):
Tested on rig and works as expected

